PR #6, opened 2017-01-12
Updating .about.yml to remove ATF information

theresaanna:
Adds in OMB-specific information. I assumed for the purposes of this document that we are going to be implementing an instance of eRegs.

### DIFF
--- a/.about.yml
+++ b/.about.yml
@@ -3,16 +3,16 @@
 
 # This is a short name of your project that can be used as a URL slug.
 # (required)
-name: atf-eregs
+name: omb-eregs
 
 # This is the display name of your project. (required)
-full_name: ATF eRegulations
+full_name: OMB eRegulations
 
 # What is the problem your project solves? What is the solution? (required)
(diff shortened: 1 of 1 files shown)


.about.yml
```diff
@@ -3,24 +3,24 @@
...
-name: atf-eregs
+name: omb-eregs
 
 # This is the display name of your project. (required)
-full_name: ATF eRegulations
+full_name: OMB eRegulations
 
 # What is the problem your project solves? What is the solution? (required)
-description: The Bureau of Alcohol, Tobacco, Firearms and Explosives (ATF) maintains a large amount of complex information that the public needs to use, including regulations with important requirements related to munitions, firearms, and explosives. 18F adapted and customized eRegulations, an open source project started by the Consumer Financial Protection Bureau in 2013, to make ATF's regulations easier to find, read, and understand.
+description: The Office of Management and Budget (OMB) has policy documents they would like to be in a more usable and reference-able form. We are implementing an instance of eRegulations and expanding the platform to support additional document types.
```
cmc333333: I think we'll probably want to swap `policy documents` for `policies` based on what Matt's saying, but we can get to that later.

.about.yml
```diff
@@ -3,24 +3,24 @@
...
 
 # What is the measurable impact of your project? (required)
-impact: We produced a web application that offers substantial support for reading complex legal information, which helps ATF serve the public.
+impact: The application makes it easier for employees at OMB to understand, reference and use policy better, enabling them to work more effectively.
```
cmc333333: We'll need to regroup at some point about the intended audience -- I thought _agencies_ were the primary users.

PR #17, opened 2017-01-25
Import script, admin, some data types

cmc333333:
This is best reviewed changeset by changeset. It
* Adds an import script (+ test) for loading data from a CSV
* Activates the "Requirements" model in Django admin
* Converts NA/None OMB Policy IDs to blank
* Ensures policy types are within a predefined enum


theresaanna:
It all makes sense to me.

PR #18, opened 2017-01-27
More CSV model cleanup

cmc333333:
* Use date fields for policy creation/sunset
* Normalize Policy data into a separate model


theresaanna:
We're aalllllmost at the point where I'd want to defer to @tadhg-ohiggins on merging. Still, this is all looks straightforward and reasonable to me.

PR #19, opened 2017-01-27
Add install instructions

cmc333333:
Along the way, tested out the latest data file from OMB and fixed some little issues.


theresaanna:
Looks great, and makes some of what's here quite transparent!

PR #21, opened 2017-01-31
Adds Webpack build

theresaanna:
I don't recommend going through commit by commit. Just scope the files changed.

Adds a Webpack build running in Travis that runs babel, sass and eslint loaders. Also added the USWDS.


.travis.yml
```diff
@@ -1,12 +1,16 @@
 language: python
 sudo: false
 python: 3.5
+before_install:
+  - nvm install node
```
cmc333333: No biggie right now, but we should probably pick a version of node to build against at some point.

package.json
```diff
@@ -0,0 +1,40 @@
...
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "start": "webpack --config mywebpack.config.js"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/18F/omb-eregs.git"
+  },
+  "author": "",
+  "license": "ISC",
```
cmc333333: Another for later -- we probably want to use author/license values. [Current recommendations](https://github.com/18F/18f-cli/blob/develop/npm/index.js#L24-L25) are "18f" and "CC0-1.0"

cmc333333:
Two thumbs up!

PR #23, opened 2017-02-01
Convert Keyword fields to Taggit tags

cmc333333:
This will probably be easiest to understand commit by commit. It adds the taggit (and autosuggest) modules to replace our boolean keyword fields with a tag-based autocomplete. In the process it migrates existing data and updates the loader script.


reqs/management/commands/import_reqs.py
```diff
@@ -45,6 +46,33 @@ def policy_from_row(row):
...
+    """Creates or retrieves Keyword models"""
+    def __init__(self):
+        self.cache = {}
+
+    @staticmethod
+    def keywords(row):
+        to_return = []
+        for field, value in row.items():
+            if field == 'Other (Keywords)':
+                to_return.extend(kw.strip()
```
tadhg-ohiggins: Can we break this out? Nested for loops in comprehensions are always a little tough to grasp.

reqs/migrations/0013_auto_20170131_1829.py
```diff
@@ -0,0 +1,84 @@
...
+    Keyword = apps.get_model('reqs', 'Keyword')
+    KeywordConnect = apps.get_model('reqs', 'KeywordConnect')
+    Requirement = apps.get_model('reqs', 'Requirement')
+    keyword_cache, connections = {}, []
+    for req in Requirement.objects.all():
+        req_keywords = []
+        for field_name, tag_name in attr_to_keyword.items():
+            if getattr(req, field_name):
+                req_keywords.append(tag_name)
+        req_keywords.extend(kw.strip()
```
tadhg-ohiggins: Same comment here: I think something like
```
for kw_semi in req.other_keywords.split(";"):
    for kw_comma in kw_semi.split(","):
        if kw_comma.strip():
            req_keywords.append(kw_comma.strip())
```
is clearer

tadhg-ohiggins:
Looks good apart from my minor comments.

reqs/migrations/0013_auto_20170131_1829.py
cmc333333: I'm against deeply nested (aka "pointy") code, but I think we can satisfy both of us. Giving it a go

PR #30, opened 2017-02-06
Initial cloud.gov setup

### LINKED ISSUE #22: Cloud.gov setup
We have the cloud.gov IAA signed, now we just need to set up up the org/spaces/manifests. This requires contacting cloud.gov folks (I think there's a form iirc) and then granting our people the right perms.

We can include getting the app ready for cloud.gov in this ticket or break that into another one.

cmc333333:
Resolves #22.

There are several bits here that can be optimized, but I figured this is a reasonable place to start.


.travis.yml
```diff
@@ -2,15 +2,25 @@ language: python
...
+- pip install -r requirements_dev.txt
+- npm install
 script:
-  - py.test
-  - flake8
-  - wget https://github.com/ombegov/policy-v2/raw/master/assets/Phase1_CombinedQA_AllPhase1_Nov21.csv
-  - iconv -f Windows-1252 -t utf-8 Phase1_CombinedQA_AllPhase1_Nov21.csv > data.csv
-  - python manage.py migrate
-  - python manage.py import_reqs data.csv   # just looking for explosions
-  - webpack
```
cmc333333: Reformatting from the `travis` command, sorry.

PR #38, opened 2017-02-10
Use Docker for local development

cmc333333:
This tries to standardize development using docker by using it to contain all of the Python and JS environment. It also introduces a make file (which should be executable on Linux and OS X without issue) as the command center for the app.

There's a lot of boiler plate here (notably, the docker-compose files) that we can probably clean up and no doubt some edge cases I missed. That said, I could get the app running in one command on OSX (without Node/Python/Postgres configured) and have imported data and an admin user with one command a piece.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,13 @@
 # No need to commit databases
 *.sqlite3
 
+# Javascript stuff
+node_modules
+reqs/static/js/bundle.js
+
+# Data
+*.csv
+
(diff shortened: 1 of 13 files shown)


cmc333333:
Booo. Make is only installed on OSX if developer tools are installed.

cmc333333:
This is superseded by #39, which removes the Make dependency.

PR #39, opened 2017-02-13
Use Docker for local development (take 2)

cmc333333:
This tries to standardize development using docker by using it to contain all of the Python and JS environment. The `docker-compose` command serves as the primary entry point to the app, with single-use "services" acting as commands. I don't think this is _too_ much of a pain (it's akin to running `bundle exec`) but I'd appreciate feedback. I can get the app running in one command on OSX (without Node/Python/Postgres configured), which is pretty awesome. Hypothetically, the same should apply on Windows.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,13 @@
 # No need to commit databases
 *.sqlite3
 
+# Javascript stuff
+node_modules
+reqs/static/js/bundle.js
+
+# Data
+*.csv
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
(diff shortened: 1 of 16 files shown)


toolness:
This is neat!  I was not aware of the funky `<<` key in docker-compose's services definitions.  Um, where is that documented?  I'm not seeing it anywhere in the [Compose file reference](https://docs.docker.com/compose/compose-file/)... Is it a more general YAML thing?

cmc333333:
Thanks @toolness! Yes, `<<` and `&` are used for references/anchors within a single document; they're just a naive dictionary merge. https://learnxinyminutes.com/docs/yaml/ has some examples, but I've seen them more in the rails context.

LinuxBozo:
Without using some type of external build tool/script, I think this is a great approach.

PR #40, opened 2017-02-14
Update requirements in place when importing

### LINKED ISSUE #36: Update existing requirements on CSV import
Currently, the CSV import more or less assumes that the database is clear. Remove that assumption and edit the existing policies + requirements rather than only inserting.

cmc333333:
Previously, we assumed the database would be empty. Resolves #36


cmc333333:
Hm. So the Travis error is due to a sqlite limitation. Should we fix it in the code (by cutting a single SQL query into several), or just assume #39 (which ensures Postgres)?

PR #45, opened 2017-02-16
Express server

### LINKED ISSUE #41: Add a Node app to Docker/cloud.gov
As we'll be trying isomorphic JS, we'll need to add hooks for running the app in our different environments.

cmc333333:
This adds a NodeJS server into the mix and adds commands for running in on Docker and in cloud.gov.

Lots of improvements can be made here, including:
* development mode (including hot reload)
* add a dependency between Node and the api in the compose.yml file (so that we only run `docker-compose up prod` or `docker-compose up dev` and get both the Node and Python apps)
* rejigger the directory structure to make cloud.gov deploys less data intensive (currently, we're shipping up all of the Python and Node files for both types of apps)

That said, I think this is a reasonable first step and should resolve #41.

## Breaking changes
The docker-compose Django apps have been renamed `prod-api` and `dev-api`


theresaanna:
I agree that this is a good place to start! We can add all of that additional functionality in time. 👍

PR #46, opened 2017-02-16
Allow cross-origin API usage.

cmc333333:
We'll want to replace this replace this with a more strict strategy once we
have a working API. In the mean time, open the flood gates to all.


theresaanna:
Open the gates!

PR #47, opened 2017-02-16
Isomorphic React Routes Buzzword Buzzword

cmc333333:
Builds on #45 ([diff](https://github.com/18F/omb-eregs/compare/41-node-app-container...42-react-routing)) - this adds an a two-page app which will render page content on the server side during initial page load (and for non-JS users) and client-side when that's available.

Server-side always generates html like this:
```html
<html>
    <body>
        <div id="app">
            <div>
                <h1>App</h1>
                ... [rest of the rendered content]
            </div>
        </div>
        <script src="/static/browser.js"></script>
    </body>
</html>
```
If the user doesn't have JS, the script won't be included, so all of the links will cause a full page reload. If the user *does* have JS, they will trigger React to render the _same_ content inside the "app" id except that all of the link clicks are caught and rendered within the browser. For browsers without Browser history the library we're using falls back to full page refresh.

What's super cool (and the promise of isomorphic JS) is that the server and browser code are thin shims; both call into the same React application.


ui/server.jsx
```diff
@@ -0,0 +1,40 @@
...
+
+import routes from './routes';
+
+const app = express();
+const env = cfenv.getAppEnv();
+
+app.use(morgan('combined'));
+app.use('/static', express.static(path.join('ui-dist', 'static')));
+
+app.get('*', (req, res) => {
```
theresaanna: TODO: are there any libraries or patterns that make this a bit neater? I know that Node tends to get messy this way. Definitely not something to hold up merge for, especially since this is a first step. It feels wrong to handle the response and define the base HTML in the same place, but I'm also new to JSX. This may very well be the pattern!

theresaanna:
I had one comment but this is a fantastic foundation!

Issue #51, opened 2017-02-17
Add API Endpoint for Requirements


cmc333333:
Should allow for filtering by keyword.

cmc333333:
Resolved by #61

Issue #50, opened 2017-02-17
Add API Endpoint for Keywords


cmc333333:
To view a list of keywords, we need to be able to retrieve them.

cmc333333:
Resolved by #61

PR #61, opened 2017-02-21
Initial Django Rest Framework endpoints

### LINKED ISSUE #48: Add Django Rest Framework

cmc333333:
This gives us an endpoint for keywords, policies, and requirements and some pretty nifty filtering capabilities for each. I've not doubt missed some potentially useful combinations, but we can definitely filter by most of the useful fields at this point (including requirements by a collection of keywords).

This should resolve #48, #50, #51


tadhg-ohiggins:
👍 
django-filter looks like a nice find.

PR #66, opened 2017-02-24
Click a keyword, get requirements list

theresaanna:
Adds functionality where a user can get a list of keyword links and click on any link to see the requirements associated with that keyword. This is bare bones and simply functionality.

- Creates the `Requirements` component
- Moves keywords into their own `Keywords` component
- NEEDS TESTS. I don't want to write anything else without writing tests first.


ui/components/keywords.jsx
```diff
@@ -0,0 +1,43 @@
...
+function Keyword(props) {
+  return <li>{props.keyword}</li>;
+}
+
+function Keywords(props) {
+  return (
+    <div>
+      <h1>Keywords</h1>
+      <ul>
+        {props.data.map(keyword => <Link to={{ pathname: '/requirements/', query: { keywords__name__in: keyword.name } }}><Keyword keyword={keyword.name} /></Link>)}
```
cmc333333: Something for the future -- we can use `keywords__name=myKeyword` when we're looking for a single match. Those filters are really powerful; I think @donjo may be able to get some interesting data just by manipulating the URL.

cmc333333:
:+1: this will be super useful @theresaanna, nice work!

PR #71, opened 2017-03-02
Devopsy tweaks and updated docs

cmc333333:
This should be a wee bit nicer for initial setup and adds some docs for maintenance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,8 +24,8 @@ automatically with Docker on Windows and OS X).
 Let's start by adding an admin user.
 
 ```bash
-docker-compose run manage.py migrate  # set up database
-docker-compose run manage.py createsuperuser
+docker-compose run --rm manage.py migrate  # set up database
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -79,14 +79,41 @@ There are two types of entry points:
...
   * `pip-compile`
   * `npm`
   * `webpack`
 
+### Resolving common container issues
+
+If a Javascript dependency has been added (indicated by an error within
+`node_modules`), run
+```sh
+docker-compose run --rm npm install
```
donjo: `docker-compose build` and `npm install` have to be run the first time through right? If so, should we add a short section at the beginning to mention this? Otherwise it looks like you can jump right in to the migrate/superuser stuff.

cmc333333: This PR actually fixes that oversight -- `npm install` is ran if there's no `node_modules` dir

PR #74, opened 2017-03-03
35 new data swipe

tadhg-ohiggins:
Use new data file /
Cope with even more date styles /
Improve the test suite.


reqs/management/commands/import_reqs.py
```diff
@@ -42,6 +42,19 @@ def convert_date(string):
             raise ValueError("Not a date: {0}".format(string))
 
 
+def find_key(options, row, highlander=True):
```
cmc333333: Looks like highlander's never used and I'm not sure what it's supposed to do -- can we drop it?

reqs/management/commands/import_reqs.py
```diff
@@ -79,17 +95,18 @@ def priority_split(text, *splitters):
...
+        self.fields = fields
 
-    @staticmethod
-    def keywords(row):
+    def keywords(self, row):
         to_return = []
-        for field, value in row.items():
-            if field == 'Other (Keywords)':
+        for field in self.fields:
+            value = row.get(field, None)
```
cmc333333: Heads up that with `dict`s you don't need to `d.get(field, None)` -- you can just `d.get(field)` (it'll default to `None` if not present).

reqs/models.py
```diff
@@ -31,18 +31,18 @@ def tags_for(cls, model, instance=None, **extra_filters):
...
     law = 'Law'
+    memorandum = 'Memorandum'
+    national_action = 'National Action Plan'
     plan = 'Plan'
-    hspd = 'HSPD'
     ppd = 'PPD'
-    national_action = 'National Action Plan'
+    review = 'Policy Review'
+    standards = 'Standards'
+    strategy = 'Strategy'
```
cmc333333: Heads up that re-ordering these will likely require a Django migration -- the database won't change, but Django likes to serialize the full configuration of models (including the order of options).

reqs/tests/import_reqs_tests.py
```diff
@@ -140,3 +140,101 @@ def test_repeat_row():
...
+    Test that we can handle a variety of names for fields (in this case, the
+    OMB Policy ID field).
+    """
+    numbers = iter(range(1, 200))
+    row = {'policyNumber': '123', 'policyTitle': 'title 1',
+           'uriPolicyId': 'http://example.com/a',
+           'policyType': 'Guidance', 'policyIssuanceYear': '12/20/2001',
+           'policySunset': 'NA'}
+    for key in ("ombPolicyId", "ombPolicyID"):
+        num = next(numbers)
```
cmc333333: This'll just be 1, then 2, yes? Would it make sense to use `enumerate` instead?

reqs/tests/import_reqs_tests.py
```diff
@@ -140,3 +140,101 @@ def test_repeat_row():
...
+        row[key] = "policy"
+        row["policyNumber"] = str(num)
+        policy_proc = import_reqs.PolicyProcessor()
+        policy = policy_proc.from_row(row)
+        assert policy.policy_number == num
+        del row[key]
```
cmc333333: Should this be asserting that `policy.omb_policy_id` is being set correctly?

reqs/tests/import_reqs_tests.py
```diff
@@ -140,3 +140,101 @@ def test_repeat_row():
...
+    """
+    Test that we can handle a variety of names for fields (in this case, the
+    OMB Policy ID field).
+    """
+    numbers = iter(range(1, 200))
+    row = {'policyNumber': '123', 'policyTitle': 'title 1',
+           'uriPolicyId': 'http://example.com/a',
+           'policyType': 'Guidance', 'policyIssuanceYear': '12/20/2001',
+           'policySunset': 'NA'}
+    for key in ("ombPolicyId", "ombPolicyID"):
```
cmc333333: This _might_ be a good use case for `pytest.mark.parametrize` -- it'd generate multiple tests which would tell us exactly which key name was wonky.

tadhg-ohiggins:
@cmc333333 I cleaned up the tests around field names, and thank you for pointing out `pytest.mark.parametrize`, which I wasn't familiar with and is perfect for this use case. Switching to use that eliminated some of the messy things from my prior approach.

I agree that handling multiple CSV formats adds complexity, but I hope that my recent changes clean that up somewhat. I'm not quite ready to assume that future field names changes won't happen, and leaving that complexity in the code lets us handle those hypothetical changes fairly easily.

reqs/management/commands/import_reqs.py
```diff
@@ -11,6 +11,17 @@
...
+# We have a set of field names that change depending on the input file, and
+# storing that information in one spot is probably the best appraoch.
+FIELDS = {
+    "verbs": ("reqVerb", "req_verb", "verb"),
+    "req_ids": ("reqId", "reqID", "reqid", "req_Id", "req_ID", "req_id"),
+    "entities": ("Impacted Entity", "agenciesImpacted"),
+    "citations": ("Citation", "Citation ", "citation"),
+    "uri_policy_ids": ("uriPolicyID", "uriPolicyId"),
+    "omb_policy_ids": ("ombPolicyID", "ombPolicyId")
+}
```
cmc333333: Great idea, I like this much more

reqs/management/commands/import_reqs.py
```diff
@@ -65,8 +76,8 @@ def __init__(self):
     def from_row(self, row):
         """Retrieve/create/update a Policy object"""
         policy_number = int(row['policyNumber'])
-        uri_key = find_key(("uriPolicyID", "uriPolicyId"), row)
-        omb_key = find_key(("ombPolicyID", "ombPolicyId"), row)
+        uri_key = find_key(FIELDS["uri_policy_ids"], row)
+        omb_key = find_key(FIELDS["omb_policy_ids"], row)
```
cmc333333: Small detail, but would it make sense for `find_key` to lookup the values in `FIELDS` instead of calculating the field names in the caller?

Issue #77, opened 2017-03-06
Add authentication to UI pages


cmc333333:
Required before we start the ATO process. I'm thinking basic auth will do just fine given that we're over HTTPS.

PR #92, opened 2017-03-07
Add ability to require basic auth on UI pages.

cmc333333:
Also includes docs on how to set this up locally.

Not all of #77 as we need to configure the environments, too.

### DIFF
--- a/ui/basic-auth.js
+++ b/ui/basic-auth.js
@@ -1,0 +1,18 @@
+/* Exports a single function, which installs a basic auth check if that's how
+ * we're configured. Uses a constant-time buffer comparison */
+import bufferEq from 'buffer-equal-constant-time';
+import httpAuth from 'http-auth';
+
+export default function (envCreds, expressApp) {
+  if (envCreds && envCreds.UI_BASIC_AUTH
+      && Object.keys(envCreds.UI_BASIC_AUTH).length > 0) {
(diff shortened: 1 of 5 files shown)


ui/basic-auth.js
```diff
@@ -0,0 +1,18 @@
+/* Exports a single function, which installs a basic auth check if that's how
+ * we're configured. Uses a constant-time buffer comparison */
+import bufferEq from 'buffer-equal-constant-time';
+import httpAuth from 'http-auth';
+
+export default function (envCreds, expressApp) {
```
theresaanna: Is this a standard pattern for doing the check? I ask only because I don't feel qualified to truly review this piece. I have spent time wrapping my head around it enough to know that I wouldn't be able to see any nuanced issues. If it's the way everyone does this sort of thing, then I'm happy to rely on the knowledge of many. If you feel very comfortable with this code, @cmc333333, that's also good enough for me.

cmc333333: Eep, sorry to miss this @theresaanna -- the library would prefer if we had credentials in a file (which doesn't work so well with twelve-factor-style environment variables), but they mention this style in [their examples](https://github.com/http-auth/http-auth#custom-authentication).

This uses constant-time string comparison for the reasons I mentioned in http-auth/http-auth#78

PR #93, opened 2017-03-07
Add minimal policy->req interface.

### LINKED ISSUE #87: Initial Policy Filter
Right now we have a keywords filter that goes:
1. pick "keywords"
2. pick a specific keyword
3. page through results of requirements

Make something similar for policies:
1. pick "policies"
2. pick a specific policy
3. page through results of requirements

cmc333333:
Copy-paste then tweak the Keywords UI into a Policies UI, where the user can
see a list of all policies, and click through to the requirements contained in
that policy. There's a lot of overlap with the Keywords code, but I figure
we'll be replacing both soon, so I'm not too worried.

Should resolve #87.

### DIFF
--- a/reqs/serializers.py
+++ b/reqs/serializers.py
@@ -10,7 +10,7 @@ class PolicySerializer(serializers.ModelSerializer):
         model = Policy
         fields = (
             'policy_number', 'title', 'uri', 'omb_policy_id', 'policy_type',
-            'issuance', 'sunset'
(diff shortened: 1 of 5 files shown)


theresaanna:
I would agree with your assessment that these will be replaced or at least largely reworked. LGTM otherwise.

cmc333333:
Thanks for looking it over @theresaanna !

Issue #94, opened 2017-03-07
Sidebar nav: keyword viewer


cmc333333:
As a first step to #88, display the currently selected keywords in the left sidebar. No need for interactivity.

cmc333333:
Resolved in #95

PR #95, opened 2017-03-08
Active Keyword viewer

cmc333333:
This adds a small widget to the left side of the screen which displays the active keyword filters and allows the user to remove them. Users can only add new keywords by modifying the URL, but this should satisfy #94.

This also switches to using keyword ids rather than strings for consistency with future filters.

Looks like:
screen shot 2017-03-08 at 11 19 12 am


cmc333333:
Quick note that, on re-evaluation, it probably makes sense to use a multiselect for progressive enhancement. I think we can still start here, though.

ui/components/req-filter-ui.jsx
```diff
@@ -0,0 +1,71 @@
+import axios from 'axios';
+import React from 'react';
+import { Link } from 'react-router';
+
+import { apiUrl } from '../globals';
+
+function FilterItem({ keywordIds, keyword, query }) {
```
theresaanna: I don't want to be nitpicky, and I don't know what the standard is wrt naming React components, but might we want to consider simplifying naming here by changing `FilterItem` -> `Filter` and `ReqFilterUI` to something like `FilterList`. To my mind this feels clearer, but I could be alone in this. Certainly not something to hold up this PR. I know it glosses over the fact that we will have other filters in the future, and, to my mind, we might do something that looks like https://facebook.github.io/react/docs/composition-vs-inheritance.html#specialization where we compose elements that are special versions of `Filter` and `FilterList` or whatever they are called, and create a more generic implementation of these components when the time comes.

cmc333333: Can do! Thanks for the pointer, too; I've been avoiding classes so far, but inheritance is exactly where they should shine.

cmc333333:
Moving test-writing to a separate PR

PR #98, opened 2017-03-08
Use npm shrinkwrap

### LINKED ISSUE #97: Implement npm shrinkwrap
(To pin our dependencies)

cmc333333:
Pin the exact versions of our libraries (and their dependencies). This is akin to `Gemfile.lock` file. `npm install` should use the data automatically.

Only potential caveat here: be sure to use `npm install --save` when adding new dependencies.

Also removes bound ports from a few of the service commands (to prevent port conflicts).

Should resolve #97


cmc333333:
Thanks @donjo !

PR #102, opened 2017-03-09
Add stylesheet structure and fonts

donjo:
This PR covers 2 things
- adds some font files from the USWDS, integrating them into the webpack build
- establishes a Sass file structure where styles can be broken down into separate files for each component.


ui/assets/css/_fonts.scss
```diff
@@ -0,0 +1,52 @@
...
+ Fonts
+ ==========================================================================
+ Place all @font-face rules in fonts.scss
+ */
+
+ @font-face {
+   font-family: 'Source Sans Pro';
+   font-style: normal;
+   font-weight: 300;
+   src: url('/font/SourceSansPro-Light.ttf');
```
cmc333333: Should these be relative, like `../font/[name].ttf`? I'm seeing 404s at this absolute URL.

donjo: Ah yes, I forgot to swing back around and update these. Relative paths didn't seem to work but I updated to include the _right_ absolute path

PR #106, opened 2017-03-10
An initial set of UI tests

### LINKED ISSUE #78: Setup architecture for testing React components
More generally, JS unit tests.

cmc333333:
Using [jest](https://facebook.github.io/jest/) as a test runner and [enzyme](http://airbnb.io/enzyme/) as a react-traversing toolkit. This doesn't do anything with mocks, but I figure it's enough to get us started.

Should resolve #78

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["es2015", "react"]
+}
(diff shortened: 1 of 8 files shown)


cmc333333:
:+1: Thanks for reviewing!

Issue #105, opened 2017-03-10
Get clarification on whether `None` is really a category.


tadhg-ohiggins:
Requirement 1050.30 has the string `None` in the `Other` field, and we need to know if this should indeed be a category.

I've asked about this in Slack.

Thomas-Romanoff:
Fixed

Issue #104, opened 2017-03-10
Get clarification on similar keyword terms


tadhg-ohiggins:
There are keywords in the data that seem very close to each other, so we should determine whether they are really just one category and which term is the canonical one:

- `Policy Exceptions` and `Policy Exemptions`.
- `Governance - Organization` and `Governance - Org Structure`.
- `Research` and `Research & Development`.

I've asked about this in Slack.

tadhg-ohiggins:
From Matt in Slack:

> they are not distinct. go ahead and use your judgement or defer to the more common term.

So unless anyone objects I'll do the following:

- [ ] Change all instances of `Policy Exceptions` to `Policy Exemptions`.
- [ ] Change all instances of `Governance - Org Structure` to `Governance - Organization`.
- [ ] Change all instances of `Research & Development` to `Research`.

Issue #103, opened 2017-03-10
Figure out what's wrong with requirements 1315.01–1315.26


tadhg-ohiggins:
These all have `X` in the `Other` field, which should be populated with a list of comma-separated keywords.

Presumably the columns have been inadvertently shifted over, and they're not supposed to have keyword of `X`, but we need to confirm this and get a corrected version.

I've asked about this in Slack.

tadhg-ohiggins:
The requirements in question all derive from final guidance on implementing the Plain Writing Act of 2010, found in this memo: <https://obamawhitehouse.archives.gov/sites/default/files/omb/memoranda/2011/m11-15.pdf>.

All the data in the columns for these requirements seems to be in the right place except for the keywords columns.

The data suggests that the categories for these requirements are:

- Customer Services (which seems to make sense)
- Definitions (which seems to make sense)
- “X” as a category in “Other”, which seems clearly wrong. The memo includes requirements for compliance reports, so the most likely explanation is that the last “X” should be moved left one column.

Thomas-Romanoff:
Fixed

PR #108, opened 2017-03-12
101 keyword inconsistencies

tadhg-ohiggins:
Often the keywords /
Are inconsistent or wrong /
This mitigates that.

There are still keyword related issues, <https://github.com/18F/omb-eregs/issues/103>, <https://github.com/18F/omb-eregs/issues/104>, and <https://github.com/18F/omb-eregs/issues/105>, but since they require information from OMB I'll address them in separate branches/PRs.

### DIFF
--- a/reqs/management/commands/import_reqs.py
+++ b/reqs/management/commands/import_reqs.py
@@ -21,43 +21,50 @@ FIELDS = {
     "uri_policy_ids": ("uriPolicyID", "uriPolicyId"),
     "omb_policy_ids": ("ombPolicyID", "ombPolicyId")
 }
+# The same appears to be true for keywords.
+# Executive decision: keyword values must be init-caps.
+# Here we're listing the problematic values and what they should be.
+KEYWORDS = {
+    "Emergency Preparedness?": "Emergency Preparedness"
+}
 
 
-def convert_omb_policy_id(string):
-    if string in ('NA', 'None'):
+def convert_omb_policy_id(value):
+    if value in ('NA', 'None'):
         return ''
-    return string
+    return value
 
 
-def convert_policy_type(string):
-    """Raises a ValueError if the string type can't be found"""
-    string = string.strip()
-    if 'memo' in string.lower():
+def convert_policy_type(policy_type):
+    """Raises a ValueError if the policy type can't be found"""
+    policy_type = policy_type.strip()
+    if 'memo' in policy_type.lower():
         return PolicyTypes.memorandum.value
-    elif 'circular' in string.lower():
+    elif 'circular' in policy_type.lower():
         return PolicyTypes.circular.value
-    elif string in ('', 'NA'):
+    elif policy_type in ('', 'NA'):
         return ''
     else:
-        return PolicyTypes(string).value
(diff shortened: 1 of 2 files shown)


reqs/management/commands/import_reqs.py
```diff
@@ -21,43 +24,63 @@
...
+KEYWORDS = {
+    "Commodity It": ["Commodity IT"],  # This one is due to str.title()
+    "Data Management/Standards. Reporting": [
+        "Data Management/Standards",
+        "Reporting"
+    ],
+    "Definition": ["Definitions"],
+    "Emergency Preparedness?": ["Emergency Preparedness"],
+    "Record Management": ["Records Management"],
+}
```
cmc333333: I like the declarative style

reqs/management/commands/import_reqs.py
```diff
@@ -111,12 +134,35 @@ def __init__(self, fields):
         self.cache = {}
         self.fields = fields
 
+    def normalize_keywords(self, values):
```
cmc333333: Should this be a static method? `self` isn't used.

reqs/management/commands/import_reqs.py
```diff
@@ -111,12 +134,35 @@ def __init__(self, fields):
...
+        """
+        Executive decisions:
+        +   Keywords cannot contain whitespace at the start or end.
+        +   All whitespace will be converted to a single space.
+        +   Keyword values must be init-caps.
+        +   Use of hyphens as separators requires a single space around the
+            hyphen.
```
cmc333333: I'm uncertain about these last two, but I'm cool with using them to clean the data; I don't think they can hurt.

reqs/tests/import_reqs_tests.py
```diff
@@ -251,3 +258,39 @@ def test_varying_policy_headers(opfid_field, uri_field):
...
+    Test that we don't add keywords with different capitalization into the
+    data.
+    """
+    csv_str = "\n".join([
+        SAMPLE_CSV2[0],
+        SAMPLE_CSV2[1].replace("Software Development Lifecycle/Agile", keyword)
+    ])
+    blank_csv_file.write(csv_str)
+
+    call_command('import_reqs', str(blank_csv_file))
```
cmc333333: No biggie, but do you think that we could avoid all of this setup (e.g. by just calling `keywords`)? The more setup, the more complicated the tests are to understand -- for example, I had to jump to the top of the file to figure out why "Software" was in this list.

reqs/management/commands/import_reqs.py
tadhg-ohiggins: I'm not adamant that these should be the standards, but I do think we need standards of some kind for consistency.

reqs/tests/import_reqs_tests.py
tadhg-ohiggins: Good point on “Software” being somewhat random. I still wanted this to be more comprehensive than just keywords, though. My change (which I'll commit shortly) here is to add a new line to the sample CSV that doesn't include the `x` in the Software row and to test against that instead, removing that bit of mystery.

PR #109, opened 2017-03-13
Normalize some stray variant keywords.

### LINKED ISSUE #104: Get clarification on similar keyword terms

tadhg-ohiggins:
There are many words / 
But we don't need all of them /
In our keyword list.

For keywords, change:
- `Policy Exceptions` to `Policy Exemptions`.
- `Governance - Org Structure` to `Governance - Organization`.
- `Research & Development` to `Research`.

Fixes #104


tadhg-ohiggins:
Turns out that we should be using `Governance - Org Structure` because that matches the existing header in the CSV.

- [ ] Change `Governance - Organization` to `Governance - Org Structure`.

Issue #110, opened 2017-03-13
What to do with many rows that have reqStatus set to a date?


tadhg-ohiggins:
Many rows in the current CSV (<https://github.com/ombegov/policy-v2/blob/master/assets/AllPhasesCombinedTaggingTemp_Feb16.csv>) have dates in the ninth column, which according to the CSV header should be `reqStatus`.

For example, all of the 5.xx and 21.xx requirements. Many of these have non-date content in the `policySunset` field, strongly suggesting that these fields have been transposed.

tadhg-ohiggins:
Until I hear otherwise, I'm assuming those values have been transposed, and have added a function for detecting this and swapping them.

tadhg-ohiggins:
Addressed by <https://github.com/18F/omb-eregs/pull/111>.

PR #111, opened 2017-03-14
Fix transposed reqStatus/policySunset values

tadhg-ohiggins:
Something should be here /
But instead it's over there. / 
Sunlight limns seagulls.

I had sample CSV lines from the old data in the tests instead of from the new data. When I corrected this, the tests revealed the problems with transposed values outlined in <https://github.com/18F/omb-eregs/issues/110>.

This PR changes the tests to use the correct data, and handles the transposed values.


reqs/management/commands/import_reqs.py
```diff
@@ -93,6 +93,36 @@ def find_key(key_name, row):
...
 
+def handle_transposed_reqstatus(row):
+    """
+    There are a number of rows that have date content in the ``reqStatus``
+    field and non-date content in the ``policySunset`` field.
+
+    We're going to assume that this means they were transposed, and correct
+    those rows.
+    """
+    if "reqStatus" not in row:
```
cmc333333: Should this check the existence of policySunset as well?

reqs/management/commands/import_reqs.py
```diff
@@ -93,6 +93,36 @@ def find_key(key_name, row):
...
+
+    status_date, sunset_date = None, None
+    try:
+        status_date = dateutil_parser.parse(
+            row["reqStatus"]).strftime("%Y-%m-%d")
+    except ValueError:
+        pass  # We expect this to not be a date.
+    try:
+        status_date = dateutil_parser.parse(
+            row["policySunset"]).strftime("%Y-%m-%d")
```
cmc333333: Are these `strftime`s needed?

reqs/management/commands/import_reqs.py
```diff
@@ -93,6 +93,36 @@ def find_key(key_name, row):
...
+    except ValueError:  # dateutil's error messages aren't friendly
+        logger.warning("Not a date in policySunset: {0}".format(
+            row["policySunset"]))
+    if status_date is not None and sunset_date is None:
+        # Looks like they've been transposed.
+        row["reqStatus"] = row["policyStatus"]
+        row["policySunset"] = status_date
```
cmc333333: The logic around these `try`s and `if` _feels_ like it can be simplified. What do you think about adding a helper function and then simplifying to:

```python
if is_date(row['reqStatus']) and not is_date(row['policySunset']):
    (row['reqStatus'], row['policySunset']) = (row['policySunset'],
                                               row['reqStatus'])
```
?

reqs/management/commands/import_reqs.py
tadhg-ohiggins: Sure; I left it out because it didn't come up with the current data, but it's probably worth adding.

reqs/management/commands/import_reqs.py
tadhg-ohiggins: Yes, because the row values get passed to something later that converts them to dates, and this seemed like the better option than special handling only for rows that get this transposition done to them.

tadhg-ohiggins: Oh, I see what you mean; it might be better to do it your way.

reqs/management/commands/import_reqs.py
tadhg-ohiggins: Seems like a good idea, I'll do something like that.

PR #116, opened 2017-03-15
Keyword autocompleter

### LINKED ISSUE #88: Sidebar nav
Should be able to add and remove keywords from the set of filters (continue to use union/OR until told otherwise)

cmc333333:
This builds on #112 by adding a autocompleting widget for users with client-side Javascript. Easiest to review changeset by changeset.

This should resolve #88


cmc333333:
Yay! Thanks for reviewing, @theresaanna !

PR #117, opened 2017-03-16
Attach the config service to the UI servers.

### LINKED ISSUE #77: Add authentication to UI pages

cmc333333:
Lest they won't see the password config.

Both demo and prod environments are configured, so this will close #77.


cmc333333:
Thanks @tadhg-ohiggins !

PR #119, opened 2017-03-16
Add ability to order requirements in API

tadhg-ohiggins:
This adds support for a `sort` parameter to the `/requirements/` endpoint.

@cmc333333 I wasn't able to make this play nicely with the code for sorting by number of matching keywords when using `OrderingFilter`, which is why this is custom code. It's possible I missed something. Hopefully the tests should still be useful even if I did.

### DIFF
--- a/reqs/tests/views_tests.py
+++ b/reqs/tests/views_tests.py
@@ -106,12 +106,12 @@ def test_requirements_ordered_by_multiple_keys(params, result):
 
 @pytest.mark.django_db
 @pytest.mark.parametrize('params', (
-    "sort=gibberish",
-    "sort=-gibberish",
-    "sort=-",
-    "sort=asdf",
-    "sort=policy__",
-    "sort=policy__gibberish",
+    "gibberish",
+    "-gibberish",
+    "-",
+    "asdf",
+    "policy__",
+    "policy__gibberish",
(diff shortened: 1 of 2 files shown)


reqs/tests/views_tests.py
```diff
@@ -54,3 +54,73 @@ def test_requirements_queryset_order():
...
+    ('sort=-req_id', (2, 1, 3), (10, 11, 12), ["3", "2", "1"]),
+    ('sort=policy__policy_number', (1, 2, 3), (20, 30, 10), ["3", "1", "2"]),
+    ('sort=-policy__policy_number', (1, 2, 3), (20, 30, 10), ["2", "1", "3"]),
+), ids=repr)
+def test_requirements_ordered_by_key(params, req_ids, policy_numbers, result):
+    """
+    We should be able to pass in arbitrary sort fields.
+    """
+    client = APIClient()
+    for i in range(len(req_ids)):
```
cmc333333: Super small, but it might be simpler to use `zip` and avoid `range`/`len`/`i`:
```
for req_id, policy_number in zip(req_ids, policy_numbers):
```

tadhg-ohiggins: 💯

PR #120, opened 2017-03-17
Add a policy search widget

cmc333333:
This generalizes the FilterList/Filters to remove ties to keywords, then configures a new set to work with policy titles. This does **not** need to be reviewed before today's demo.

Looks like:
screen shot 2017-03-17 at 10 28 36 am

### DIFF
--- a/ui/components/filter-list.jsx
+++ b/ui/components/filter-list.jsx
@@ -48,6 +48,7 @@ Filter.propTypes = {
  * this would need to modify to add/remove lookup values */
 export const searchParam = {
   keywords: 'keywords__id__in',
+  policies: 'policy_id__in',
 };
 
 export default function FilterList({ existingFilters, lookup, router }) {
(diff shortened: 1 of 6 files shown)


ui/components/filter-list.jsx
```diff
@@ -7,79 +7,93 @@ import React from 'react';
...
-      query: React.PropTypes.shape({
-        keywords__id__in: React.PropTypes.string,
-      }),
-    }),
+    location: React.PropTypes.shape({}),
   }),
 };
 
-export function fetchData({ location: { query } }) {
+export function fetchKeywords({ location: { query } }) {
```
theresaanna: Would it make sense to keep to maybe keep a `fetchData` and have it take care of the shared bits between these two functions, where `fetchPolicies` and `fetchKeywords` can pass `fetchData` their relevant individual values? If this is going to get reworked anyway, no worries. I can also see an argument to be made for this approach being clearer, so I just bring it up for consideration.

cmc333333: Do you think it'd make sense to encapsulate all of the API interactions into one module/class at this point? I've seen that in several of these frontend projects, but I don't have a strong feeling either way.

theresaanna: I do think it would make sense.

cmc333333:
Will pull out the API in a separate PR. Thanks @theresaanna !

PR #135, opened 2017-03-21
Fix a bug around sorting requirements

cmc333333:
To figure the bug out, I added Django Debug Toolbar; kept it in as it'll probably help us debug in the future.

Easiest to review changeset by changeset.


cmc333333:
Thanks @tadhg-ohiggins !

PR #138, opened 2017-03-22
Try to fix Excel's mangling of decimal values

tadhg-ohiggins:
How many problems /
Are due to Excel? The skies /
Refuse to answer.

Since the rows should be sequential, looking for prior rows with requirement IDs ending in `9` and trying to determine the current value seems like a reasonable approach.

### DIFF
--- a/reqs/management/commands/import_reqs.py
+++ b/reqs/management/commands/import_reqs.py
@@ -219,12 +219,32 @@ class RowProcessor:
         self.keywords = KeywordProcessor(keywords)
         self.connections = []
         self.req_ids = set()
+        self.last_id = "0.0"
+
+    def fix_excel_decimals(self, reqid):
+        """
+        Excel strips trailing zeroes, so for example what should be 1.20 is 1.2
+        in the data.
+        Since the rows are sequential (or should be!), we should be able to
+        account for this.
+        """
+        policy, requirement = reqid.split(".")
+        last_policy, last_requirement = self.last_id.split(".")
+        if policy == last_policy:
+            if last_requirement.endswith("9"):
+                sequential = str(int(last_requirement) + 1)
+                if sequential[:len(requirement)] == requirement:
+                    return "{0}.{1}".format(policy, sequential)
+        return reqid
 
     def add(self, row):
         req_key = find_key("req_ids", row)
         req_id = row[req_key]
         if req_id in ('None', ''):
             raise ValueError("Requirement without ID")
+        if "." not in req_id:
+            raise ValueError("Requirement without . separator")
+        req_id = self.fix_excel_decimals(req_id)
         if req_id in self.req_ids:
             raise ValueError("Duplicated Req ID: {0}".format(req_id))
         row = handle_transposed_reqstatus(row)
@@ -251,6 +271,7 @@ class RowProcessor:
             req_id=req_id, defaults=params)
         self.connections.extend(self.keywords.connections(row, req.pk))
         self.req_ids.add(req_id)
+        self.last_id = req_id
 
 
 class Command(BaseCommand):
--- a/reqs/management/commands/import_reqs.py
+++ b/reqs/management/commands/import_reqs.py
@@ -221,6 +221,16 @@ class RowProcessor:
         self.req_ids = set()
         self.last_id = "0.0"
 
+    def validate_requirement_id(self, reqid):
+        if reqid in ('None', ''):
+            raise ValueError("Requirement without ID")
+        if "." not in reqid:
+            raise ValueError("Requirement without . separator")
+        reqid = self.fix_excel_decimals(reqid)
+        if reqid in self.req_ids:
+            raise ValueError("Duplicated Req ID: {0}".format(reqid))
+        return reqid
+
     def fix_excel_decimals(self, reqid):
         """
         Excel strips trailing zeroes, so for example what should be 1.20 is 1.2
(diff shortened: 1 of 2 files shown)


reqs/management/commands/import_reqs.py
```diff
@@ -219,12 +219,32 @@ def __init__(self, keywords=None):
...
+        """
+        Excel strips trailing zeroes, so for example what should be 1.20 is 1.2
+        in the data.
+        Since the rows are sequential (or should be!), we should be able to
+        account for this.
+        """
+        policy, requirement = reqid.split(".")
+        last_policy, last_requirement = self.last_id.split(".")
+        if policy == last_policy:
+            if last_requirement.endswith("9"):
```
cmc333333: Minor: this code would be less pointy if we used `if policy... and last_req...:`

reqs/management/commands/import_reqs.py
```diff
@@ -219,12 +219,32 @@ def __init__(self, keywords=None):
...
+
+    def fix_excel_decimals(self, reqid):
+        """
+        Excel strips trailing zeroes, so for example what should be 1.20 is 1.2
+        in the data.
+        Since the rows are sequential (or should be!), we should be able to
+        account for this.
+        """
+        policy, requirement = reqid.split(".")
+        last_policy, last_requirement = self.last_id.split(".")
```
cmc333333: Would it make sense to store these as a tuple of ints instead of re-splitting them?

reqs/management/commands/import_reqs.py
```diff
@@ -251,6 +271,7 @@ def add(self, row):
             req_id=req_id, defaults=params)
         self.connections.extend(self.keywords.connections(row, req.pk))
         self.req_ids.add(req_id)
+        self.last_id = req_id
```
cmc333333: It'd be nice if all of the logic around `last_id` were maintained in one place -- does it make sense to set it within `fix_excel_decimals`?

reqs/management/commands/import_reqs.py
```diff
@@ -219,12 +219,32 @@ def __init__(self, keywords=None):
...
+                    return "{0}.{1}".format(policy, sequential)
+        return reqid
 
     def add(self, row):
         req_key = find_key("req_ids", row)
         req_id = row[req_key]
         if req_id in ('None', ''):
             raise ValueError("Requirement without ID")
+        if "." not in req_id:
+            raise ValueError("Requirement without . separator")
```
cmc333333: Should this check happen inside `fix_excel_decimals`? Just wondering if we can keep all that logic in one place.

reqs/management/commands/import_reqs.py
```diff
@@ -219,12 +219,32 @@ def __init__(self, keywords=None):
...
+        in the data.
+        Since the rows are sequential (or should be!), we should be able to
+        account for this.
+        """
+        policy, requirement = reqid.split(".")
+        last_policy, last_requirement = self.last_id.split(".")
+        if policy == last_policy:
+            if last_requirement.endswith("9"):
+                sequential = str(int(last_requirement) + 1)
+                if sequential[:len(requirement)] == requirement:
```
cmc333333: This feels risky due to dropping data. What about something like:
```python
if sequential.rstrip('0') == requirement:
```
The situation I'm imagining is something like `1.234569` followed by `1.23`, which shouldn't come up, but I don't think we need to risk it.

reqs/tests/import_reqs_tests.py
```diff
@@ -313,3 +313,32 @@ def test_keyword_normalization(blank_csv_file, keyword, expected):
...
 def test_handle_transposed_reqstatus(test_input, expected):
     assert import_reqs.handle_transposed_reqstatus(test_input) == expected
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize("ids,expected", (
+    (["1.01", "1.02"], ["1.01", "1.02"]),
+    (["1.09", "1.1"], ["1.09", "1.10"]),
+    (["1.59", "1.6"], ["1.59", "1.60"]),
+    (["1.99", "1.1"], ["1.100", "1.99"]),  # Django order_by sort issue.
```
cmc333333: Ah, yeah, we'll probably want to convert these to int fields at some point.

reqs/management/commands/import_reqs.py
tadhg-ohiggins: Possibly, but it made more sense to me to store it in its native form.

reqs/management/commands/import_reqs.py
tadhg-ohiggins: 👍

reqs/management/commands/import_reqs.py
tadhg-ohiggins: Good point; I added this to the tests and used your approach.

reqs/management/commands/import_reqs.py
tadhg-ohiggins: Good point; I created a validation method that I moved all this to.

reqs/management/commands/import_reqs.py
tadhg-ohiggins: It makes more sense to me to set it only after it's been successfully inserted, as with adding to `self.req_ids`.

reqs/tests/import_reqs_tests.py
```diff
@@ -342,3 +343,20 @@ def test_fix_excel_decimals(blank_csv_file, ids, expected):
     assert len(reqs) == 2
     assert reqs[0].req_id == expected[0]
     assert reqs[1].req_id == expected[1]
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize("requirement_id", ("", "100", "0" "1,000"), ids=repr)
+@pytest.mark.xfail(raises=ValueError)
```
cmc333333: Interesting -- this _does_ include the test results differently from the rest of the results. Any reason to use this over `pytest.raises`?

reqs/management/commands/import_reqs.py
```diff
@@ -221,6 +221,16 @@ def __init__(self, keywords=None):
         self.req_ids = set()
         self.last_id = "0.0"
 
+    def validate_requirement_id(self, reqid):
```
cmc333333: Excellent idea

reqs/tests/import_reqs_tests.py
tadhg-ohiggins: Nope, I was just experimenting with the decorator. I'll switch it to the context manager for consistency.

PR #149, opened 2017-03-23
Move API-related functions into a single module.

cmc333333:
This should also create a nicer interface for calling API endpoints.


cmc333333:
Thanks @theresaanna !

PR #150, opened 2017-03-23
Adds sunset date to requirements and breaks newlines in requirement text

theresaanna:
As described in the title:
- If there is a sunset date associated with a requirement, it will be displayed.
- Where there are newlines in the requirement text, break the requirement text up with `<br />`s.

On this last point, it adds `<span>`s around each line of a requirement because there needs to be a wrapper tag. I am open to ways around this.
Also, not all logical breaks are made with this method. I have to dig into this to figure out whether any sort of break exists in the req text and what it is. This seems to capture at least 50% of cases, though.


ui/components/requirements.jsx
```diff
@@ -47,7 +49,12 @@ function Requirements({ keywords, pagedReqs, policies, router }) {
         </div>
         <ul className="list-reset">
           { pagedReqs.results.map(requirement =>
-            <Requirement key={requirement.req_id} requirement={requirement} />) }
+            <Requirement
+              key={requirement.req_id}
+              requirement={requirement}
+              sunset={requirement.policy.sunset}
```
cmc333333: It looks like we're accessing the sunset date via `requirement` in the `Requirement` component; does that mean we don't need to pass it in separately here?

ui/components/requirements.jsx
```diff
@@ -11,16 +11,18 @@ function Requirement({ requirement }) {
...
       <div className="req-id col col-1 mb2">
-        {requirement.req_id}
+        { requirement.req_id }
       </div>
-      <div className="req-text col col-11">
-        {requirement.req_text}
+      <div className="req-text col col-11 pl1">
+        { requirement.req_text.split('\n').map(line => (
+          <span className="req-text-line mb1">{ line }<br /></span>
+          ))}
```
cmc333333: Nice solution! It might make sense to pull this out into a utility function if we end up printing the text out in more places.

ui/components/requirements.jsx
```diff
@@ -66,6 +73,9 @@ Requirements.propTypes = {
   keywords: FilterList.propTypes.existingFilters,
   pagedReqs: React.PropTypes.shape({
     results: React.PropTypes.arrayOf(React.PropTypes.shape({
+      policy: React.PropTypes.shape({
+        sunset: React.PropTypes.string,
+      }),
```
cmc333333: If we remove the `sunset` attribute, I _think_ we can remove this type check, too. It might make sense to just reference `Requirement.propTypes.requirement` (would require re-ordering the file, though).

PR #152, opened 2017-03-24
Add by-policy tab

### LINKED ISSUE #123: Add functional by-policy tab
In the second tab of the requirements UI, _visually_ group together (i.e. by sorting) requirements into policies. Sketch for this is

cmc333333:
Building on #149, this splits up the Requirements component into:
* Container (which includes the left-side filters)
* Tabs
* ByKeyword
* ByPolicy

Requirements always include the Container and Tabs, but then, depending on path, include either ByKeyword or ByPolicy.

Looks like:

screen shot 2017-03-24 at 5 56 33 pm

screen shot 2017-03-24 at 5 56 44 pm

Should resolve #123

### DIFF
--- a/ui/components/requirements/tabs.jsx
+++ b/ui/components/requirements/tabs.jsx
@@ -1,0 +1,70 @@
+import React from 'react';
+import { Link } from 'react-router';
+
+function Tab({ active, link, tabName }) {
+  if (active) {
+    return <li className="inline-block mr4 bold">{ tabName }</li>;
+  }
+  return (
+    <li className="inline-block mr4">
+      <Link to={link} >{ tabName }</Link>
+    </li>
+  );
+}
+Tab.defaultProps = {
+  active: false,
+  link: { pathname: '', query: {} },
+  tabName: '',
+};
+Tab.propTypes = {
+  active: React.PropTypes.bool,
+  link: React.PropTypes.shape({
+    pathname: React.PropTypes.string,
+    query: React.PropTypes.shape({}),
+  }),
+  tabName: React.PropTypes.string,
+};
+
+export default function Tabs({ router }) {
+  if (router.routes.length < 2) {
+    return null;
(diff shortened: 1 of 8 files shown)


ui/components/requirements/tabs.jsx
```diff
@@ -0,0 +1,70 @@
...
+      pathname: `/${pathprefix.concat([childRoute.path]).join('/')}`,
+      query: modifiedQuery,
+    },
+  }));
+
+  return (
+    <div>
+      <span className="mr4">Organize by</span>
+      <ul className="list-reset inline-block">
+        { tabs.map(tab => <Tab {...tab} />) }
```
theresaanna: Can you explain `{...tab}`? I haven't seen this before.

ui/components/requirements/tabs.jsx
```diff
@@ -0,0 +1,70 @@
...
+Tab.propTypes = {
+  active: React.PropTypes.bool,
+  link: React.PropTypes.shape({
+    pathname: React.PropTypes.string,
+    query: React.PropTypes.shape({}),
+  }),
+  tabName: React.PropTypes.string,
+};
+
+export default function Tabs({ router }) {
```
cmc333333: From my understanding, functions are preferred to class based components whenever we aren't tracking any state (which is the bulk of our components). AirBnB's lint configuration will actually yell at you if you don't follow this strategy.

ui/components/requirements/tabs.jsx
cmc333333: That's a [JSX-specific](https://facebook.github.io/react/docs/jsx-in-depth.html#spread-attributes) version of the [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_operator), which more or less says "treat this object like like a mapping between attributes and values".

PR #156, opened 2017-03-27
Add New Relic support to both Python and Node apps

### LINKED ISSUE #151: Add New Relic
In particular, we want ping monitoring.

cmc333333:
The credential services are already configured and pinging setup, so this should resolve #151.


cmc333333:
Thanks @theresaanna @tadhg-ohiggins

PR #159, opened 2017-03-28
Replace autocomplete libraries

### LINKED ISSUE #136: Fix issue with commas in keywords in Django admin view
The Django admin keyword-handling currently breaks when given a keyword with commas in it, with “IT Transparency (Open Data, FOIA, Public Records, etc.)” resulting in this:
image

We use <https://bitbucket.org/fabian/django-taggit-autosuggest> for that functionality, which appears to pull in <https://github.com/hlsolutions/jquery-autosuggest/>, and I think the problem is related to this line: <https://github.com/hlsolutions/jquery-autosuggest/blob/master/src/jquery.autoSuggest.coffee#L655>

I'm not sure yet how to fix this.

cmc333333:
This should resolve #136. It _did_ require a bit of custom code (which we might be able to contribute back).


reqs/admin.py
```diff
@@ -1,7 +1,31 @@
...
+
+
+class TaggitWidget(TaggitSelect2):
+    """Account for commas in tags by wrapping each entry in double quotes"""
+    def value_from_datadict(self, data, files, name):
+        values = data.getlist(name)
+        values = [v.replace('"', '') for v in values]   # remove double quotes
+        values = ['"{0}"'.format(v) for v in values]
+        return ','.join(values)
+
```
tadhg-ohiggins: I can imagine a tag where the quotation marks are important; should we do something like this?
```
def handle_quotation_marks(value):
    """Account for commas and quotation marks in tags."""
    num_marks = value.count('"')
    if num_marks % 2 != 0:
        value = value.replace('"', '')
    else:
        while '"' in value:
            marks = ("“", "”")
            value = value.replace('"', marks[value.count('"') % 2], 1)

    return '"{0}"'.format(value)
```
(I have this + tests locally and can add it if you think this is a reasonable idea.)

cmc333333: That seems like a nice solution. Let's do it!

cmc333333:
Thanks @tadhg-ohiggins !

PR #160, opened 2017-03-28
Adds styling for filter items in left hand sidebar

theresaanna:
Implements: 
screen shot 2017-03-29 at 11 10 15 am


ui/assets/css/_req-filter-ui.scss
```diff
@@ -1,11 +1,15 @@
...
 
-  li {
-    /* Handing indent */
-    padding-left: 15px;
-    text-indent: -15px;
+  .active-filter {
+    background-color: $color-gray-warm-light;
+    font-weight: bold;
+
+    .remove-filter-link {
```
cmc333333: I've been warned pretty strongly against nesting selectors too deep. Can this be more general purpose?

theresaanna: It may be made more general purpose. I think that'll become clearer as the designs and style guide start firming up. My personal feeling is that once we start doing the same types of things over and over, that's when it's time to refactor into mixins. I'm not terribly worried about nesting as long as we keep the CSS files small - no longer than a couple page scrolls. AFAIK there are no performance concerns, just maintainability and duplication ones. I can remove the top level `req-filter-ui` definition to remove one layer of nesting.

PR #162, opened 2017-03-28
Initial navbar w/ usa header

donjo:
will implement a 2 part header component (official USA site disclaimer separate from main header)

todo:

- [ ] finish main header + add logo
- [ ] move colors to global sass vars (or use them if they already exist)


ui/assets/css/_header.scss
```diff
@@ -0,0 +1,19 @@
+.usa-disclaimer {
+  background-color: $color-gray-lightest;
+  color: $color-gray-dark;
+  font-size: 12px;
```
cmc333333: Relatedly -- does it make sense to pull these out into more generic mixins? I assume we'll want to reuse this font config elsewhere.

donjo: @cmc333333 for this one in particular I don't know if we will use it anywhere else but I think we should definitely do that for some of the UI text. 

@theresaanna I'll make a separate task to establish a baseline and figure on what kind of flexible unit we want to go with

ui/assets/css/_header.scss
```diff
@@ -0,0 +1,19 @@
...
+  padding: 0.25rem 1rem;
+  line-height: 15px;
+}
+
+.navbar {
+  background-color: $color-primary-darkest;
+  height: 80px;
```
donjo: Yeah possibly, let's see what @micahtaylor thinks. I haven't put any mobile thought into this yet since it still looks the same at narrower resolutions but if we ever have to add anything to the right side of the nav it would change.

PR #164, opened 2017-03-29
Update which fields are required.

### LINKED ISSUE #158: Define required fields
For requirements, we're currently requiring the following fields be non-blank:
* Req id
* Issuing body
* Policy Section
* Policy Sub section
* Req text
* Verb
* Impacted entity
* Req deadline
* Citation

Validate this list and change as needed.

cmc333333:
Adding policy to the list of required fields (I'm not sure why this
wasn't the case already). This assumes no one's changed a requirement to have no policy.

Made Policy Section, Sub Section, Verb, Impacted Entity, Req Deadline,
Citation blank-able.

Should resolve #158


cmc333333:
Thanks @tadhg-ohiggins

PR #165, opened 2017-03-29
Tweaks for 2017-03-28 data file

tadhg-ohiggins:
The sun in the sky, /
The ocean deeps, every wind: /
They all despair at Excel.

Address duplicate requirement IDs by appending zeroes to them; other tweaks to handle the latest data file.


reqs/management/commands/import_reqs.py
```diff
@@ -238,12 +239,22 @@ def fix_excel_decimals(self, reqid):
         Since the rows are sequential (or should be!), we should be able to
         account for this.
         """
+        while reqid.endswith("."):  # Handle a hack around Excel problems
+            reqid = reqid[:-1]
```
cmc333333: Super minor: Would
```python
reqid = reqid.rstrip('.')
```
be cleaner?

reqs/tests/import_reqs_tests.py
```diff
@@ -81,7 +81,7 @@ def test_imports_correctly2(csv_file2):
     call_command('import_reqs', str(csv_file2))
 
     reqs = list(Requirement.objects.order_by('req_id'))
-    assert len(reqs) == 2
+    assert len(reqs) == 3
```
cmc333333: Minor -- should we be checking any information about the middle entry?

reqs/management/commands/import_reqs.py
```diff
@@ -238,12 +239,22 @@ def fix_excel_decimals(self, reqid):
...
+        if "." not in reqid:
+            raise ValueError(
+                "Requirement ID without . separator: {0}".format(reqid))
         policy, requirement = reqid.split(".")
         last_policy, last_requirement = self.last_id.split(".")
         if policy == last_policy and last_requirement.endswith("9"):
             sequential = str(int(last_requirement) + 1)
             if sequential.rstrip("0") == requirement:
-                return "{0}.{1}".format(policy, sequential)
+                reqid = "{0}.{1}".format(policy, sequential)
```
cmc333333: Do we need this logic (figuring out the sequent, checking if we're in sequence, saving the last req id, etc.) at all anymore?

tadhg-ohiggins: I think we still need it. If we have the following IDs:
1.08, 1.09, 1.1, 1.11
then the sequential logic will correctly change the third one to `1.10`. If we remove it, it'll get imported as `1.1`, which won't be a duplicate, but will need correcting later.

reqs/tests/import_reqs_tests.py
tadhg-ohiggins: I don't think we need to add that in this test; this was a hack around all three rows now getting imported since we're now not rejecting duplicates.

PR #167, opened 2017-03-29
Fix autocomplete bug

cmc333333:
We moved around some functions recently and subtly broke autocomplete in the
process. This fixes that issue and adds tests to catch similar issues in the
future.


cmc333333:
Thanks @theresaanna !

PR #173, opened 2017-03-30
Tweaks to admin

cmc333333:
Three quick wins that OMB's requested:
* Remove taggit's Tag model from the admin
* Add a search box for filtering models. Super naive approach, but this works
* Add OMB policy id to policy names


cmc333333:
Thanks @tadhg-ohiggins !

PR #175, opened 2017-03-30
Adds keywords to requirements

theresaanna:
For each requirement, the list of associated keywords is shown. This does not link the keywords. 

@cmc333333, I'd like to work together on linking them if you have time. My best thinking involves fetching the `keywords` endpoint and comparing strings to get the appropriate ID. I wonder if there's not a more elegant solution, given that this information is already in the app.

### DIFF
--- a/ui/components/requirements/requirement.jsx
+++ b/ui/components/requirements/requirement.jsx
@@ -18,14 +18,11 @@ export default function Requirement({ requirement }) {
             Sunset date by { requirement.policy.sunset || 'none' }
           </div>
           <div className="topics">
-            <ul className="topics-list list-reset">
-              <li className="inline">Topics: </li>
+            <span>Topics: </span>
+            <ul className="topics-list list-reset inline">
               { requirement.keywords.map((keyword, index, keywords) => (
-                <li className="inline">
+                <li key={keyword} className="inline">
                   { keyword }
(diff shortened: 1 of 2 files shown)


ui/components/requirements/requirement.jsx
```diff
@@ -17,6 +17,19 @@ export default function Requirement({ requirement }) {
           <div className="sunset-date">
             Sunset date by { requirement.policy.sunset || 'none' }
           </div>
+          <div className="topics">
+            <ul className="topics-list list-reset">
+              <li className="inline">Topics: </li>
```
cmc333333: Should this header be an `li` or should it be outside the `ul`?

ui/components/requirements/requirement.jsx
```diff
@@ -17,6 +17,19 @@ export default function Requirement({ requirement }) {
...
+              { requirement.keywords.map((keyword, index, keywords) => (
+                <li className="inline">
+                  { keyword }
+                  { index !== keywords.length - 1 &&
+                  <span>, </span>
+                  }
```
cmc333333: This threw me for a loop until I saw it in action. What do you think about adding the commas in css instead? That'd simplify this logic a lot as we wouldn't need to track the index at all.

Unrelated: does the `li` need to have a key?

PR #176, opened 2017-03-30
Add django-reversion to admin view, and a basic test for it.

tadhg-ohiggins:
That which was before /
Can be again, but the rain /
Never falls upwards.

Add tracking of version history for changes made to objects in the admin view, with the ability to both see the history of a given object and revert to prior states.

### DIFF
--- a/devops/dev_api.sh
+++ b/devops/dev_api.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
 
 ./manage.py migrate --noinput
+./manage.py createinitialrevisions
 ./manage.py runserver 0.0.0.0:$PORT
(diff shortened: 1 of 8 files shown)


devops/dev_api.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
 
 ./manage.py migrate --noinput
+./manage.py createinitialrevisions
```
tadhg-ohiggins: I'm not sure this is the correct place to put this, but note that it's non-destructive; if the revisions are already present it doesn't do anything. (This also applies to the same line in `prod_api.sh`.)

cmc333333: If running this is non-destructive and it doesn't take too long in the common case, I think this is a fine place.

PR #191, opened 2017-04-03
Add Code Climate JS test running to Travis

theresaanna:
With this PR, `npm test` runs `jest` to generate test coverage data. On Travis, it'll send to Code Climate. Also makes indentation consistent in `.travis.yml`.


.travis.yml
```diff
@@ -16,16 +16,20 @@ script:
...
+- mv autopilot-linux $HOME
+- cf install-plugin -f ~/autopilot-linux
 deploy:
   provider: script
   script: devops/deploy.sh dev
   skip_cleanup: true
   on:
     branch: master
+after_success:
+- docker-compose run --rm dev npm install codeclimate-test-reporter
```
cmc333333: I'm not sure if this makes things more complicated, but we have a docker-compose command/service for `npm`, meaning these two commands are almost identical:
```
docker-compose run --rm npm install codeclimate-test-reporter
docker-compose run --rm dev npm install codeclimate-test-reporter
```

The difference is that the first runs `install codeclimate-test-reporter` in the `npm` container (which prefixes everything with `npm `) while the second runs `npm install codeclimate-test-reporter` in the `dev` container (which doesn't have any command prefix).

No requested changes, just hoping to clear up any confusion.

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "This repository contains code necessary to run the White House Office of Management and Budget (OMB) instance of [eRegulations](https://eregs.github.io) (a regulation parser, API, and viewer).",
   "main": "index.js",
   "scripts": {
-    "test": "jest",
+    "test": "jest --coverage",
```
cmc333333: Do we need the flag here if `collectCoverage` is specified in the `jest` config?

theresaanna: We do not!

.travis.yml
theresaanna: Thank you for that!

PR #192, opened 2017-04-03
Don't specify node minor version.

cmc333333:
Cloud Foundry buildpacks might drop support for a minor version; 6.x is the
LTS.


cmc333333:
Thanks @theresaanna !

PR #196, opened 2017-04-04
Use MAX Login on cloud.gov

### LINKED ISSUE #130: Max integration -- Code back end

cmc333333:
Connects to MAX for login needs. For the most part the django-cas-ng library does what we need; the exception is that MAX usernames are meaningless, so we override that bit to treat the email address as a username.

To emphasize, this strategy requires that we create user accounts _ahead_ of their first login; we're using MAX for authentication not authorization. I've create user accounts on the demo server for everyone on the 18F side.

This requires a `docker-compose build` if running Python tests; shouldn't need it otherwise. Should resolve #130

### DIFF
--- a/omb_eregs/max_backend.py
+++ b/omb_eregs/max_backend.py
@@ -16,6 +16,7 @@ class MAXBackend(CASBackend):
 
         user = User.objects.filter(username=email).first()
         if user:
+            # Trim to thirty chars to match Django User max lengths
             user.first_name = attributes.get('First-Name', '')[:30]
             user.last_name = attributes.get('Last-Name', '')[:30]
             user.email = email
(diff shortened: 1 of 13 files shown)


omb_eregs/max_backend.py
```diff
@@ -0,0 +1,25 @@
...
+        max_username, attributes, _ = client.verify_ticket(ticket)
+        if not max_username:    # bad ticket
+            return None
+
+        email = attributes.get('Email-Address', '')
+
+        user = User.objects.filter(username=email).first()
+        if user:
+            user.first_name = attributes.get('First-Name', '')[:30]
+            user.last_name = attributes.get('Last-Name', '')[:30]
```
tadhg-ohiggins: Would it be worth noting here that the truncation is due to the limits for those fields on the Django auth user model?

cmc333333: Good idea, will add

Issue #181, opened 2017-04-01
Link topics under requirements


theresaanna:
Implement the actual links on each topic in the list under each requirement.

PR #200, opened 2017-04-05
Include keyword ids in nested models of Requirements API

cmc333333:
This is needed for #181 (so we know which keyword id was clicked)


cmc333333:
Thanks @tadhg-ohiggins

Issue #166, opened 2017-03-29
Establish root font sizes


donjo:
establish baseline font sizes so we can `rem` and `em` appropriately in our sass files.

theresaanna:
@donjo My approach currently is to use `em`s, at the default 16px = 1em. I don't know much about `rem`. Do you feel we should do anything more comprehensive than this?

donjo:
@theresaanna I looked into this more and I think the only reason I thought it was "broke" was because I was using to working with 10px as a base on other projects. 16px is fine now that I know what I'm working with.

FWIW Basscss uses `rem` throughout most of the sizing properties they offer. There's a time and place for both; `em` uses the parent element to calculate size while `rem` uses the document root. If I were to pick a random piece of text on the page and make it `1em` it would be 15.008px because of the following rule.

```css
body {
   font-size: 0.938em; 
}
```

but if I set it to `1rem` it would be 16px because we didn't override font-size in `html {}` so it falls back to the browser default which is 16px almost all of the time. 


(I don't know if anything actionable has to come out of this, you can close it out)

theresaanna:
@donjo thank you for that. I would think we actually want to be using `rem` most of the time, then. Thank you for reminding me - I'd known the difference long ago but had forgotten. I'm going to create another ticket to do that and close out this one.

PR #206, opened 2017-04-06
Adds styling for filter form when there's no JS

theresaanna:
Also sets the root font size to 16px, meaning 1rem = 16px;

screen shot 2017-04-06 at 11 42 57 am


theresaanna:
@cmc333333 I probably should have referenced this conversation: https://github.com/18F/omb-eregs/issues/166

PR #208, opened 2017-04-07
Docker refactor and initial integration tests

### LINKED ISSUE #187: Integration tests

cmc333333:
Best reviewed changeset by changeset; should resolve #187. This PR has two major functions which became intertwined.

First, it refactors our docker setup to avoid generating an image for the Python builds. Both the Node app and the Django app will store their libraries on a shared volume, meaning we don't need to rebake an image after every library change. This _does_ impact performance, but it's still decent and will [hopefully improve soon](https://github.com/docker/for-mac/issues/77#issuecomment-292192768). In the process, startup scripts were added for both the Python and Node apps to synchronize the user's environment with the requisite libraries automatically. This means we don't need to run `docker-compose build` nor `docker-compose run --rm npm install`.

Second, it adds a few integration tests, which drive chrome and firefox around and verify that a few elements are present on the screen. These tests can be ran relatively easily via `./devops/integration-tests.sh`.


cmc333333:

PR #209, opened 2017-04-10
Display more metadata with requirements

tadhg-ohiggins:
More information, /
More properties, more info. /
Surely thus better.

Adds:

- Issuing body
- Policy title
- Policy issuance date
- OMB Policy ID

Also displays “Sunset date: none” if there's no sunset date.

Other than OMB Policy ID, these are all arbitrary additions, and not my idea of what the essential properties are.

### DIFF
--- a/ui/components/requirements/requirement.jsx
+++ b/ui/components/requirements/requirement.jsx
@@ -52,8 +52,8 @@ export default function Requirement({ requirement }) {
           <Metadata
             className="applies-to mr2"
             name="Applies to"
-            value=""
-            nullValue="Applies to: [not implemented]"
+            value={requirement.impacted_entity}
+            nullValue="Applies to: unknown"
           />
           <Metadata
             className="issuing-body"
--- a/ui/components/requirements/requirement.jsx
+++ b/ui/components/requirements/requirement.jsx
@@ -1,5 +1,39 @@
 import React from 'react';
 
+export function Metadata({ className, name, value, nullValue, separator }) {
+  /**
+   * Returns a div element that has className equal to the argument plus 'metadata', and content
+   * that is either <name><separator><value> or, if value is null and nullValue is not, the content
+   * is the nullValue argument.
+   *
+   * If value and nullValue are both null, returns null.
+   */
+  if (!value && !nullValue) {
+    return null;
+  }
+  const classes = className.split(' ');
+  const classString = classes.includes('metadata') ? className : `${classes.join(' ')} metadata`;
+  const content = value ? `${name}${separator}${value}` : nullValue;
+  return (
+    <div className={classString}>
+      {`${content}`}
+    </div>);
+}
+
+Metadata.propTypes = {
+  className: React.PropTypes.string.isRequired,
+  name: React.PropTypes.string.isRequired,
+  value: React.PropTypes.string.isRequired,
+  nullValue: React.PropTypes.string,
+  separator: React.PropTypes.string,
+};
+
+Metadata.defaultProps = {
+  nullValue: '',
+  separator: ': ',
+};
+
+
 export default function Requirement({ requirement }) {
   // We could have multiple lines with the same text, so can't use a stable ID
   /* eslint-disable react/no-array-index-key */
@@ -7,7 +41,6 @@ export default function Requirement({ requirement }) {
     <span key={idx} className="req-text-line mb1">{ line }<br /></span>
   ));
   /* eslint-enable react/no-array-index-key */
-  const sunsetString = requirement.policy.sunset ? 'Sunset date by $requirement.policy.sunset' : 'Sunset date: none';
   return (
     <div className="req p2 clearfix max-width-3">
       <div className="req-id col col-1 mb2 mr1">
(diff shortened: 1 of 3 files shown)


ui/components/requirements/requirement.jsx
```diff
@@ -7,6 +7,7 @@ export default function Requirement({ requirement }) {
     <span key={idx} className="req-text-line mb1">{ line }<br /></span>
   ));
   /* eslint-enable react/no-array-index-key */
+  const sunsetString = requirement.policy.sunset ? 'Sunset date by $requirement.policy.sunset' : 'Sunset date: none';
```
cmc333333: I think you'd like to use string interpolation here:
```javascript
`Sunset date by ${requirement.policy.sunset}`
```

It'd be worthwhile to add a unit test here -- hopefully the examples in `ui/__test__` explain what do to, but we can dive in deeper if you'd like.

ui/components/requirements/requirement.jsx
```diff
@@ -15,13 +16,37 @@ export default function Requirement({ requirement }) {
...
           </div>
-          <div className="sunset-date">
-            Sunset date by { requirement.policy.sunset || 'none' }
+          { requirement.issuing_body ? (
+            <div className="issuing-body metadata">
+              Issuing body: { requirement.issuing_body }
+            </div>
+          ) : (
+            ''
+          ) }
```
cmc333333: Generally, we don't want this kind of logic in the template as it makes them harder to read. Mind moving this into a constant instead? Also, `null` is the more generally way to say "nothing here".

tadhg-ohiggins: That makes sense, but I'm unclear on the right way to do this—it seems slightly odd for a const to have a full element in it. Is there an equivalent here to the Python approach of something like this?

```
element = """
<div className="{0} metadata">
    {1}: {2}
</div>
"""

element.format("issuance", "Policy issuance", requirement.policy.omb_policy_id)
```

ui/components/requirements/requirement.jsx
tadhg-ohiggins: Hmm, I thought that was the right way to do string interpolation in this context, will investigate further.

Tests are good, I will add some.

ui/components/requirements/requirement.jsx
cmc333333: Almost! JSX is creating objects behind the scenes, so we can't do straight string interpolation, but we could do:
```jsx
function Metadata({ className, key, value }) {
    if (!value) {
        return null;
    }
    return (
        <div className={`${className} metadata`} >
            {`${key} : ${value}`}
        </div>);
}
....
<someContainer>
    <Metadata className="issuance" key="Policy issuance" value={requirement.policy.omby_policy_id} />
</someContainer>
```

donjo:
This looks good so far. I don't think there are any pieces of metadata we'd like to utilize that aren't included. We may make some cuts/move things around as the design gets more refined but just having the stuff in the component makes it easy to iterate upon.

ui/components/requirements/requirement.jsx
tadhg-ohiggins: Woo destructuring assignment!

tadhg-ohiggins: Just going to note this here: `key` turns out to be some kind of secretly-hidden reserved word—if you try code similar to the above using an argument named `key` its value will show up in the output as `undefined`. If you change `key` to `name` it all works fine.

(This is true on my machine, anyway…)

cmc333333: Ugh, sorry about that. React uses `key` to indicate unique ids when it needs to figure out what children to update.

ui/components/requirements/requirement.jsx
theresaanna: @tadhg-ohiggins the way CM pointed out is a fancy new ECMAScript 6 way of doing it. They're called template literals https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals.

ui/components/requirements/requirement.jsx
theresaanna: @tadhg-ohiggins @cmc333333 I'm wondering if we could use React's conditional rendering to make this a little cleaner? https://facebook.github.io/react/docs/conditional-rendering.html
It may be overkill, but maybe we have components for each piece of metadata that render conditionally based on the presence of the value we're checking in these statements.

tadhg-ohiggins:
I rewrote a bunch of this and added tests, so it likely needs re-review.

ui/components/requirements/requirement.jsx
```diff
@@ -15,13 +49,41 @@ export default function Requirement({ requirement }) {
...
-          </div>
-          <div className="sunset-date">
-            Sunset date by { requirement.policy.sunset || 'none' }
-          </div>
-          <div className="topics">
+          <Metadata
+            className="applies-to mr2"
+            name="Applies to"
+            value=""
+            nullValue="Applies to: [not implemented]"
```
cmc333333: Mind adding `requirement.impacted_entity` here? It'll at least get that ball rolling.

tadhg-ohiggins: Sure, happy to. I thought we were keeping that out on purpose.

PR #223, opened 2017-04-12
Add beta label to header

theresaanna:
Adds a "beta" label to the header, in compliance with ATO requirements.


donjo:
@theresaanna I'd maybe throw a space in between Policy and Beta, or if you want to go the padding route, something like `0.2rem` on the `<sup>`

PR #225, opened 2017-04-13
Pass API 404's to the user.

cmc333333:
This resolves two issues ZAP found:
* using a large page number gives a 404 from the API, but had been 500ing the UI
* cookies should only be sent over SSL (when SSL is available)

~~I think it makes the most sense to write an integration test once #208's merged.~~


cmc333333:
Thanks @tadhg-ohiggins!

PR #229, opened 2017-04-13
Make keywords (within a requirement) a link.

### LINKED ISSUE #181: Link topics under requirements

cmc333333:
With test. Should resolve #181


cmc333333:
Thanks @tadhg-ohiggins

PR #230, opened 2017-04-13
Include DAP for page tracking.

cmc333333:
We need to explicitly trigger a pageview when the user loads another page,
too.

This doesn't attempt to resolve the link usage drama mentioned here:
https://github.com/digital-analytics-program/gov-wide-code/pull/47 but should
be enough to get us started.


cmc333333:
Thanks @theresaanna

Issue #231, opened 2017-04-14
Separate components from logic in ui


cmc333333:
At the moment, almost all of our logic lives inside the components JSX files. Components indicate which AJAX calls need to be made to populate themselves.

As pointed out during one of our react reviews, this isn't a good long term solution. It
* Makes editing templates very painful
* Conversely, makes finding logic cumbersome
* Can lead to inefficiencies around API calls (the same data may be requested multiple times by different components).

The proposal is to follow an MVC-style pattern where components aka views are focused on markup, containers (using the language of react-redux) aka controllers are added to handle the conversion from location to data necessary for the components. This might involve reworking our react-resolver calls to only be on top-most containers. This *might* involve inserting redux or implementing something similar to keep the application's constant state.

cmc333333:
Merged

PR #254, opened 2017-04-18
Add an API cache to the frontend.

cmc333333:
A short-term solution for #231, this sets up a per-endpoint
least-recently-used cache to prevent repeated API calls from being made.


cmc333333:
Thanks @theresaanna, I took lots of cues from the eRegs design :)

PR #255, opened 2017-04-18
Changing word and running tests.

micahtaylor:


tadhg-ohiggins:
They've always been called topics, and we've always been at war with Eastasia.

cmc333333:
Thanks @tadhg-ohiggins !

PR #256, opened 2017-04-18
Search for topics when no input is entered.

### LINKED ISSUE #238: Show some search results before users start searching Topics
Why: Showing the first few items in the list before they are filtered down by a substring search helps users understand what they can be searching instead of having to guess which words may be a topic. Also provides users with a way to browse topics until we have a homepage in place.

This should just be a configuration on the ajax search widget to show results when there's zero-length input.

cmc333333:
We had incorrectly guessed that we didn't want to account for empty searches.

Should resolve #238


cmc333333:
Thanks @tadhg-ohiggins

PR #257, opened 2017-04-18
Update System diagram.

### LINKED ISSUE #234: Update system diagram to use naming/placement of Calc's
We want the IT people to say "this looks like something I've seen before" and notice the differences quickly

cmc333333:
Should close #234

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,10 +213,7 @@ And deploy!
 ```
 
 ## System diagram
-OMB eRegs Diagram Agency/Public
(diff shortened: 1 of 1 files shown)


cmc333333:
Please hold on merging until we get more feedback from Holly

cmc333333:
Okay, should be good for review now!

cmc333333:
Thanks @tadhg-ohiggins!

Issue #233, opened 2017-04-17
Use Travis to deploy prod


cmc333333:
* Create a "prod" or "release" etc. branch. Look at other 18F repos for naming
* Create a deploy user in cloud.gov prod space
* Modify deploy scripts to use different auth depending on dev or prod
* Modify travis script to zero-downtime deploy to prod on changes to that branch
* Add a nightly deploy cron job to travis

cmc333333:
Based on #259, that last step is quite simple -- there's no configuration file or anything, we just tell Travis what branch to build from and how often (daily, weekly, monthly) and it does.

cmc333333:
Re naming, it looks like "release" is the way to go. Identity is using `stages/release`, FEC uses `release/xxxx` (before hitting master), DATA Act uses `master`; the others I skimmed didn't have a separate release branch.

cmc333333:
This is all working, though it revealed #264

PR #261, opened 2017-04-19
Deploy to prod from Travis

cmc333333:
When there are changes to a as-of-yet-non-existent "release" branch, run a deploy script for production.

Part of #233


cmc333333:
Thanks @tadhg-ohiggins 🤞

PR #263, opened 2017-04-20
Log auth events to console

### LINKED ISSUE #228: Log more authentication events
We should modify the max integration to log when authorization fails and when it is successful.

cmc333333:
This builds on #262 to use the `ereqs_admin` app; the only new changeset is 76f8cac.

This adds (and tests) logging for authentication (successful and otherwise); nothing super special, but should be enough to resolve #228


cmc333333:
Thanks @tadhg-ohiggins !

PR #265, opened 2017-04-21
Request users not to cache content

### LINKED ISSUE #264: Set cache headers
When there's no cache headers present, Cloud Front caches content for 24 hours. Changes to code and content may not be visible during this time, and due to the relationship between the UI and API, the UI may not be accessible at all.

Resolve by adding Django caching middleware (perhaps copying 18F/calc#1260) and similar for Node.

cmc333333:
Without setting explicit headers, CloudFront has a cache timeout of 24 hours. This sets that timeout to zero (which is what we assumed) for the time being. It should resolve #264.

### DIFF
--- a/omb_eregs/settings.py
+++ b/omb_eregs/settings.py
@@ -57,6 +57,7 @@ if DEBUG:
     INSTALLED_APPS += ('debug_toolbar', )
 
 MIDDLEWARE = (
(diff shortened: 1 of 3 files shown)


cmc333333:
Thanks @tadhg-ohiggins

PR #266, opened 2017-04-25
Hotfix #265

cmc333333:
We need these cache headers in production before the next release.


cmc333333:
Thanks @theresaanna and indirectly @tadhg-ohiggins !

Issue #240, opened 2017-04-17
Implement new policy page style. 


cmc333333:
Why: To make the differences between the requirement view and the policy view more distinct, we are going to try a new version of the policy page that doesn’t show summaries of requirements within a policy, just the policies themselves and some important metadata, along with a count of how many requirements apply to the current set of filters.

[Please update with the latest mocks]

theresaanna:
Assigned to @micahtaylor for mocks. @micahtaylor, please assign to me when mocks are ready! Thanks.

micahtaylor:
@theresaanna, is this what we are calling the policy listing page?
That is found here - https://gsa.invisionapp.com/share/M6AU3YB49#/229078016_Policy_Listing

PR #268, opened 2017-04-26
Add requirement counts to policy API

cmc333333:
Part of #240, this adds counts to the policy API representing the number of requirements and the number of matching requirements.

This makes heavy use of Django's Subquery feature (added in 1.11), so it also upgrades that lib. Easiest to review step by step.

CC @theresaanna: this is the API we'll want to use on the policy page.


cmc333333:
Thanks @tadhg-ohiggins

PR #269, opened 2017-04-26
Log changes to console

### LINKED ISSUE #219: Add logging around permissions changes
Django logs lots of state changes but doesn't do so for user permissions. Use Django signals to create this record.

cmc333333:
As users make changes in the admin screen, Django generates minimal logs which are then stored in the database. This also logs those changes to the console. Additionally, Django's admin doesn't do a great job of tracking how many-to-many fields change (which is problematic as both permissions and groups are many-to-many), so this also specifically logs those.

Should resolve #219

### DIFF
--- a/ereqs_admin/apps.py
+++ b/ereqs_admin/apps.py
@@ -1,7 +1,7 @@
 import logging
 
 from django.apps import AppConfig
-from django.db.models.signals import post_migrate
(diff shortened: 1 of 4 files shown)


cmc333333:
Thanks @tadhg-ohiggins

PR #270, opened 2017-04-28
Remove password from MAX authenticated users

### LINKED ISSUE #252: Hide password field in admin view
MAX auth is based on the username (which has to be a .gov email address), but we don't use the password field, and hence should hide it in the admin interface.

cmc333333:
Adds new forms to account for an environment where passwords aren't used. These forms should only be active when there's a configured MAX_URL (so, to test locally, you'll need to uncomment that environment variable from the docker-compose file).

Should close #252


cmc333333:
Thanks for the review, @tadhg-ohiggins

PR #277, opened 2017-05-02
Add PDF upload/storage to admin interface (using S3)

tadhg-ohiggins:
There are PDFs /
That ain't gonna store themselves. /
Above, the fans whir.

Add PDF upload using S3 to the admin interface.

Uses minio to fake an S3 instance for local development, and real S3 on [cloud.gov](cloud.gov).

Thanks @cmc333333 for comments on earlier commits and help untangling the local Docker versus cloud.gov configurations.

Since there was no way to really test the cloud.gov side of this prior to deploying, I will probably hold off on merging to master until tomorrow morning.


docker-compose.yml
```diff
@@ -27,17 +34,35 @@ services:
...
+             "bucket": "pdfs",
+             "secret_access_key": "LOCAL_KEY",
+             "region": "irrelevant fake region",
+             "endpoint": "http://minio:9100"
+            },
+            "label": "s3",
+            "name": "storage-s3"
+            }
+           ]
+        }
```
cmc333333: Minor, but would you mind cleaning up this JSON so that it and the `config` key above are consistently formatted?

docker-compose.yml
```diff
@@ -67,8 +92,8 @@ services:
       INTERNAL_API: 'http://prod-api:9001/'
       PUBLIC_API: 'http://0.0.0.0:9001/'
       VCAP_SERVICES: >
-        {"config": [{"name": "config", "credentials": {"UI_BASIC_AUTH": {
-        }}}]}
+        {"config": [{"name":
+            "config", "credentials": {"UI_BASIC_AUTH": { }}}]}
```
cmc333333: Another small detail, but the previous formatting made it pretty easy to add a new set of credentials (as described in the README). Would you mind keeping it that way if it's not an issue?

manifest_dev.yml
```diff
@@ -9,12 +9,16 @@ applications:
...
+      - storage-s3  # S3 basic service
     command: ./devops/run_api.sh
     routes:
       - route: omb-eregs-api-demo.app.cloud.gov
     env:
       NEW_RELIC_APP_NAME: OMB Demo API/Admin
       MAX_URL: https://login.test.max.gov/cas/login
+      AWS_ACCESS_KEY_ID:
+      AWS_SECRET_ACCESS_KEY:
+      AWS_STORAGE_BUCKET_NAME:
```
cmc333333: I don't think these keys are needed anymore, right?

manifest_dev.yml
```diff
@@ -9,12 +9,16 @@ applications:
     services:
       - database  # aws-rds medium-psql
       - config    # user-provided service w/ DJANGO_SECRET_KEY and NEW_RELIC_LICENSE_KEY
+      - storage-s3  # S3 basic service
```
cmc333333: This needs to be added to `manifest_prod.yml` too, yes?

tadhg-ohiggins: Yup. I also have to actually create the S3 service for production, which I've so far held off on doing (until things work in dev, basically).

docker-compose.yml
tadhg-ohiggins: Sure, this was an unintended stray whitespace change.

docker-compose.yml
tadhg-ohiggins: Sure, although I'm not sure I've achieved real consistency here…

PR #284, opened 2017-05-03
Fix some performance issues in debug mode

cmc333333:
Root of the problem was django debug toolbar having issues with Django 1.11


cmc333333:
Thanks for testing, @tadhg-ohiggins !

PR #287, opened 2017-05-03
Fix issue around preloading topics.

### LINKED ISSUE #285: Problem with loading tags
Per https://github.com/18F/omb-eregs/issues/282#issuecomment-298986725 there's some issue loading topics on the edit form for requirements.

Investigate and fix. This is likely worth a hotfix.

cmc333333:
We updated to Django 1.11 this past sprint, which introduced an
incompatibility with django-autocomplete-light. Annoyingly, when going to the
edit screen, existing topics were not populated (though they were saved
correctly.)

This should resolve #285

### DIFF
--- a/reqs/tests/admin_tests.py
+++ b/reqs/tests/admin_tests.py
@@ -71,6 +71,19 @@ def test_reqs_in_policy(admin_client):
     assert req3.req_id not in result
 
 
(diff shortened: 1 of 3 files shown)


cmc333333:
Thanks for testing this, @tadhg-ohiggins !

PR #288, opened 2017-05-03
Hotfix version of #287

cmc333333:
I think the fix in #287 warrants a hotfix, so I've cherry-picked that commit.


cmc333333:

PR #290, opened 2017-05-04
Configure New Relic to ignore 401s.

### LINKED ISSUE #274: Modify new relic config to ignore 401s
Unauthorized requests aren't errors.

See https://docs.newrelic.com/docs/agents/nodejs-agent/installation-configuration/nodejs-agent-configuration#error_ignore

cmc333333:
By default, New Relic will treat a 401 as an error (hence affecting downtime
alerts, etc.). 401's do not indicate an error state for us, so we should
ignore them the same way 404s are.

Should resolve #274 

Tested on demo -- appears to continue to log hits while not registering errors on 401s


cmc333333:

PR #291, opened 2017-05-04
Add extension validator for PDF uploads in admin Policy form.

tadhg-ohiggins:
The name of a thing /
Should meet our expectations /
Sky should look like sky.

### DIFF
--- a/ereqs_admin/tests/admin_tests.py
+++ b/ereqs_admin/tests/admin_tests.py
@@ -92,6 +92,7 @@ def test_pdf_upload(admin_client):
     form_text = form.content.decode('utf-8')
     assert 'id_document_source' in form_text
     assert 'First Policy' in form_text
+    assert not policy.document_source
     data = {
         'policy_number': policy.policy_number,
         'title': 'First Policy Edited',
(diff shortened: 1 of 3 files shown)


ereqs_admin/tests/admin_tests.py
```diff
@@ -112,7 +113,37 @@ def test_pdf_upload(admin_client):
...
+    }
+    not_pdf_path = '{0}/not-a-pdf.txt'.format(pathlib.Path(__file__).parent)
+    with open(not_pdf_path, 'rb') as f:
+        data['document_source'] = f
+        resp = admin_client.post(policy_url, data)
+    resp_text = resp.content.decode('utf-8')
+    assert 'The file must be a PDF.' in resp_text
+    updated_policy = Policy.objects.get(id=policy.id)
+    assert updated_policy.title == 'First Policy'
+    assert not policy.document_source
```
cmc333333: Not a big deal, but this test has most of the same setup as the previous -- does it make sense to move that into a fixture?

ereqs_admin/tests/not-a-pdf.txt
```diff
@@ -0,0 +1 @@
+This file is not a PDF.
```
cmc333333: Ceci n'est pas une PDF.

ereqs_admin/tests/admin_tests.py
tadhg-ohiggins: My usual rule is to refactor to remove duplication after more than one repetition, so I'd do that if we had a third test of PDF-related things.

PR #293, opened 2017-05-06
Add a flag to mark policies as non-public.

tadhg-ohiggins:
Don't show hidden things. /
The butterflies will emerge /
When they are ready.

Add a `nonpublic` field to Policies, and stop the API from returning both policies marked as nonpublic and requirements whose policies are marked as nonpublic.


reqs/admin.py
```diff
@@ -15,6 +15,10 @@ def is_extension_pdf(uploaded_file):
 class PolicyForm(forms.ModelForm):
     document_source = forms.FileField(validators=[is_extension_pdf])
 
+    def __init__(self, *args, **kwargs):
+        super(PolicyForm, self).__init__(*args, **kwargs)
+        self.fields['document_source'].required = False
```
cmc333333: Would it be better to set `required=False` in line 16 instead?

reqs/models.py
```diff
@@ -69,6 +69,7 @@ class Meta:
     sunset = models.DateField(blank=True, null=True)
     policy_status = models.CharField(max_length=256, blank=True)
     document_source = models.FileField(blank=True)
+    nonpublic = models.NullBooleanField()
```
cmc333333: Do we really want to allow this boolean to be null (i.e. three-value logic)? I'd prefer if we picked a default instead:
* `False` is pseudo implicit given that we only ever use this logic to `exclude`
* `True` might make sense as a safety for OMB (forcing them to actively "publish")

Either way, we'd want a migration script which sets all existing policies to `False`

reqs/tests/views_tests.py
```diff
@@ -210,3 +210,37 @@ def test_policies_counts_filter_by_multiple_topics(policy_setup):
...
     assert response['results'][1]['relevant_reqs'] == 1
+
+
+@pytest.mark.django_db
+def test_excludes_nonpublic():
+    """The API endpoint should not include policies flagged as nonpublic"""
+    policies = [mommy.make(Policy, policy_number='0'),
+                mommy.make(Policy, policy_number='1', nonpublic=True)]
+    reqs = [mommy.make(Requirement, policy=policies[0], _quantity=3),
+            mommy.make(Requirement, policy=policies[1], _quantity=4)]
```
cmc333333: Minor: naming these policies/requirements `public_policy` `nonpublic_policy`, etc. might make reading the logic a tad easier.

reqs/views.py
```diff
@@ -141,6 +143,12 @@ def filter_queryset(self, request, queryset, view):
             return self.priority_ordering(request, queryset)
 
 
+class HiddenPolicyFilter(BaseFilterBackend):
+
+    def filter_queryset(self, request, queryset, view):
+        return queryset.exclude(policy__nonpublic=True)
```
cmc333333: As I understand it, these `Filter` classes are for dynamic (per request) logic. Should we just override `get_queryset` instead?

reqs/admin.py
tadhg-ohiggins: Yes, yes it would.

reqs/views.py
tadhg-ohiggins: Good point, I'll make that change.

reqs/models.py
tadhg-ohiggins: I was thinking of the (future, maybe) Requirement case (where null would effectively mean “inherit”) instead of the Policy case when I made it three-value logic; I've made it default to `False`.

Issue #251, opened 2017-04-17
Clean up feedback from OMB


yowill:
- ~~Hide password in admin user view.~~ #252
- ~~Remove password field when MAX is in use.~~ #252
- Figure out a way to flag official guidance vs policy.
- ~~Policy podcasts.~~
- ~~Add memo ID near policy title — that is, near the bold text here, add the memo ID~~: 
![image](https://cloud.githubusercontent.com/assets/2626258/25106523/cf10e064-237e-11e7-832a-19e859bb07ef.png)
- ~~Potentially add memo ID to the list of properties on requirements, just after policy title.~~
- Add slug field to requirements and the ability to search/link by it. Potentially only show slug if present, requirement ID otherwise. See also #241.

tadhg-ohiggins:
#252

theresaanna:
This issue needs splitting out and assigning.

cmc333333:
> Figure out a way to flag official guidance vs policy.

Is this different from the existing policy "type" field?

> Add slug field to requirements and the ability to search/link by it. Potentially only show slug if present, requirement ID otherwise.

This seems super sketchy still. I'd vote for removing it and see if it comes up again.

cmc333333:
Per brief convo with @tadhg-ohiggins, I'm going to close this; we'll see if the remaining requests come up again.

PR #295, opened 2017-05-08
Add a "title_with_number" policy attribute.

cmc333333:
Per #251, we'd like to include a policy number (where available) in the UI.
Rather than calculate it there, this just adds a computed field to the API
which includes the information.

It then uses this field when displaying the policy titles in the policies tab.


cmc333333:

PR #303, opened 2017-05-09
Up the number of UI services we're running.

cmc333333:
The UI instances are rather minimal -- not only does Node have a light
footprint (hovering around 50M usage), most users will only hit the service
once on initial page load. As a result, it looks like we can spin up several
more UI servers to handle request volume by dropping their memory consumption.


cmc333333:
Thanks @theresaanna

Issue #296, opened 2017-05-08
Activate and tweak pdf link


cmc333333:
Text should say "View original" with an external link icon. It should point to the PDF, if available, or the original url if not.

![policy listing](https://cloud.githubusercontent.com/assets/326918/25818320/3f8f20d8-33f8-11e7-84ae-02898cdf9d38.png)

cmc333333:
Merged!

PR #305, opened 2017-05-10
Activate download link

cmc333333:
Per #296, swap out the text/icon and point the download link to either the pdf (where available) or the original url.

Looks like:
screen shot 2017-05-10 at 10 36 25 am


cmc333333:
Thanks @theresaanna

PR #307, opened 2017-05-10
Add a tabIndex to the react-select autocompleter.

cmc333333:
This makes the autocompleter tab-accessible.

This is part (perhaps all?) of #292


cmc333333:
Hah, glad to be of helpful! Thanks @theresaanna!

PR #311, opened 2017-05-10
Reduce number of queries in the policies edit page.

cmc333333:
We had been making one query per requirement on the policies edit page (to
gather the topics names). We can prefetch all of those topics in a single
query instead, though it requires a few hoops.

This cuts load time of one of the pages from about 30 seconds to 6 for me (in
dev mode), so, still not fast, but much better.


cmc333333:
Thanks @tadhg-ohiggins !

PR #318, opened 2017-05-11
README typo.

cmc333333:
Nothing to see here...


cmc333333:
Thanks for not seeing with me, @tadhg-ohiggins

PR #326, opened 2017-05-12
Django settings tweaks

cmc333333:
The CORS headers one is a little hard to test (browsers generally ignore CORS stuff for localhost). What I'd recommend is 
* [verifying](http://pythex.org/) the regex works to allow CORS on all of the api endpoints, but not /static/... or /admin/....
* Check the headers (look for "Access-Control-Allow-Origin: *") using curl:
```
curl -I http://localhost:9001/static/admin/img/icon-addlink.svg -H "Origin: http://example.com/"
```
Unfortunately, that won't work on the admin site (due to the authentication redirect), unless you want to pass along a cookie.

Re session expiry, I'm just going off Django's [docs](https://docs.djangoproject.com/en/dev/ref/settings)

### DIFF
--- a/omb_eregs/settings.py
+++ b/omb_eregs/settings.py
@@ -94,7 +94,7 @@ if DEBUG:
 # Allow most URLs to be used by any service; do not allow the admin to be
 # accessed this way
 CORS_ORIGIN_ALLOW_ALL = True
-CORS_URLS_REGEX = r'^/(?!admin).*$'
+CORS_URLS_REGEX = r'^/(?!admin)(?!static).*$'
 
 # Request the browser not allow the CSRF cookie to be used in JS (not: this
 # means we can't have AJAX forms)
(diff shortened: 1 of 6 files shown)


omb_eregs/settings.py
```diff
@@ -94,7 +94,7 @@
 # Allow most URLs to be used by any service; do not allow the admin to be
 # accessed this way
 CORS_ORIGIN_ALLOW_ALL = True
-CORS_URLS_REGEX = r'^/(?!admin).*$'
+CORS_URLS_REGEX = r'^/(?!admin)(?!static).*$'
 
 # Request the browser not allow the CSRF cookie to be used in JS (not: this
```
tadhg-ohiggins: Unrelated to this commit, but should “not:” be “note:” here?

cmc333333:
When reviewing the docs around [`whitenoise` and CORS](http://whitenoise.evans.io/en/stable/django.html#WHITENOISE_ALLOW_ALL_ORIGINS), I thought through again that static files _should_ be accessible, regardless of origin. So I'm going to close this PR and open a separate one with just the other changesets.

PR #327, opened 2017-05-12
Sprint Integrale Release

cmc333333:
All the sprint's work. Nice job all!


cmc333333:
Note the `push` CI is failing due to deployment step (we've been using all the resources in the demo environment to ensure testing is unperturbed). We should be able to ignore.

PR #328, opened 2017-05-12
Insert newrelic browser logging.

cmc333333:
The Python client does this automatically, but the Node client does not.
Ironically, we probably want statistics around the Node app's browser more
than anything. Activate it here (it degrades to an empty string when newrelic
is not enabled).

### DIFF
--- a/ui/components/html.jsx
+++ b/ui/components/html.jsx
@@ -1,22 +1,44 @@
+import newrelic from 'newrelic';
 import React from 'react';
 import serialize from 'serialize-javascript';
 
 import config from '../config';
 import DAP from './dap';
 
-
-export default function Html({ contents, data }) {
+function scriptTag(data) {
   const jsStr = `
     window.__REACT_RESOLVER_PAYLOAD__ = ${serialize(data)};
     window.APP_CONFIG = ${serialize(config)};
   `;
   // Avoid escaping the JS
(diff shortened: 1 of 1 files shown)


cmc333333:
This is broken, please don't review yet

cmc333333:
Apologies for the false start, this should be ready to review now. To test locally, modify your docker-compose.yml so that under the `dev` service, you add a new env var for the app name and under the `VCAP_SERVICES`, you add the new relic license key (which can be found via `cf env ui` in the demo or prod environments).
The diff looks something like
```diff
diff --git a/docker-compose.yml b/docker-compose.yml
index 44675fd..c8bbcbc 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -100,9 +100,10 @@ services:
       NODE_ENV: ''
       INTERNAL_API: 'http://dev-api:8001/'
       PUBLIC_API: 'http://0.0.0.0:8001/'
+      NEW_RELIC_APP_NAME: Test
       VCAP_SERVICES: >
         {"config": [{"name": "config", "credentials": {"UI_BASIC_AUTH": {
-        }}}]}
+        }, "NEW_RELIC_LICENSE_KEY": "key-goes-here"}}]}
     depends_on:
       - dev-api
```

cmc333333:
Thanks for testing, @theresaanna

PR #336, opened 2017-05-15
More security things

cmc333333:
Changesets should be self-explanatory

### DIFF
--- a/omb_eregs/settings.py
+++ b/omb_eregs/settings.py
@@ -107,6 +107,8 @@ SECURE_CONTENT_TYPE_NOSNIFF = True
 USING_SSL = env.get_credential('USING_SSL', 'TRUE').upper() == 'TRUE'
 SESSION_COOKIE_SECURE = USING_SSL
 CSRF_COOKIE_SECURE = USING_SSL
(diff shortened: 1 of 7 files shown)


omb_eregs/settings.py
```diff
@@ -107,6 +107,7 @@
 USING_SSL = env.get_credential('USING_SSL', 'TRUE').upper() == 'TRUE'
 SESSION_COOKIE_SECURE = USING_SSL
 CSRF_COOKIE_SECURE = USING_SSL
+CAS_RENEW = True
```
tadhg-ohiggins: Is this what makes MAX force users to log in again if they've logged out from the admin UI? If so, we might want a comment to that effect.

reqs/admin.py
```diff
@@ -8,7 +8,8 @@
 
 
 def is_extension_pdf(uploaded_file):
-    if not uploaded_file.name.endswith('pdf'):
+    if (not uploaded_file.name.endswith('pdf')
+            or uploaded_file.content_type != 'application/pdf'):
```
tadhg-ohiggins: :+1:

PR #345, opened 2017-05-17
Create shared, "proxy" network in docker

cmc333333:
We'd like to use a `localhost` URL when presenting uploaded files to end users
(so they can access those files outside of Docker). ~~To do this, we'll attach
the APIs to the minio network stack. This conflicts with exposing ports
directly, however, so we need *minio* to expose those ports.~~

To do this, we'll create a "proxy" container which exists solely for other containers to join its network stack. This will allow apps to talk to each other via `localhost` just as the end user will.

@tadhg-ohiggins what do you think of this approach?

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,19 +146,6 @@ services:
       MAX_URL: https://example.com/etc
 ```
 
-### Running w/ Local PDF Upload
-
-The default configuration includes a [Minio](https://www.minio.io/) server
-that we use to mock S3 for local development.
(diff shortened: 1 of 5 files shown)


cmc333333:
Quick note that this doesn't quite work yet -- we can't see the UI.

I'm trying a related approach (having a single container act as the "proxy" for several apps, talking to each other through localhost) now.

cmc333333:
Thanks for reviewing and talking through this, @tadhg-ohiggins!

PR #346, opened 2017-05-17
Remove separate private/public API urls.

cmc333333:
We had two different URLs to account for communications within the docker
network and between the browser and API. Since #345 puts all of that network
into one proxy component, we can collapse into a single configuration option.


cmc333333:
Thanks @tadhg-ohiggins

PR #349, opened 2017-05-25
Beginnings of a homepage

theresaanna:
As I said in the channel, I don't know that my solution for the form is a _good_ one, but it is one. Looking for feedback.

Scope it:
The "about" text is weird. Merriweather renders so much differently in Chrome. Should I do this in a light variant instead?
screen shot 2017-05-25 at 3 16 25 pm

### DIFF
--- a/ui/components/filters/autocompleter.js
+++ b/ui/components/filters/autocompleter.js
@@ -8,7 +8,6 @@ import { Async } from 'react-select';
 
 import SearchView from './search-view';
 import { apiParam, redirectQuery, search } from '../lookup-search';
-import redirectWhiteList from '../redirectWhiteList';
 
 export default class Autocompleter extends React.Component {
   constructor(props, context) {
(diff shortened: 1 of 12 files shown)


ui/components/filters/autocompleter.js
```diff
@@ -45,7 +47,13 @@ export default class Autocompleter extends React.Component {
...
   render() {
     const { insertParam, lookup } = this.props;
-    const { pathname, query } = this.context.router.location;
+    const { query } = this.context.router.location;
+    let { pathname } = this.context.router.location;
+    // if we are coming from a page other than /requirements or
+    // /policies (ex. the homepage), default to going to /requirements
+    if (redirectWhiteList.includes(pathname) === false) {
+      pathname = '/requirements';
+    }
```
cmc333333: I think there's confusion here over `redirectWhiteList`. When a user doesn't have JS, they're sent to a search results page whose links are populated by a set of `redirect` fields. As a security measure, we prevent search results from redirecting to _any_ page by using a whitelist.

So, the `pathname` variable should contain the _destination_ of the redirect. Thus far, the destination has been the same page, but that's not the case anymore. We probably want to check `this.props.pathname` instead of the current page (`this.context.router.location`).

ui/components/homepage.jsx
```diff
@@ -0,0 +1,49 @@
...
+export default function Homepage() {
+  const filterControls = [
+    React.createElement(FilterListView, {
+      autocompleter: React.createElement(Autocompleter, {
+        insertParam: 'topics__id__in',
+        lookup: 'topics',
+        pathname: '/requirements',
+      }),
+      key: 'topic',
+    }),
```
cmc333333: We'll probably need to tweak this a bit in the next iteration (with agencies) as the Autocompleter currently submits/redirects on the first entry selection. There's a multi-select option, too, which seems appropriate.

PR #347, opened 2017-05-17
Agencies API

cmc333333:
This
* Adds fields in requirements for agencies and agency groups
* Adds agency and agency group endpoints
* Allows requirements to be filtered by agency and/or group

Easiest to review changeset by changeset. Should resolve #338 and parts of #340.

PR #351, opened 2017-05-30
Add view + endpoints for filtering reqs by agency

### LINKED ISSUE #340: Agencies: API queries
* We should include agency + group information in the requirements endpoint
* We need to be able to query agencies and groups (e.g. to get their name)
* We need to allow users to transitively search (so that if I look for GSA, I get requirements tagged with GSA + requirements tagged with any agency group GSA is a part of)

cmc333333:
In #347, we added a mechanism for requirements to be filtered _directly_ by agency or agency group. This PR modifies that logic to allow us to filter by agencies indirectly (i.e. filter by agencies which are part of a _group_ which is associated with a requirement).

To do that, it creates a database view which includes all agencies associated with a requirement (directly or indirectly).

Should resolve #340


cmc333333:
[rebased]

cmc333333:
Thanks for the review @tadhg-ohiggins !

PR #355, opened 2017-05-31
Remove add permission for agencies.

cmc333333:
As agencies are populated via an import script, we should remove the ability
for admins to create new instances.


cmc333333:
Thanks @tadhg-ohiggins

PR #356, opened 2017-05-31
First pass at full-text search

cmc333333:
This isn't a complete solution, but should allow us to build out a UI and tweak functionality independently.

* Add a search index on requirement text
* Surface full-text search filtering through the API
* Display searched terms in the agency-facing UI


cmc333333:
Thanks @tadhg-ohiggins and @theresaanna !

PR #368, opened 2017-06-01
Move issuing body from requirement to policy.

### LINKED ISSUE #278: Move "Issuing Body" from Requirement to Policy
Based on feedback from @Thomas-Romanoff , the "Issuing Body" is shared between all requirements in a policy.

cmc333333:
The issuing body field should be consistent across all requirements within a
single policy. This moves that meta-data to the Policy model to match that
expectation.

Should resolve #278


cmc333333:
[rebased and added a merge migration]

cmc333333:
Thanks for reviewing @tadhg-ohiggins !

PR #370, opened 2017-06-02
Print layout for requirements and policies lists

theresaanna:
Simplifies page layout and formats requirements and policies lists appropriately for printing.

Do we want to leave in the content rule path information on the topic links?


ui/assets/css/_print.css
```diff
@@ -0,0 +1,38 @@
...
+    border-radius: 0;
+  }
+
+  .topics-list a, a:link {
+   text-decoration: none;
+   color: #111;
+  }
+
+  .topics-list a:link:after {
+    content: " (" attr(href) ") ";
```
micahtaylor: Let's remove this for now. If we have more links in the future we'll consider it again.

PR #372, opened 2017-06-05
UI for site search

theresaanna:
Adds a form to the header for site search. 
screen shot 2017-06-07 at 9 14 03 am


ui/components/app.jsx
```diff
@@ -7,7 +7,7 @@ function App(props) {
   return (
     <div>
       <Disclaimer />
-      <Navbar />
+      <Navbar router={props.router} />
```
cmc333333: We might be able to add the router as a `contextType` of the `Search` component. Then we wouldn't need to thread it through each of the intervening components. Looks like we have an example in the `Autocompleter`

ui/components/search/search.jsx
```diff
@@ -0,0 +1,48 @@
...
+    super(props);
+
+    this.state = { searchTerm: '' };
+    this.handleSubmit = this.handleSubmit.bind(this);
+    this.handleChange = this.handleChange.bind(this);
+  }
+
+  handleChange(e) {
+    this.setState({ searchTerm: e.target.value });
+  }
```
cmc333333: Adding state here is a bit of a mixed bag. On one hand, it means that changes to the rest of the page won't affect the state of this input (so the text the user wrote will remain, even if they change the URL by adding/removing a filter), if I understand correctly. On the other hand, it makes the logic harder to follow since we don't have a simple data-in-xml-out situation; for example, do we ever need to clear the state? What do you think we should do?

If we're going to keep state, I'd like to add a test to validate that state gets set onChange.

ui/components/search/search.jsx
```diff
@@ -0,0 +1,48 @@
...
+
+  handleSubmit(e) {
+    e.preventDefault();
+    const { location } = this.props.router;
+    const query = redirectQuery(location.query, 'req_text__search', this.state.searchTerm);
+    const paramStr = querystring.stringify(query);
+    this.props.router.push(`/requirements/?${paramStr}`);
+  }
```
cmc333333: I think we can avoid capturing this submission event in Javascript. Can we use an HTML form instead?
```
function hiddenFields(query) {
    const modifiedQuery = Object.assign({}, query);
    del modifiedQuery.page;
    del modifiedQuery.req_text__search;
    return Object.keys(modifiedQuery).map(k =>
        <input type="hidden" key={k} name={k} value={modifiedQuery[k]} />);
}

// will need to select actionPath from the current location
<form method="GET" action={ actionPath }>
    <input name="req_text__search" type="text" placeholder="Search..." />
    { hiddenFields(location.query) }
    <input type="submit" value="Submit" />
</form>
```

The benefit is that this works without JS.

ui/components/search/search.jsx
```diff
@@ -39,10 +39,12 @@ export default class Search extends React.Component {
...
   router: React.PropTypes.shape({
-    location: React.PropTypes.shape,
+    location: React.PropTypes.shape({
+      query: React.PropTypes.shape({}),
+      pathname: React.PropTypes.string,
+    }),
     push: React.PropTypes.func,
-  }).isRequired,
+  }),
 };
```
cmc333333: Nice! Glad this worked!

ui/components/search/search.jsx
```diff
@@ -1,38 +1,24 @@
...
   }
 
-  handleSubmit(e) {
-    e.preventDefault();
-    const { location } = this.context.router;
-    const query = redirectQuery(location.query, 'req_text__search', this.state.searchTerm);
-    const paramStr = querystring.stringify(query);
-    this.context.router.push(`/requirements/?${paramStr}`);
+  inputName() {
+    return this.context.router.location.pathname.includes('requirements') ? 'req_text__search' : 'requirements__req_text__search';
```
cmc333333: Small detail, but it might make sense to check for `policies` instead (i.e. reverse the logic). The `else` case here will be triggered by the home page (and any future pages); I'd think we'd want to default that to searching on `/requirements` rather than `/policies`.

ui/components/search/search.jsx
```diff
@@ -1,38 +1,24 @@
...
   render() {
     return (
-      <form onSubmit={this.handleSubmit}>
-        <input
-          value={this.state.searchTerm}
-          onChange={this.handleChange}
-          type="text"
-          placeholder="Search..."
-        />
+      <form method="GET" action={this.context.router.location.pathname}>
```
cmc333333: I think we need to do something similar to `inputName()` here so that we submit the form to `/requirements` on the home page.

PR #375, opened 2017-06-06
Add a spinner during AJAX load

### LINKED ISSUE #373: AJAX spinner
When data is being loaded asynchronously, we should include some sort of loading indicator. This should *not* be present when rendering the HTML on the server.

cmc333333:
React-resolver doesn't quite give us the tools needed to implement a loading-spinner during AJAX requests, but we can extend their [Resolver](https://github.com/ericclemmons/react-resolver/blob/master/src/Resolver.js) to get there.

Looks like
spinner

Should resolve #373

### DIFF
--- a/ui/components/spinner.jsx
+++ b/ui/components/spinner.jsx
@@ -1,0 +1,22 @@
+import React from 'react';
+
+export default function Spinner({ height }) {
+  const borderWidth = Math.ceil(height / 10);
+  const indicatorStyle = { height: `${height}px` };
+  const spinnerStyle = {
+    borderWidth: `${borderWidth}px`,
+    height: `${height}px`,
+    width: `${height}px`,
+  };
+  return (
+    <div className="loading-indicator p2" style={indicatorStyle}>
+      <span className="loading-spinner" style={spinnerStyle} />
+    </div>
+  );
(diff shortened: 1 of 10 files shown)


cmc333333:
[Delayed till post-sprint]

ui/components/spinner.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+
+export default function Spinner({ height }) {
+  const borderWidth = Math.ceil(height / 10);
+  const indicatorStyle = { height: `${height}px` };
```
theresaanna: It'd be really nice if we could have the dimensions of the spinner flex automatically with its container. Any thoughts about why that might be/not be possible? Definitely a future improvement, though.

cmc333333: Agreed, that would be wonderful. I hacked around a little with 100% width/height and setting max values on containers, but had two issues that I didn't know how to get around:

1. The indicator _must_ be a square, otherwise it looks like a wobbly oval
2. The divs we're loading (selected filters, main page content) are arbitrarily tall

Hopefully this is easy enough to play with now; maybe using a gif would solve those issues?

Issue #280, opened 2017-05-03
Add a new "office" field to policies


cmc333333:
This should represent the office that is responsible for managing the policy.

It's unclear whether it makes sense to hard code these values or allow them to be modified in the Admin (I'd lean towards the latter).

cmc333333:
#376 resolved

PR #376, opened 2017-06-06
Add an Office model + field.

cmc333333:
OMB's asked that policies be divided by managing office. This creates a model
to represent each office and a nullable connection on Policies. We may
mass-populate this data in the future, but for now, this should be enough to
satisfy #280.


cmc333333:
Thanks for the quick review, @tadhg-ohiggins !

PR #380, opened 2017-06-08
Add agency data step to README.

cmc333333:
This also adds a `createinitialrevisions` step to give each of these new
objects an initial version in history.


cmc333333:
Thanks @tadhg-ohiggins

PR #382, opened 2017-06-09
Comment out Agency UI.

cmc333333:
As we don't have data yet, temporarily hide these inputs from the user-facing
UI.


cmc333333:
We'll get this next time, thanks @tadhg-ohiggins

PR #387, opened 2017-06-09
Hiding search box on print.

micahtaylor:


micahtaylor:
Yeah, I was surprised basecss did not have that already. We should as it is ideal and inline with the front end ideal.

PR #388, opened 2017-06-09
Add the office name to the Django admin.

cmc333333:
Django models should always have a __str__; otherwise, when the model is in a
select field, it'll just say, e.g. "Agency Object".

Compare:

## Old
screen shot 2017-06-09 at 4 13 13 pm

## New
screen shot 2017-06-09 at 4 14 04 pm


cmc333333:
Thanks @tadhg-ohiggins

PR #395, opened 2017-06-12
Theme Django Admin

cmc333333:
We'd like to change the colors of the Django admin. While we could do this with a custom stylesheet, [django-admin-interface](https://github.com/fabiocaccamo/django-admin-interface) has already pulled out a bunch of the colors we need to override. We don't need its editable color interface, but we can play its game by pre-populating a theme model with appropriate colors (and never activating the admin interface to edit it).

This resolves over a dozen color contrast issues.

Looks like
uswds-django


cmc333333:
Thanks @tadhg-ohiggins!

Issue #393, opened 2017-06-12
Allow individual requirements to be hidden


cmc333333:
We have a `nonpublic` flag on policies; we need to do the same for requirements.

Be sure to adjust the requirements number on the policy page.

yowill:
CM to follow up with Matt & Tom

cmc333333:
Matt has confirmed this is a priority.

I'm going to flip the logic to use a `public` flag everywhere, then use this logic table to determine visibility:

| policy | requirement | req visible |
|---|---|---|
| public    | public      | yes     |
| public    | nonpublic   | no      |
| nonpublic | public      | no      | 
| nonpublic | nonpublic   | no      |

PR #396, opened 2017-06-13
s/nonpublic/!public

cmc333333:
As a first step towards #393, invert the logic and fieldname from `nonpublic` to `public`.


cmc333333:
Closing in favor of #397, which includes this logic.

PR #397, opened 2017-06-13
Allow individual requirements to be hidden

### LINKED ISSUE #393: Allow individual requirements to be hidden

cmc333333:
This PR

* renames the `nonpublic` field to `public` everywhere to avoid the double negation
* adds a new `public` field to requirements
* inspects that field when displaying requirements (at /requirements) or their count (at /policies)

This should close #393 unless we get a correction around logic from Matt.


cmc333333:
[re-un-re-un-re-based] Thanks @tadhg-ohiggins

Issue #374, opened 2017-06-05
Unable to restore deleted policy


carodew:
On my local build, I was unable to restore any deleted policies even though I deleted a few. 

## Steps to recreate 
1. Through the admin interface, I deleted a few policies, confirmed in the recent actions log: 
![image](https://cloud.githubusercontent.com/assets/10144074/26802594/126af906-4a0f-11e7-93e2-a4692bdc310b.png)

2. from the Policies list page, I clicked on "Recover deleted policies:"  
![image](https://cloud.githubusercontent.com/assets/10144074/26802868/1df9ebd2-4a10-11e7-992b-a146148fc324.png)

3. However,  the next page doesn't show any deleted policies to recover: 
![image](https://cloud.githubusercontent.com/assets/10144074/26802916/50dcb4bc-4a10-11e7-91d5-b378c8ae0784.png)

carodew:
Please let me know if I can provide more info.

tadhg-ohiggins:
While I haven't been able to reproduce this problem locally, I have found some other history-related issues that I'm going to keep here until we have somewhere else to put them:

- [ ] The contents of `issuing_body` aren't stored in history, presumably because the script that added it didn't `save()` inside of a `reversion.create_revision()` decorator/context manager—something we should keep in mind for future changes.
- [ ] Deleting a policy object deletes (after warning you) all of its related requirement objects, which is what we want. However, restoring the policy object via history doesn't restore them—you have to go to the **Recover Deleted Requirements** section and restore them one by one. We may have to live with this, but it's problematic for policies with many requirements.
- [ ] Requirements don't save their topics, so those relationships cannot be restored after deletion (or other changes). This is likely due to my not setting up the relationships correctly as described here: <https://django-reversion.readthedocs.io/en/release-1.8.1/api.html#advanced-model-registration>. I think this is important to fix.

tadhg-ohiggins:
I'll try to reproduce the original bug on a different machine, too.

cmc333333:
I wonder if the original bug came from not having the initial set of revisions? Restarting the server creates those revisions, but maybe we had an order-of-operations issue?

Agreed that we should save imported reqs/policies/topics w/ the revision library.

carodew:
I'm happy to Do Things and retest if that's helpful at all.

cmc333333:
Re: "The contents of issuing_body aren't stored in history":

[...]

Issue #371, opened 2017-06-05
Heuristic evaluation of admin interface, focusing on quick wins


carodew:
Review pages in the admin interface and look for quick things we can change before release. 

Bonus – also come up with research questions.

cmc333333:
Results are being put in https://docs.google.com/document/d/1jpLJDbhgn8PyzEH5TN6ShmiOIXv9ZvbVEyzVpGzCq88/edit

cmc333333:
@carodew did the initial review and we implemented several updates in #381 and #395, so I'm going to close this.

Issue #282, opened 2017-05-03
Remove ability to add new tags within a requirement


cmc333333:
To add a barrier, force the user to add the topic via a separate Admin screen before they can use it.

Thomas-Romanoff:
@cmc333333 Just found that whenever I edit a requirement- the system de-tags the requirement, meaning i have to re-tag each requirement that I edit.

cmc333333:
Pulling this into the sprint as it's required by #374

cmc333333:
Resolved by #398

PR #398, opened 2017-06-14
Remove taggit

cmc333333:
We want to stop using taggit for several reasons:
* Per #282, we don't want to allow users to add new topics while editing requirements (this removes the key value of taggit)
* Per #374, taggit doesn't play well with Django-reversion
* Per #371, our use of taggit resulted in join-table-models which confuse users, particularly when deleting

This PR does that in several steps and may be easiest to review changeset by changeset.


cmc333333:
Thanks for the review @tadhg-ohiggins !

PR #400, opened 2017-06-15
Create versions in import/sync scripts

cmc333333:
This is step 2 of fixes around #374; it creates versions when importing/synchronizing Policies, Requirements, Topics, Agencies, and AgencyGroups.

This PR also includes a somewhat unrelated test refactor which converts some test data into a factory method -- I ended up not needing this, but figured it wouldn't hurt to keep.


cmc333333:
Thanks @tadhg-ohiggins

PR #403, opened 2017-06-16
342 add agency relationships

tadhg-ohiggins:
Yet more TLAs /
That requirements /
Often refer to.

Adds a Django management command that extracts info from the requirement `impacted_entity` field about which agencies and agency groups that requirement refers to.


reqs/tests/link_agencies_tests.py
```diff
@@ -0,0 +1,123 @@
...
+        req_agencies = {_.abbr for _ in req.agencies.all()}
+        assert req_agencies == agencies[i]
+        req_groups = {_.slug for _ in req.agency_groups.all()}
+        assert req_groups == agency_groups[i]
+
+
+@pytest.mark.django_db
+def test_link_requirements(populate_agencies):
+    dod_req = mommy.make(Requirement, impacted_entity="DOD")
+    dod_agency = Agency.objects.get(abbr="DOD")
```
cmc333333: Here's a use case where you may want to yield something in the `populate_agencies` fixture. Instead of re-querying, you may provide an interface like `populate_agencies.dod` and `populate_agencies.all_agencies`

reqs/management/commands/link_agencies.py
```diff
@@ -0,0 +1,90 @@
...
+from django.core.management.base import BaseCommand
+
+from reqs.models import Agency, AgencyGroup, Requirement
+
+
+class Command(BaseCommand):
+    help = ("Associate agencies and agency groups with requirements "  # noqa
+            "based on impacted entity field.")
+
+    def handle(self, *args, **options) -> None:
```
tadhg-ohiggins: I was using them as comments (although also running them through a rogue mypy locally) as an alternative to docstring comments about what types a function expects.

I didn't want to lay out a policy for the project, but I do think that regardless of whether or not we expect total mypy coverage, it's better to have type annotations on Python 3 projects than not to have them.

reqs/management/commands/link_agencies.py
```diff
@@ -0,0 +1,90 @@
...
+        self.process_requirements()
+
+    def process_requirements(self) -> None:
+        """
+        Go through each requirement.
+        If its impacted_entity field is blank, return the requirement.
+        Otherwise, pass the value of impacted_entity to extract_entities,
+        then pass the result to create_relationship.
+        """
+        for requirement in Requirement.objects.all():
```
tadhg-ohiggins: Whoops, good catch!

reqs/management/commands/link_agencies.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+    matched_entities = []
+    cond_lookup_groups = (
+        ("cfo" in entities and "federal cfo council" not in entities,
+         "cfo-act"),
+        (all([
+            "all agencies" in entities,
+            "micro" not in entities,
+            "paperwork" not in entities,
+        ]), "all-agencies"),
```
tadhg-ohiggins: Seems reasonable, although (see next comment) I think a lot of this will end up being rewritten in the near future due to more data coming from OMB.

reqs/management/commands/link_agencies.py
```diff
@@ -0,0 +1,90 @@
...
+    )
+    cond_lookup_agencies = (
+        ("dhs" in entities or "department of homeland" in entities, "DHS"),
+        ("gsa" in entities or "general services" in entities, "GSA"),
+        ("dod" in entities or "defense" in entities, "DOD"),
+        ("doj" in entities or "justice" in entities, "Justice"),
+    )
```
tadhg-ohiggins: It looked to me like that would create too many false positives. I focused on only the agencies that seemed to have a reasonable number of results in my ad-hoc testing.

Right now it looks like we might get better data about how to match strings to agencies from OMB, in which case the above code will need to be entirely rewritten anyway (which was kind of my expectation in the first place).

reqs/management/commands/link_agencies.py
```diff
@@ -0,0 +1,90 @@
...
+
+    for cond, lookup in cond_lookup_groups:
+        if cond:
+            matched_entities.append(AgencyGroup.objects.get(slug=lookup))
```
tadhg-ohiggins: Good point; shouldn't be hard to do.

reqs/management/commands/link_agencies.py
```diff
@@ -0,0 +1,90 @@
...
+
+    for cond, lookup in cond_lookup_agencies:
+        if cond:
+            matched_entities.append(Agency.objects.get(abbr=lookup))
```
tadhg-ohiggins: I can go either way on this one, but generally, if in doubt, I split it into more functions.

reqs/management/commands/link_agencies.py
```diff
@@ -0,0 +1,90 @@
...
+
+    return matched_entities
+
+
+@singledispatch
```
tadhg-ohiggins: `singledispatch` was almost definitely overkill here; I was experimenting with whether or not keeping the types separate made sense. I'm still not sure, but I think your outline above makes more sense.

reqs/management/commands/link_agencies.py
```diff
@@ -0,0 +1,90 @@
...
+@reversion.create_revision()
+def create_relationship(entity: Agency, requirement: Requirement) -> None:
+    requirement.agencies.add(entity)
+    reversion.set_comment("Add agency.")
```
tadhg-ohiggins: Heh, true, but it was better than “-”, which is what it was before… I'll make it more informative.

reqs/management/commands/link_agencies.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+@create_relationship.register(AgencyGroup)
+@reversion.create_revision()
+def crg(entity: AgencyGroup, requirement: Requirement) -> None:
+    requirement.agency_groups.add(entity)
+    reversion.set_comment("Add agency group.")
+
+
+def reset_agency_relationships() -> None:
```
tadhg-ohiggins: I used it a lot in local development, and am not sure whether or not to get rid of it.

reqs/tests/link_agencies_tests.py
```diff
@@ -0,0 +1,123 @@
+import pytest
+from model_mommy import mommy
+
+from reqs.management.commands import link_agencies
+from reqs.management.commands.sync_agencies import Command as SynchCommand
```
tadhg-ohiggins: I blame you for this, since you named it `sync_agencies`. (I'll probably get around this by not providing an alias, since refactoring I did after initially importing it like that means it's not really necessary.)

reqs/tests/link_agencies_tests.py
```diff
@@ -0,0 +1,123 @@
...
+        'agencyName': 'Department of Foo',
+        'agencyType': '',
+        'CFO_Act': None,
+        'CIO_Council': None,
+    })
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize("impacted_entity,agencies,agency_groups", [
+    (("",), (set(),), (set(),)),
```
tadhg-ohiggins: I think Python needs to give me a way to easily create empty tuples and lists with comprehensions.

reqs/tests/link_agencies_tests.py
```diff
@@ -0,0 +1,123 @@
...
+     ({"DHS", "DOD", "Justice", "GSA"},), (set(),)),
+    (("All agencies except Paperwork\nDHS\nGSA\nDOJ\nDOD",),
+     ({"DHS", "DOD", "Justice", "GSA"},), (set(),)),
+    (("department of homeland\ngeneral services\nJustice\nDefense",),
+     ({"DHS", "DOD", "Justice", "GSA"},), (set(),)),
+], ids=repr)
+def test_linking(populate_agencies, impacted_entity, agencies,
+                 agency_groups):
+    reqs = [mommy.make(Requirement, impacted_entity=_) for _ in
+            impacted_entity]
```
tadhg-ohiggins: I don't mind changing it, but there's a philosophical disagreement here in that I think it means “won't be using the variable outside the current context”…

reqs/tests/link_agencies_tests.py
tadhg-ohiggins: That's a better approach.

reqs/management/commands/link_agencies.py
cmc333333: That sounds like a fine policy ;) Let's document it so that we're consistent.

reqs/management/commands/link_agencies.py
cmc333333: Gotcha gotcha. Maybe we just modify the docstring, then?

Random thing I learned recently: you can access a many-to-many table via the `through` field. This means we can cut the number of queries here:
```python
Requirement.agencies.through.objects.all().delete()
Requirement.agency_groups.through.objects.all().delete()
```

reqs/tests/link_agencies_tests.py
cmc333333: I've got the most enjoyable dictionary on twitter on my side: https://www.merriam-webster.com/dictionary/sync

Doesn't matter to me, so long as our docs/scripts are up to date :)

tadhg-ohiggins:
@cmc333333 I addressed most of your comments in my most recent changes and think this is good to go, but I'll hold off on merging it until morning in case you want to have another look.

cmc333333:
@tadhg-ohiggins thanks for making many of those changes! Still good to go in my book.

PR #406, opened 2017-06-16
Add mechanisms to mass-create new versions

cmc333333:
As part of #374, this adds two new features:
* a `make_revisions` command to create a new revision of all instances of a requested model
* a mechanism to tag particular database migrations as needing a revision after running

The first extends Reversion's command to get some niceness around specifying particular models. The second hooks into Django's signals to know when migrations are finished and which were ran. It *may* be too cute but I'm thinking we'll forget to create the revisions when pushing. A potential stumbling block: if creating revisions (or migrating data in general) takes too long, we'll timeout (and fail) our auto-deploy.


cmc333333:
Thanks @tadhg-ohiggins

PR #408, opened 2017-06-19
Extend VersionAdmin with schema mismatch warnings.

cmc333333:
When users save models, a copy of all the model's fields is stored in a
version history. As the database schema changes, that version becomes less
compatible. We can warn the user when this occurs, though, by comparing the
fields that would serialize now to the fields stored in the database.

There's a nasty snag where the serialized field names don't quite match up
when comparing the names on an unsaved Django model. This works around that by
grabbing a model from the database and showing a slightly different warning
when that's not possible.

Pretty easy to change the message text if we don't like this version.

### DIFF
--- a/ereqs_admin/revision_admin.py
+++ b/ereqs_admin/revision_admin.py
@@ -1,0 +1,49 @@
+import json
+
+from django.contrib import messages
+from django.core import serializers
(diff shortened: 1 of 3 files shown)


cmc333333:
Thanks @tadhg-ohiggins

PR #407, opened 2017-06-19
[WIP] Implement mobile layout

theresaanna:
screen shot 2017-06-19 at 6 03 08 am
screen shot 2017-06-19 at 6 05 03 am

Issue #308, opened 2017-05-10
Fix issues raised by the Google Accessibility Developer Tools


theresaanna:
<img width="797" alt="screen shot 2017-05-10 at 1 19 04 pm" src="https://cloud.githubusercontent.com/assets/126935/25919350/5665cb4a-3583-11e7-9c83-adc49b39d25f.png">

cmc333333:
#407 and #409 solved all but two of these issues:

* [Severe] Elements with ARIA roles must ensure required owned elements are present (1)
An accessible autocomplete tool #g-accessibility pointed me to has the same issue, so I'm thinking this is a false positive.
* [Warning] These elements are focusable but either invisible or obscured by another element (1)
I think this is a false positive as it is visible through a transparent div.

I'd be in favor of closing this ticket as is.

yowill:
@theresaanna Also, this task too, can you update us on the status?

cmc333333:
@yowill I'll finish this one up.

cmc333333:
With the latest version of our app and accessibility tools, we have only one warning on all of our pages:

<img width="846" alt="screen shot 2017-09-29 at 11 24 28 am" src="https://user-images.githubusercontent.com/326918/31023311-71b89314-a509-11e7-9b86-d45c53b1b592.png">

There are, however, a few more warnings with HTML Codesniffer which should be resolved before closing this ticket.

PR #409, opened 2017-06-19
Accessibility tweaks

cmc333333:
Partial solution to #308:
* Tie labels to input elements via aria-label/labelledby
* Add contrast to autocomplete element, some unseen links
* Swap a css image and an image input w/ a button+img
* Add aria-labels for uninformative links.

This doesn't solve:
* [Severe] Elements with ARIA roles must ensure required owned elements are present (1)
    * I'm trying to get a sense from #g-accessibility whether or not this is actually significant --[an "accessible" autocomplete](https://haltersweb.github.io/Accessibility/autocomplete.html) has the same error.
* [Warning] The web page should have a title that describes topic or purpose (1)
    * This will be resolved w/ #407 (nice)
* [Warning] These elements are focusable but either invisible or obscured by another element (1)
    * I think this is a false positive as it _is_ visible

### DIFF
--- a/ui/__tests__/components/filters/fallback-view.test.js
+++ b/ui/__tests__/components/filters/fallback-view.test.js
@@ -6,6 +6,7 @@ import FallbackView from '../../../components/filters/fallback-view';
 
 describe('<FallbackView /> form', () => {
   const props = {
+    'aria-labelledby': 'some_id',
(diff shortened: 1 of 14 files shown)


cmc333333:
Thanks for reviewing, @theresaanna

PR #411, opened 2017-06-20
Remove the current search field from inputs.

### LINKED ISSUE #384: Text search switching tabs bug
1. Go to https://omb-eregs-demo.app.cloud.gov/requirements?req_text__search=something
2. Click "Policies"
3. Type "more" in the banner search
4. See that ui says "moresomething" and the params are: 

https://omb-eregs-demo.app.cloud.gov/policies?requirements__req_text__search=more&requirements__req_text__search=something&x=0&y=0

cmc333333:
As we want to keep the user on the same page/with the same filters when they
search, we create hidden input values to match their current query. We weren't
deleting the _existing_ search parameter when viewing policies (though we were
if viewing requirements).

Should resolve #384

### DIFF
--- a/ui/__tests__/components/search/search.test.js
+++ b/ui/__tests__/components/search/search.test.js
@@ -1,0 +1,33 @@
+import { shallow } from 'enzyme';
+import React from 'react';
+
+import Search from '../../../components/search/search';
+
+
+describe('<Search />', () => {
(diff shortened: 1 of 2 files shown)


cmc333333:
Thanks @theresaanna

PR #413, opened 2017-06-21
Fix bug around error pages, allow empty search

### LINKED ISSUE #412: Error in Non-JS empty topic search
1. Turn off JS
2. Go to the homepage
3. Click the plus next to an empty topic selection
4. See a 500

This indicates two problems:
1. `UserError`s are generating 500s rather than 401s
2. A query string shouldn't be required

cmc333333:
To resolve #412, this

1. Fixes error handlers (to test, try raising an error in one of the components)
2. Allows non-JS topic search to be empty


cmc333333:
Thanks @theresaanna!

PR #414, opened 2017-06-21
Add a blacklist of "masked" impacted entities.

cmc333333:
Per a request from OMB, we want to temporarily hide impacted entity meta data
when it is one of a handful of values. OMB wants to keep the information for
manual cleanup and analysis.

This also removes the default ("unknown") as it was part of the requested
blacklist.

@tadhg-ohiggins what do you think about this? It's pretty straight forward and achieves the requested ends. I'm not keen on the precedent, but not terribly opinionated.


tadhg-ohiggins:
I think doing this as a one-off as part of ongoing data cleanup is probably fine; if we find ourselves adding new values or doing the same with other fields then we should probably find a different approach.

cmc333333:
Sounds good, thanks @tadhg-ohiggins

PR #415, opened 2017-06-21
Remove Docker update from CI.

cmc333333:
The new build images on Travis include a current-enough version of Docker and
don't include `docker-engine` in their app repos.


cmc333333:
Thanks @micahtaylor !

PR #416, opened 2017-06-22
No results view

micahtaylor:


ui/components/pagers.jsx
```diff
@@ -10,6 +10,9 @@ export default function Pagers({ count }, { router }) {
 
   if (count === 0) {
     pageInt = 0;
+    return (
+      <div className="alert p1 m1 border">No requirements match your search, try removing some filters to see more results.</div>
```
micahtaylor: @theresaanna I could use your expertise on how to handle this better.

micahtaylor: I think I figured it out.

cmc333333:
So we're on the same page, this is not going to be merged in favor of #417, yes?

micahtaylor:
Yes!

Issue #236, opened 2017-04-17
Show number of results returned on requirements/policy view


cmc333333:
Why: Helps users understand how the filter mechanisms are working

Should probably include total count per latest mockups, too.

micahtaylor:
Is this count cover under these:
Policy listing - https://gsa.invisionapp.com/share/M6AU3YB49#/229078016_Policy_Listing
Req listing - https://gsa.invisionapp.com/share/M6AU3YB49#/229078015_Requirement_Listing

theresaanna:
Should this issue get rolled up into one something like "Implement requirements list view changes"? @micahtaylor @tadhg-ohiggins

micahtaylor:
I suppose it could.

tadhg-ohiggins:
Some screenshots of my current implementation:

Requirements, all:
![image](https://cloud.githubusercontent.com/assets/2626258/25973734/3b0af2e4-365a-11e7-94a8-677aa3194d56.png)

Requirements, filtered by topic:
![image](https://cloud.githubusercontent.com/assets/2626258/25973340/88b47d0a-3658-11e7-9500-a0bed8208f9f.png)

Requirements, no results:
![image](https://cloud.githubusercontent.com/assets/2626258/25973822/a87d626c-365a-11e7-8d4e-b7eacc05956b.png)

Policies, all:
![image](https://cloud.githubusercontent.com/assets/2626258/25973749/544a3a44-365a-11e7-9791-db681031f04a.png)

Policies, filtered by topic:
![image](https://cloud.githubusercontent.com/assets/2626258/25973786/6e5e3cf0-365a-11e7-9702-f6d5c5d030e9.png)

Policies, no results:
![image](https://cloud.githubusercontent.com/assets/2626258/25973805/89ddb924-365a-11e7-8bae-b7deda43a312.png)

nicoleslaw:
We discussed this in Slack, but I would recommend that we show the full number of policies on load, and then narrow that with the search.

After talking through it, I find a nice example with GOV.UK: 

`101,433 publications` on load
`3,075 publications containing technology` after filtering

https://www.gov.uk/government/publications

https://www.gov.uk/government/publications?keywords=technology&publication_filter_option=all&topics%5B%5D=all&departments%5B%5D=all&official_document_status=all&world_locations%5B%5D=all&from_date=&to_date=

micahtaylor:
Adding @carodew to run with this one so we can get consensus. 🎉  
[...]

PR #417, opened 2017-06-22
Show number of results in requirement and policy views

tadhg-ohiggins:
There are things to count. /
Count them, and birds above, /
And beams of sunlight.

Addresses #236.


ui/__tests__/components/thing-counters.test.jsx
```diff
@@ -0,0 +1,46 @@
...
+import ThingCounter from '../../components/thing-counters';
+import mockRouter from '../util/mockRouter';
+
+
+describe('<ThingCounter />', () => {
+  it('states zero results when there are no results', () => {
+    const context = { router: mockRouter() };
+    const result = shallow(<ThingCounter count={0} chunk={0} singular={ "requirement" } plural={ "requirements" }/>, { context });
+    expect(result.text()).toMatch(/No requirements match your search, try removing some filters to see more results./);
+    expect(result.find("div").length).toEqual(1);
```
cmc333333: NBD, but there's a `expect(...).toHaveLength()` function, if you like that more.

ui/__tests__/components/thing-counters.test.jsx
```diff
@@ -0,0 +1,46 @@
...
+import React from 'react';
+
+import ThingCounter from '../../components/thing-counters';
+import mockRouter from '../util/mockRouter';
+
+
+describe('<ThingCounter />', () => {
+  it('states zero results when there are no results', () => {
+    const context = { router: mockRouter() };
+    const result = shallow(<ThingCounter count={0} chunk={0} singular={ "requirement" } plural={ "requirements" }/>, { context });
```
cmc333333: I _think_ you can just do `singular="requirement"` here. You may also be able to do `count=0` rather than using so many curlies.

ui/__tests__/components/thing-counters.test.jsx
```diff
@@ -0,0 +1,46 @@
...
+    const context = { router: mockRouter({ query: { topic: 'something' } }) };
+    const result = shallow(<ThingCounter count={1} chunk={0} singular={ "requirement" } plural={ "requirements" }/>, { context });
+    expect(result.text()).toMatch(/1 requirement matches your search./);
+    expect(result.find("div").length).toEqual(1);
+    expect(result.find(".alert .p1 .m1 .border").length).toEqual(0);
+  });
+  it('states two results when there are two filtered results', () => {
+    const context = { router: mockRouter({ query: { topic: 'something' } }) };
+    const result = shallow(<ThingCounter count={2} chunk={0} singular={ "requirement" } plural={ "requirements" }/>, { context });
+    expect(result.text()).toMatch(/2 requirements match your search./);
```
cmc333333: 2 requirements, ah ah.

ui/__tests__/components/thing-counters.test.jsx
```diff
@@ -0,0 +1,46 @@
...
+    expect(result.find("div").length).toEqual(1);
+    expect(result.find(".alert .p1 .m1 .border").length).toEqual(0);
+  });
+  it('states one total result when there is one total result', () => {
+    const context = { router: mockRouter() };
+    const result = shallow(<ThingCounter count={1} chunk={0} singular={ "requirement" } plural={ "requirements" }/>, { context });
+    expect(result.text()).toMatch(/1 requirement./);
```
cmc333333: I may be missing something, but is this test different than 'states one result when there is one filtered result'? I think the period here matches any character.

Maybe you wanted a mock router with query params?

ui/__tests__/components/thing-counters.test.jsx
```diff
@@ -0,0 +1,46 @@
...
+    expect(result.find("div").length).toEqual(1);
+    expect(result.find(".alert .p1 .m1 .border").length).toEqual(0);
+  });
+  it('states two total results when there are two two results', () => {
```
cmc333333: Did you want "two total results" here?

Same question re: how this differs from 'states two results when there are two filtered results'

ui/components/policies/policies-view.jsx
```diff
@@ -2,10 +2,16 @@ import React from 'react';
 
 import Pagers from '../pagers';
 import Policy from './policy-view';
+import ThingCounter from '../thing-counters';
 
 export default function PoliciesView({ policies, count, topicsIds }) {
+  const chunk = 25;
```
cmc333333: It looks like the chunk size isn't used in `ThingCounter` -- can we remove it throughout?

ui/components/thing-counters.jsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+
+export default function ThingCounter({ count, singular, plural }, { router }) {
+  const query = router.location.query;
+  const queryKeys = Object.keys(query);
+  const noun = count === 1 ? singular : plural;
+  const verb = count === 1 ? 'matches' : 'match';
+  const classes = count === 0 ? 'alert p1 m1 border' : '';
+
+  const unfiltered = queryKeys.length === 0 || (queryKeys.length === 1 && queryKeys.indexOf('page') !== -1);
```
cmc333333: This is a good, descriptive variable name. The logic feels a little hard to follow, so I'll throw out some odds and ends that may help in case you aren't familiar:
* `Object.assign({}, router.location.query)` would make a shallow copy of the query
* `del someObject.key` follows Python expectations for deleting an attribute. It's a noop if `key` isn't present
* `someArray.includes(key)` returns a bool and should be polyfilled for browser support already

Maybe we do something like:
```js
const queryWithoutPage = Object.assign({}, router.location.query);
del queryWithoutPage.page

const unfiltered = Object.keys(queryWithoutPage).length > 0
```
That doesn't feel particularly elegant, either...

ui/components/thing-counters.jsx
```diff
@@ -0,0 +1,45 @@
...
+  const searchText = unfiltered ? '.' : ` ${verb} your search.`;
+
+  const noMatches = `No ${noun} ${verb} your search, try removing some filters to see more results.`;
+  const someMatches = `${count} ${noun}${searchText}`;
+
+  const text = count === 0 ? `${noMatches}` : `${someMatches}`;
```
cmc333333: Since we're taking the whole string, we can simplify this to:
```js
const text = count === 0 ? noMatches : someMatches;
```

ui/components/thing-counters.jsx
```diff
@@ -0,0 +1,45 @@
...
+  return (
+    <div className={classes}>
+      {`${text}`}
```
cmc333333: Similarly, this can be `{text}`

ui/__tests__/components/thing-counters.test.jsx
tadhg-ohiggins: Yup, I meant “two total results”.

When the result isn't filtered, i.e. when there are no query parameters, then there's no “that match your search” rider to the sentence.

ui/components/policies/policies-view.jsx
tadhg-ohiggins: I missed that one, and yes, it can go.

ui/components/thing-counters.jsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+
+export default function ThingCounter({ count, singular, plural }, { router }) {
+  const query = router.location.query;
+  const queryKeys = Object.keys(query);
+  const noun = count === 1 ? singular : plural;
+  const verb = count === 1 ? 'matches' : 'match';
+  const classes = count === 0 ? 'alert p1 m1 border' : '';
```
cmc333333: I think we'll need to include the associated SASS edits, too: https://github.com/18F/omb-eregs/pull/416/files#diff-a17de2846fcbead1087a20fc2730d26d

ui/components/thing-counters.jsx
tadhg-ohiggins: Probably! I find the JSX syntax somewhat awkward, as you can likely tell…

ui/__tests__/components/thing-counters.test.jsx
cmc333333: Aha, I mixed up the two different types of test. The first two cover the extra query string, my bad.

ui/__tests__/components/thing-counters.test.jsx
cmc333333: Apologies here as well. The period might still be capturing the wrong thing, but I mixed up the crux of the issue.

ui/components/thing-counters.jsx
tadhg-ohiggins: I decided on:
` const unfiltered = queryKeys.length === 0 || (queryKeys == ['page']);`

The earlier version was overcomplicated, thanks for pointing that out!

ui/components/thing-counters.jsx
tadhg-ohiggins: Good point, I added that in my last commit.

PR #419, opened 2017-06-23
Add the "public" field to requirements' admin.

cmc333333:


cmc333333:
Thanks @tadhg-ohiggins

PR #423, opened 2017-06-30
Internet Explorer support

theresaanna:
Everything is usable back to IE7. Some aesthetics are a little janky through IE11, but only slightly.

IE7-9 look fairly janky but all features work and there are no reasons someone shouldn't be able to access and use all the content.
IE10 still doesn't have client-side JS happening, but looks a lot better. Form button height is a little off.
IE11 only has form button height issues.
Edge looks great.

It seems like feature detection couldn't get us the level of precision we need on the versions, so I've gone the useragent sniffing route. I realize it's not ideal. Am open to suggestions.

### DIFF
--- a/ui/ie.js
+++ b/ui/ie.js
@@ -26,8 +26,10 @@ function detectIE() {
 }
 
 window.onload = function() {
-  var browser = detectIE();
-  if (browser) {
-    document.body.className = 'ie';
+  var ieVersion = detectIE();
+  if (ieVersion) {
+    if (ieVersion <= 9) {
+      document.body.className = 'legacy-ie';
+    }
   }
 }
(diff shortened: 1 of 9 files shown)


ui/ie.js
```diff
@@ -0,0 +1,54 @@
...
+    // Edge (IE 12+) => return version number
+    return parseInt(ua.substring(edge + 5, ua.indexOf('.', edge)), 10);
+  }
+
+  // other browser
+  return false;
+}
+
+window.onload = function() {
+  var ieVersion = detectIE();
```
cmc333333: Can we wrap this in a try-catch? I could imagine the user agent string (which can technically be anything) parse failing and killing the rest of the JS.

theresaanna:
@cmc333333 Oh, this makes me very sad as well. I'm just not sure how else to achieve what we would need. I will implement that try/catch, good call.

PR #427, opened 2017-07-24
Switch comment style, per airbnb guide.

cmc333333:
Our nightly build/deploy is failing due to a linting error. npm's `shrinkwrap` doesn't capture the dev environment; I'm 90% positive that the airbnb style guide moved a bit from under us and didn't keep follow semantic versioning, but I didn't look too hard.

In any event, this should fix.


cmc333333:
Thanks @micahtaylor and @tadhg-ohiggins !

PR #430, opened 2017-07-31
Add header/footer/disclaimer

cmc333333:
This builds off @micahtaylor's work to add a beta dropdown in the header, a footer, and updated disclaimer text on the home page. I believe these are the necessary changes for launch (sans removing password).


cmc333333:
Thanks both!

PR #431, opened 2017-08-01
Text tweaks for launch.

cmc333333:
Per @nicoleslaw


cmc333333:
Thanks @tadhg-ohiggins

PR #438, opened 2017-08-04
Tweaks from Matt

cmc333333:
Extends #437 with linting fixes.


mattbailey0:
@tadhg-ohiggins let me tell you about committees

cmc333333:
Thanks all!

PR #444, opened 2017-08-17
add a space

wslack:
fixing a typo


wslack:
ty @tadhg-ohiggins!

tadhg-ohiggins:
@wslack JSX strips whitespace between lines when the next line opens a new element, which is why you saw the original issue, and also doesn't allow trailing spaces, which is why the tests failed.

So you have to use interpolation syntax to retain/insert the whitespace (at least, that's the way I found to do it), even though it looks rather ugly.

Thanks for catching the missing space!

Issue #455, opened 2017-09-05
Nightly re-deploy via CircleCI


cmc333333:
We're currently deploying nightly via Travis. As part of #451, we'll need to replace that with a CircleCI solution. We can crib off atf-eregs here.

cmc333333:
Resolved in #462

PR #456, opened 2017-09-05
Start CircleCI conversion

cmc333333:
In this first step, we run all of our tests in CircleCI and begin deploying with it. Most of the code is copied from the Travis file or atf-eregs.

We need to get a nightly build running (#455) before we can turn off Travis entirely.


cmc333333:
Thanks @tadhg-ohiggins

PR #457, opened 2017-09-05
Use root-relative url for search when unsure.

cmc333333:
When a user is on a non-policies/requirements page (the home page,
search-redirect pages, the privacy page), the header search bar needs to
redirect the user to the "requirements" page. Unfortunately, we were using a
relative url, which meant deeply nested pages (search-redirects) would
redirect to, e.g.  `/search-redirect/topics/requirements/` rather than
`/requirements/`.

This resolves by using a root-relative url.

### DIFF
--- a/ui/components/search/search.jsx
+++ b/ui/components/search/search.jsx
@@ -16,7 +16,7 @@ export default class Search extends React.Component {
 
   actionPath() {
     const path = this.context.router.location.pathname;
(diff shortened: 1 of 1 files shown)


cmc333333:
Thanks @tadhg-ohiggins. I'm going to rebase to pull in the circleci stuff (which should resolve the error).

PR #466, opened 2017-09-08
Add Python code coverage metrics.

### LINKED ISSUE #205: Can we concatenate `lcov` reports so that Code Climate can report on both JS and Python?
Code Climate doesn't support sending reports for multiple languages. Can we use this common format to condense both the JS and Python test data and send that to Code Climate?

cmc333333:
Code Climate's created a new code coverage reporter which allows results from
multiple sources to be submitted. This uses that binary to submit coverage
info from Python and Node.

Resolves #205

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -74,6 +74,7 @@ jobs:
             ./cc-test-reporter upload-coverage --input coverage/codeclimate.json
 
       - deploy:
+          name: Deploy
           command: |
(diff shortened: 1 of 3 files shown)


cmc333333:
Node the coverage bump -- that's due to reporting the Python code, too.

cmc333333:
Thanks @tadhg-ohiggins

PR #469, opened 2017-09-12
Allow Circle builds to be restarted.

cmc333333:
Code Climate's coverage reporter will fail if submitting coverage for the same
commit twice. We'll be doing that relatively frequently, however, in both our
nightly builds and if we ever want to restart a failed deploy.

This resolves by allowing the upload step to fail.

For verification, compare the "Submit Coverage" steps in
https://circleci.com/gh/18F/omb-eregs/70 and https://circleci.com/gh/18F/omb-eregs/71


cmc333333:
Thanks @tadhg-ohiggins; I'll rebase then merge.

PR #471, opened 2017-09-12
Add a squash commit to migrations.

cmc333333:
As a step towards removing taggit completely, we can begin to process of
squashing our migrations (which reference taggit). We have relatively
complicated migrations thus far to accommodate migrating data between multiple
schemas. Luckily, most of those don't matter with a fresh database, so our
squashed migrations is _largely_ a straight schema dump.

The two exceptions here are:
1. The "all agencies" view/fields are created by hand (Django doesn't know how
to create views)
2. There's a gin index on the search field that's similarly created by hand
(for the same reason).

The next step is to roll this migration out to everyone. Once everyone has
this migration (and runs `migrate`) we can remove migrations 0001 through 0040
(and taggit).


cmc333333:
Thanks @tadhg-ohiggins.

* history of our code is in git
* history of our database migrations is in code
* history of our content is in the database
* history of our logs is in Kibana
* history of our issue tracking is in Github
* history of our conversations is in Slack

We clearly need some more layers

PR #473, opened 2017-09-13
Move source-map-support back to a production lib.

cmc333333:
We had moved source-map-support out of the production libs because it was only
referenced in webpack. Unfortunately, that reference actually inserts a
_string_ into the resulting file, which imports source-map-support, meaning it
_is_ needed for production.

This will fix broken deploys (locally, we have the dev dependencies
installed).


cmc333333:

PR #474, opened 2017-09-13
Allow multiple managing offices

### LINKED ISSUE #405: ability to add two offices as “Managing Offices”
Feature request from Michelle with explanation –  

OFPP partners frequently with other OMB offices and would benefit from such a feature

cmc333333:
In practice, policies need to be "managed" by multiple offices. We don't
display this field anywhere but the admin, so there's not too much code
cleanup after the data migration.

Resolves #405 

Looks like
screen shot 2017-09-13 at 12 50 08 pm


cmc333333:
Thanks @tadhg-ohiggins . I'll rebase and merge.

PR #475, opened 2017-09-13
Prevent multiple migrations in prod.

### LINKED ISSUE #422: Account for multiple API instances when migrating
When the API servers start up, they migrate the database. This doesn't lead to conflicts when only one instance is starting, but can lead to a worker dying if there are multiple instances running.

We should modify our startup script to only migrate the database a la [epa-notice](https://github.com/18F/epa-notice/blob/0b621b0552b5031e9ec8a19444de88357d38ac79/cf-startup.sh#L4)

cmc333333:
We run multiple `api` servers in production, but we only want one of them to
be responsible for migrating our database (lest one instance's migration will
prevent the other from starting). This changeset adds a management command
which migrates the database only on the first instance when on cloud.gov, and
every time when not (e.g. local development).

Should resolve #422

### DIFF
--- a/api/ereqs_admin/management/commands/cf_migrate.py
+++ b/api/ereqs_admin/management/commands/cf_migrate.py
@@ -1,0 +1,12 @@
+from cfenv import AppEnv
+from django.core import management
+from django.core.management.base import BaseCommand
+
+
+class Command(BaseCommand):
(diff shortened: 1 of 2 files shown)


cmc333333:
Thanks @tadhg-ohiggins ! I'll rebase then merge.

PR #476, opened 2017-09-15
Separate two apps more thoroughly

cmc333333:
In working on the potential Next.js migration, I hit an early snag in where files are located. We've been running everything from the root of the project, which contains a mix of Python and Node scripts (though _most_ of the Node stuff has been in the `ui` directory). This finishes moving all of the Node code into `ui`, moves all of the Python code into `api` (for parity) and executes all commands within their respective subdirectories.

In addition to resolving the Next.js issue, this:
* Clearly delineates which parts of code are used for which app
* Allows our cloud.gov deploys to be smaller (as we'll only send up app-specific code)
* Makes our code layout match our system diagram

All Docker commands are still ran from the root, thought it might be sensible to change that in the future.

Easiest to review changeset-by-changeset.


cmc333333:
Thanks @tadhg-ohiggins !

PR #477, opened 2017-09-15
404 page and homepage stars

cmc333333:
This updates the design of the 404 page. Unfortunately, our existing strategy around accessing the current url (which relies on react-router) doesn't work well for 404 pages. To accommodate, this effectively mocks that extra context out. That isn't a great long term strategy, but I suspect a longer term strategy will be worked out when refactoring for Next.js

It also adds a star-based divider to the homepage (per the latest designs) and extracts some of those styles for use on the 404 page.

## 404
screen shot 2017-09-20 at 12 13 00 pm

## Homepage
screen shot 2017-09-20 at 12 27 29 pm

### DIFF
--- a/ui/__tests__/server-render.test.js
+++ b/ui/__tests__/server-render.test.js
@@ -1,0 +1,14 @@
+import { render404 } from '../server-render';
+
+describe('render404', () => {
+  it('includes the search bar', () => {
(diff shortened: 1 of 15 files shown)


cmc333333:
Awe yeah. Thanks @micahtaylor

PR #479, opened 2017-09-15
Add spaces to comments.

cmc333333:
The latest versions of eslint catch these issues better, so our builds are
failing.


cmc333333:
Thanks @tadhg-ohiggins

PR #483, opened 2017-09-20
Move top-level components into a "pages" directory

cmc333333:
Following Next.js's lead, this moves our application entry points into a "pages" directory, and wraps each with the same header and footer. Along the way, it renames the `App` component and uses JSX instead of JS in the policy and requirement pages.
That said, it's largely a bunch of renames.


cmc333333:
Thanks @tadhg-ohiggins !

PR #484, opened 2017-09-22
Lift api calls to top-level page components

cmc333333:
Rather than having each component individually describe what data it needs, this pulls API calls to the outer most component (the "pages"). This means that pages will load all at once rather than bit by bit and aligns better with the Next.js flow.


cmc333333:
Thanks @tadhg-ohiggins! I'll rebase and merge.

PR #485, opened 2017-09-22
Remove logic, add more text to header disclaimer.

### LINKED ISSUE #480: Add clarification that the site is a work-in-progress
We've been getting some feedback on the site, but many people do not understand that the site is not complete and that not all policies have been entered yet.

> I like the ease of use of the website, but it appears many policies are missing. For example, I searched both "buy american' and "regulatory reform" and there were no results. Would have assumed at least this memo would have come up for Buy Amercian and this one for Reg Reform. If you're asking for feedback on your competition, I'd still go with a Google search.

> I just tried to find some information on “FITARA” but it wasn’t  in the list of search items. 

> Also, is this pulling from test data or real data?  When I chose category management, it only showed two items and there have been numerous historical memos issued on strategic sourcing which should pull for this topic.  I also see no category for Federal Acquisition Certifications and there have been many memos issued on this topic.

It looks like we need to make it more obvious that this is not all of the policies, and we're working on that.  Perhaps a banner at the top of every page, or at least a very unignorable note near the search & results?

cmc333333:
We want to always display a description of the incomplete nature of the app. Probably easier to see the final version than review diff mode.

Should resolve #480 

Looks like
warning


cmc333333:
Thanks @micahtaylor! I'll rebase with the lint fix and then merge.

PR #495, opened 2017-09-27
Add favicons

tadhg-ohiggins:
Show icons to all.
Under the tabs, Mario
Jumps but falls just short.

Adds the assets for favicon images, and the requisite code for them.


ui/components/html.jsx
```diff
@@ -35,6 +35,58 @@ function newRelicTag() {
...
+      name="msapplication-TileImage"
+      content="mstile-144x144.png"
+    />,
+    msSquares,
+    <meta
+      name="msapplication-wide310x150logo"
+      content="mstile-310x150.png"
+    />,
+  ]);
+}
```
cmc333333: NBD, but React might complain about wanting specific `key` attributes for each element as a rendering optimization. Easy enough to use their index (though you'll need to temporarily disable the linter check for that).

ui/components/html.jsx
```diff
@@ -35,6 +35,58 @@ function newRelicTag() {
...
+  );
+
+  return ([
+    appleIcons,
+    icons,
+    <meta name="application-name" content="&nbsp;" />,
+    <meta name="msapplication-TileColor" content="#FFFFFF" />,
+    <meta
+      name="msapplication-TileImage"
+      content="mstile-144x144.png"
```
cmc333333: Does this need to have the `/static/img/favicon/` prefix?

ui/components/html.jsx
```diff
@@ -35,6 +35,58 @@ function newRelicTag() {
...
+    />,
+    msSquares,
+    <meta
+      name="msapplication-wide310x150logo"
+      content="mstile-310x150.png"
```
cmc333333: Same prefix question here.

ui/components/html.jsx
tadhg-ohiggins: These both do, thanks for catching that!

PR #500, opened 2017-09-29
Use policy number when displaying policy filter.

### LINKED ISSUE #493: Validate/Adding display of policy ID to the UI
As a power user, I want to be able to see the policy ID so that I can cite and track it.

## AC:
- [ ] If not, find a place, and mock it up
- [ ] design for both documents that have a policy ID and those that don't
- [ ] Display it on the policy profile page

cmc333333:
Resolves #493 by adding the policy number to policy-based filters.

Here's how things look with and without a policy number:

## Policy with number
screen shot 2017-09-29 at 11 00 56 am
screen shot 2017-09-29 at 11 00 39 am

## Policy without number
screen shot 2017-09-29 at 11 01 24 am
screen shot 2017-09-29 at 11 01 12 am

### DIFF
--- a/ui/components/filters/existing-container.js
+++ b/ui/components/filters/existing-container.js
@@ -74,16 +74,16 @@ export default function ExistingFiltersContainer({
     }));
 
   const policyIds = policies.map(policy => policy.id);
-  const policyFilters = policies.map(policy => React.createElement(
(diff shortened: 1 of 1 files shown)


cmc333333:
Thanks @micahtaylor !

Issue #258, opened 2017-04-18
Admin view slowness


tadhg-ohiggins:
When viewing a policy with many related requirements (such as policy 1, “25 Point Implementation Plan To Reform Federal Information Technology Management”) the local development Docker instance takes a significant amount of time to load the page for that policy (over 30 seconds).

It's possible that a future Docker caching feature will improve this.

That page does over 60 database queries, and so while it's fine in dev and prod, we might want to improve that also. @cmc333333 suggested `select_related` or `prefetch_related` in Slack as possible avenues to follow.

cmc333333:
This is almost instantaneous when we're in prod mode; may be another toolbar issue.

PR #501, opened 2017-09-29
Loosen consistency guarantees in Docker.

cmc333333:
Docker for Mac can be very slow when performing file access on a shared
directory due to consistency guarantees. This tweaks our mounted directories
so that there can be a slight delay between synchronization as the benefit of
much faster reads.

This appears to be a solution to #258 in my experience, dropping local admin
load time dropping to roughly what we see in production.

We had tried this solution before, but support wasn't fully baked.


cmc333333:
Thanks @tadhg-ohiggins; I'll rebase then merge.

PR #508, opened 2017-10-03
Beginning of tree-based documents

cmc333333:
This pulls in (and refactors) some code to represent a document as a nested set in the database, and a networkx graph in memory. I've tried to split it into reasonably specific changesets for ease of review.


api/document/models.py
```diff
@@ -19,3 +23,57 @@ class Meta:
...
+        edges = [(sort_order, idx) for _, idx, sort_order
+                 in self.tree.out_edges(self.identifier, data='sort_order')]
+        for _, identifier in sorted(edges):
+            yield self.__class__(self.tree, identifier=identifier)
+
+    def next_number(self, tag):
+        child_tag_counts = Counter(c.model.tag for c in self.children())
+        return str(child_tag_counts[tag] + 1)
+
+    def add_child(self, tag: str, number: Optional[str] = None, **attrs):
```
tadhg-ohiggins: I'm slightly confused by `number` here; why is it a string rather than an integer? Perhaps the question is also “why is it called number?”, which would also apply to the name of the `DocNode` property.

cmc333333: The idea is that we need to distinguish between multiple of the same tag within a parent, e.g. multiple sections within a subpart, multiple paragraphs within a section. The tricky bit is that, to retain semantic value, these might not always be integers. For example, we'd want `subpart_A`, `par_b__subpar_4__subsubpar_iv`, and similar. The term "number" comes from Akoma Ntoso, but we're not quite using it the way they are, so we can definitely rename. What do you think makes sense?

`marker` could make sense, but then it'd conflict with the text "(a)"/"Section 12"/etc., should we ever want to add that separately. Maybe `tag_idx`?

tadhg-ohiggins: What about `identifier`?

cmc333333:
Per convo with @tadhg-ohiggins, I'm going to
* rename `tag` to `node_type`
* rename `number` to `type_emblem`
* add comments for each database field to indicate an example
* add comments to the `next_number()` method.

I'll rebase with those changes to avoid creating a db migration.

cmc333333:
Thanks @tadhg-ohiggins! I'll rebase then merge.

PR #510, opened 2017-10-03
Add Python test watcher

tadhg-ohiggins:
Manual test starts
Should be rare, like a cat
Of silver and red.

`docker-compose run --rm ptw` will run the Python tests, then wait for file changes, which will kick off the tests again.
This removes the container startup time from test execution.


api/.docker/watch_tests.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+ptw ereqs_admin omb_eregs reqs
```
cmc333333: What do you think about passing through call arguments:
```bash
ptw ereqs_admin omb_eregs reqs "$@"
```
? This would allow us to set flags.

Alternatively, since this is just a one liner, it could be the entrypoint for the docker command. That'd also allow the args to be passed in.

docker-compose.yml
```diff
@@ -155,5 +155,11 @@ services:
     entrypoint: 'psql -h persistent_db -U postgres'
     depends_on: ['persistent_db']
 
+  watch:
```
cmc333333: Would you mind changing this to something more specific like `ptw` to distinguish it from the Node version?

api/.docker/watch_tests.sh
tadhg-ohiggins: Good idea, I'll add that.

docker-compose.yml
tadhg-ohiggins: Sure.

tadhg-ohiggins:
Thanks @cmc333333 !

PR #511, opened 2017-10-04
Handle navbar search clientside if possible

### LINKED ISSUE #506: AJAXify search
As an agency user, I don't want to wait for full page loads.

## AC
* Ensure full-text search (currently, in the header) triggers an AJAX load rather than full-page.
* Ensure the search continues to work if JS is not present.

jeremiak:
If the Javascript is executed on the client, then we should intercept the form's submit event and do a client side re-routing. If the JS isn't executed, use the form action.

Closes #506

### DIFF
--- a/ui/__tests__/components/search/search.test.js
+++ b/ui/__tests__/components/search/search.test.js
@@ -3,8 +3,32 @@ import React from 'react';
 
 import { Search } from '../../../components/search/search';
 
-
 describe('<Search />', () => {
+  describe('actionPath()', () => {
+    it('returns /requirements/ by default', () => {
+      const router = { pathname: '/no-such-path', query: {} };
+      const rendered = shallow(<Search router={router} />);
+      const actual = rendered.instance().actionPath();
+      expect(actual).toEqual('/requirements/');
+    });
+    it('returns the pathname if "policies" is in the pathname', () => {
+      const router = { pathname: '/policies', query: {} };
+      const rendered = shallow(<Search router={router} />);
+      const actual = rendered.instance().actionPath();
+      expect(actual).toEqual('/policies');
+    });
+  });
+
+  describe('handleChange()', () => {
+    it('sets the "term" value of state', () => {
+      const router = { pathname: '/policies', query: {} };
+      const rendered = shallow(<Search router={router} />);
+      const input = rendered.find('input[type="text"]');
+      input.simulate('change', { target: { value: 'fake-value' } });
+      expect(rendered.state().term).toEqual('fake-value');
+    });
+  });
+
   it('creates hidden fields for query parameters', () => {
     const router = {
       pathname: '/policies',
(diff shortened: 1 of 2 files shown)


ui/components/search/search.js
```diff
@@ -18,21 +27,41 @@ export class Search extends React.Component {
...
+    const { value } = e.target;
+    this.setState({ term: value });
+  }
+
+  handleSubmit(e) {
+    e.preventDefault();
+    const { router } = this.props;
+    const { term } = this.state;
+    const pathname = this.actionPath();
+    const query = { [this.inputName()]: term };
```
cmc333333: I think we're losing the hidden fields here. The scenario we want to allow:
* User has filtered the requirements/policies by a topic (or multiple)
* User types something in the search bar at the top of the page
* Results should be _refined_ in the sense that the existing topics should remain.

Only annoying bit here is that we want to reset the page number.

ui/components/search/search.js
```diff
@@ -18,21 +27,41 @@ export class Search extends React.Component {
...
+          className="mb0 flex items-center"
+          onSubmit={this.handleSubmit}
+        >
           <input
             aria-label="Search term"
             name={this.inputName()}
             type="text"
             placeholder="Search"
             className="search-input p1 gray-border"
+            onChange={this.handleChange}
```
cmc333333: In an earlier iteration, we had set the value to `this.state.term`, but I'm not sure if there was a benefit. Is there any risk of "redrawing" or similar that'd lose the user's value?

ui/__tests__/components/search/search.test.js
```diff
@@ -3,8 +3,33 @@ import React from 'react';
...
+    });
+  });
+
+  describe('handleChange()', () => {
+    it('sets the "term" value of state', () => {
+      const router = { pathname: '/policies', query: {} };
+      const rendered = shallow(<Search router={router} />);
+      const event = { target: { value: 'fake-value' } };
+      const instance = rendered.instance();
+      instance.handleChange(event);
```
cmc333333: Just curious: why do we call the method directly here, but use `simulate` later?

ui/components/search/search.js
jeremiak: Ah, got it now, thanks for the explanation. Will fix and add a test case

ui/components/search/search.js
jeremiak: Yes, if we set the value as an attribute from `this.state` (which is what a lot of React docs, community and official, recommend we do) then everytime the event fires, render will be called and the result will be different, resulting in ReactDOM updating the value of the field. There isn't too much of a risk where we'll lose the user's value as the above render sequence is triggered when a user changes the value of the field. However, I have noticed that it can add a small lag to the field interaction which makes the website feel a bit less professional.

ui/__tests__/components/search/search.test.js
jeremiak: No, no real reason but now that you mention it, using `simulate` gives us an added benefit of ensuring that the event handlers are set up with the proper elements. I'll start using that pattern from here on out.

PR #512, opened 2017-10-04
Add the two new test commands to readme

rememberlenny:
# What

Adding the two test commands added recently to the ReadMe.

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,15 +191,16 @@ We have unit tests for the API/admin (Python) and for the React-based frontend
 
 For Python unit tests, run:
 ```sh
-docker-compose run --rm py.test
 docker-compose run --rm flake8  # linting
-docker-compose run --rm ptw
+docker-compose run --rm py.test # run tests once
+docker-compose run --rm ptw     # watch command to run tests whenever a file changes
 ```
 
 For JS unit tests, run:
 ```sh
-docker-compose run --rm npm test
-docker-compose run --rm webpack       # lints (and builds)
+docker-compose run --rm webpack             # lints (and builds)
+docker-compose run --rm npm test            # run tests once
+docker-compose run --rm npm run test:watch  # watch command to run tests whenever a file changes
 ```
 
 We also have a suite of integration tests, which are relatively complicated to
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -193,11 +193,13 @@ For Python unit tests, run:
 ```sh
 docker-compose run --rm py.test
 docker-compose run --rm flake8  # linting
+docker-compose run --rm ptw
 ```
 
 For JS unit tests, run:
 ```sh
 docker-compose run --rm npm test
+docker-compose run --rm npm run test:watch
```
jeremiak: Looks good to me, maybe we should include a comment about what the difference between the two of these is? Or is it somewhat obvious from the way the command is structured?

rememberlenny: @jeremiak I think thats a good idea. Ill add that.

cmc333333:
Something to consider here, both "watch" commands hang the terminal, so if someone's copy-pasting, they won't complete their paste. Would it make sense to move them to a difference fenced block? Alternatively, we could add a comment to the `py.test` and `npm test` lines that says something like
```
docker-compose run --rm py.test   # or ptw for continuous feedback
```

rememberlenny:
@jeremiak Updated

rememberlenny:
I like what you are saying. Im going to add two more comments.

--
*Leonard Bogdonoff*
rememberlenny@gmail.com
Blog <http://blog.rememberlenny.com> | Portfolio <http://rememberlenny.com/>
 | @rememberlenny <http://twitter.com/rememberlenny>
949-322-8496

rememberlenny:
@cmc333333 Updated.

README.md
```diff
@@ -191,15 +191,15 @@ We have unit tests for the API/admin (Python) and for the React-based frontend
...
+docker-compose run --rm ptw # watch command to run tests whenever a file changes
 ```
 
 For JS unit tests, run:
 ```sh
-docker-compose run --rm npm test
-docker-compose run --rm npm run test:watch # run tests whenever a file changes
+docker-compose run --rm npm test # run tests once
+docker-compose run --rm npm run test:watch # watch command to run tests whenever a file changes
 docker-compose run --rm webpack       # lints (and builds)
```
cmc333333: I think we can alleviate my concern by moving the `webpack` line above the `npm` commands -- then the final command will hang, but there's nothing after it, so that's okay.

Does that make sense? We've got non-devs running these commands, so I'm trying to think through potential pitfalls for them.

Issue #504, opened 2017-10-03
JS/React Audit


cmc333333:
As a modern frontend developer, I want to ensure the project is following best practices around React, building, and testings.

## AC
* Explore the existing codebase, looking for violations of best practice
* Write tests where big holes exist
* Refactor code to be more testable and in line with best practices

Our code base isn't huge, but it's also not tiny, so we don't expect this to be 100% complete. We'd like examples of the right way to do things, though.

jeremiak:
In addition to increasing test coverage, I came up with an initial list of things though as I go through the tasks I'll make notes of opportunities for further refactoring.

Renaming/restructuring
* Ensure that there is only one component in each file to make it easier to know how components work together and where to find them. Ideally, the file name should be similar to the component name
* Remove `-view` from some filenames for consistency (`policies-view.js`, `policy-view.js`) and its mostly redundant
* Move the *Tab components out of the /pages ([`<PoliciesTab />`](https://github.com/18F/omb-eregs/blob/master/ui/pages/requirements.js#L14-L38) in requirements, for example) and into the component directory structure
* Create a /util directory to hold helper utilities so that the top level of /ui is all configuration and its easier to understand where to start
* Create an API utility, which exports { endpoints, queries } making it clearer how these things are related to each other
* Investigate removing *Container from the names of the react components in /pages. In React, Containers are frequently higher order components but below the concept of a page

New components
* Refactor the *Tab components (from the /pages components)) to be a more generic, shared component if possible. How do the current *Tab components differ from `tab-view.js`?
* A `<Link />` component that will wrap around `<ExternalLink />` and `{ Link }` from the `routes.js` - can fix https://github.com/18F/omb-eregs/issues/505 along the way
* I found `pages/search-redirect.js` a bit hard to follow, and the actual component is named `<LookupSearch />` which I confused with `lookup-search.js`. I'd like to make this component more explicit and clearer how this component is related to the other search components like `lookup-search.js` and `/components/search`

If you'd like, I can just create an issue for each one of these as I tackle them to more clearly document the scope of each piece of work.

cmc333333:
[...]

PR #516, opened 2017-10-04
One component per file

jeremiak:
Restructured some of the component files to only have on component in each file. This enables a few things:

1. It is easier to take inventory of all of our component and component names to ensure they are consistent
2. It is easier to know where to find a given component
3. It helps determine what can be DRYed up and refactored

Along the way, I also converted use of `React.createElement()` to JSX where I can ran into it.

Part of #504

### DIFF
--- a/ui/__tests__/components/requirements/policy-link.test.js
+++ b/ui/__tests__/components/requirements/policy-link.test.js
@@ -1,0 +1,13 @@
+import { shallow } from 'enzyme';
+import React from 'react';
+
+import PolicyLink from '../../../components/requirements/policy-link';
+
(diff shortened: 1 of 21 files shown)


ui/__tests__/components/requirements/policy-link.test.js
```diff
@@ -0,0 +1,13 @@
+import { shallow } from 'enzyme';
+import React from 'react';
+
+import PolicyLink from '../../../components/requirements/policy-link';
+
+describe('<PolicyLink />', () => {
```
cmc333333: Thanks for adding this test along the way!

Issue #513, opened 2017-10-04
Remove full text search from header on homepage


cmc333333:
As part of #448, we'd like to remove the "two searchbox" problem by removing the full-text search from the header on the homepage (and only the homepage).

PR #517, opened 2017-10-04
Removing the search bar from the home page

rememberlenny:
# What 

Remove the search bar from the home page. Add props to determine when the search bar should appear in the navbar.

# How

Use a prop that is set based on the router. The router pathname determines a value that gets passed to navbar.

# Reference

Resolving issue #513


ui/components/app-wrapper.js
```diff
@@ -22,10 +22,17 @@ Router.onRouteChangeStart = NProgress.start;
 export default function wrapPage(Page, dataFn) {
   function WrappedPage(props) {
     if (props.err) return <ErrorPage err={props.err} />;
-    return <HeaderFooter><Page {...props} /></HeaderFooter>;
+    return <HeaderFooter showSearch={ props.url.pathname !== "/" }><Page {...props} /></HeaderFooter>;
```
cmc333333: Do you think we can get by without testing for global state? It's difficult to test and tends to be fragile.

Maybe we could change the function to `wrapPage(Page, dataFn, headerFooterProps)` and then have `Homepage` pass in `{ showSearch: false }`?

ui/components/navbar.js
```diff
@@ -1,8 +1,15 @@
...
 import SearchContainer from './search/search';
 
 
-function Navbar() {
+export default function Navbar({ showSearch }) {
+  let searchContainer;
+  if (showSearch) {
+    searchContainer = <SearchContainer />;
+  }
+
```
cmc333333: Now that there's some logic in this function, should we add a test?

ui/components/navbar.js
```diff
@@ -18,10 +25,12 @@ function Navbar() {
...
+        { searchContainer }
       </div>
     </div>
   );
 }
 
-export default Navbar;
+Navbar.propTypes = {
+  showSearch: PropTypes.bool
+};
```
cmc333333: Should `showSearch` have a default of `true`?

ui/components/header-footer.js
```diff
@@ -91,5 +91,6 @@ export default function HeaderFooter({ children }) {
 }
 
 HeaderFooter.propTypes = {
+  showSearch: PropTypes.bool,
```
cmc333333: Heads up, we might get a linting error here (once linting is part of CI again) due to not having a default.

ui/components/navbar.js
rememberlenny: Yep! Just added some.

ui/components/app-wrapper.js
rememberlenny: I agree. I like this. Making an update.

ui/components/header-footer.js
rememberlenny: Copy!

ui/components/navbar.js
rememberlenny: @cmc333333 Added the default value!

ui/__tests__/components/navbar.test.js
```diff
@@ -0,0 +1,18 @@
...
+
+describe('<Navbar />', () => {
+  it('without a search bar when showSearch is false', () => {
+    const result = shallow(<Navbar showSearch={ false }/>);
+    expect(result.find(SearchContainer).length).toEqual(0);
+  });
+
+  it('show search bar when showSearch is true', () => {
+    const result = shallow(<Navbar showSearch={ true }/>);
+    expect(result.find(SearchContainer).length).toEqual(1);
```
cmc333333: Super minor, but `expect(...).toHaveLength(1)` will give us slightly better error messages.

ui/components/app-wrapper.js
```diff
@@ -19,13 +19,20 @@ Router.onRouteChangeError = NProgress.done;
 Router.onRouteChangeStart = NProgress.start;
 
 
-export default function wrapPage(Page, dataFn) {
+export default function wrapPage(Page, dataFn, headerFooterParams = { showSearch: true }) {
```
cmc333333: Unfortunately, default values are shared across all calls to `wrapPage`. Combine this with a mutable value and we'll get very unexpected results.

We can get around that with something like:
```js
export default function wrapFunction(Page, dataFn, headerFooterParams) {
  const hfParams = headerFooterParams || { showSearch: true };
```

ui/components/app-wrapper.js
```diff
@@ -19,13 +19,20 @@ Router.onRouteChangeError = NProgress.done;
...
   function WrappedPage(props) {
     if (props.err) return <ErrorPage err={props.err} />;
-    return <HeaderFooter><Page {...props} /></HeaderFooter>;
+    return <HeaderFooter showSearch={ headerFooterParams.showSearch }><Page {...props} /></HeaderFooter>;
```
cmc333333: Would it make sense to just pass any parameter the user requested? This way we won't have to modify this function should we add more parameters. Something like:
```js
return <HeaderFooter {...headerFooterParams}><Page {...props} /></HeaderFooter>;
```

rememberlenny: I had to look up how the spread syntax worked. This is brilliant.

rememberlenny:
@cmc333333 Updates are up.

cmc333333:
Thanks for accommodating my nits, @rememberlenny !

Issue #505, opened 2017-10-03
Ajaxify links


cmc333333:
As an agency user, I don't want to wait for full page refreshes.

## AC
* Find all links that aren't AJAXy (i.e. aren't wrapped in `<Link>`) and make then AJAXy them.

PR #521, opened 2017-10-05
Change anchor tag links to use Link component

rememberlenny:
# What

Ajaxify links

# Reference

https://github.com/18F/omb-eregs/issues/505

### DIFF
--- a/ui/components/errors/four-oh-four-error.js
+++ b/ui/components/errors/four-oh-four-error.js
@@ -1,5 +1,6 @@
 import React from 'react';
 
+import { Link } from '../../routes';
 import HeaderFooter from '../header-footer';
 
 export default function FourOhFour() {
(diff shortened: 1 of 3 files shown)


ui/components/errors/unexpected-error.js
```diff
@@ -1,13 +1,14 @@
...
 import React from 'react';
+import { Link } from '../../routes';
 
 export default function UnexpectedError({ statusCode }) {
   return (
     <section>
       <h1>{HTTPStatus[statusCode] || 'Unexpected error'}</h1>
       <p>
-        Visit our <a href="/">homepage</a> or <a href="mailto:ofcio@omb.eop.gov">contact us</a> if
+        Visit our <Link route="homepage"><a>homepage</a></Link> or <a href="mailto:ofcio@omb.eop.gov">contact us</a> if
```
cmc333333: Question to ponder: this error message will appear if something unexpected went wrong (an API call failed, some exception was thrown, etc.), but the app is still running (so Next/React are working okay). Do you think it'd be safer to avoid the AJAXy links here?

rememberlenny: @cmc333333 I agree. Ill make an update.

PR #523, opened 2017-10-05
Add endpoint for serializing trees

cmc333333:
This endpoint allows us to grab either the root of a document or some subsection. It doesn't make any attempt at paging just yet, but should be enough to help us prototype viewing documents.

### DIFF
--- a/api/document/urls.py
+++ b/api/document/urls.py
@@ -1,0 +1,8 @@
+from django.conf.urls import url
+
+from document.views import TreeView
+
+urlpatterns = [
+    url(r'^(?P<policy_id>\d+)(/(?P<identifier>[a-zA-Z0-9_-]+))?',
+        TreeView.as_view()),
+]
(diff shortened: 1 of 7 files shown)


api/document/urls.py
```diff
@@ -0,0 +1,8 @@
+from django.conf.urls import url
+
+from document.views import TreeView
+
+urlpatterns = [
+    url(r'^(?P<policy_id>\d+)(/(?P<identifier>[a-zA-Z0-9_-]+))?',
```
tadhg-ohiggins: We might want to use `[0-9]` instead of `\d` here, as in Python 3 `\d` matches anything flagged in Unicode as a numeral digit, such as `൬`. Not a big deal, but if we're expecting policy numbers to only be 0-9 we might as well go with the more restrictive option.

cmc333333: Good to know and can do. I wonder how well Django/psycopg/etc. handle that...

cmc333333:
Thanks @tadhg-ohiggins !

PR #524, opened 2017-10-06
Remove squashed migrations, taggit

### LINKED ISSUE #459: Remove taggit dependency
We removed taggit from our executing code in #398, but it's still referenced in our migrations. Squash those migrations and remove the dependency.

cmc333333:
To close #459, this removes all of the migrations which were squashed in #471 and removes taggit from our dependencies.


cmc333333:
Thanks @tadhg-ohiggins. I'm going to merge post demo, just in case.

PR #536, opened 2017-10-10
Fix route typo.

cmc333333:
This is a short-term fix for a typo'd route name. For a longer-term solution,
we should use our prop-type checking to validate the route.


cmc333333:
Thanks @rememberlenny

PR #537, opened 2017-10-10
Hotfix #536

cmc333333:


jeremiak:
This looks good to me, I can see that the search results paging bug is fixed

:shipit:

PR #540, opened 2017-10-10
Pull API related code into an API util

jeremiak:
This is part of the larger effort to have only config files in the root of `/ui` and a single entry `.js` file that makes it a bit easier to know where to start for reading the code.

There are lots of things named variants of "lookup" and "search", so I also renamed the default component in `search-redirect.js` to `<SearchRedirect />` to make it more standard and clear.


jeremiak:
@cmc333333 Yup, I can do that. Sorry, I have `prettier` set up on my editor so it does a lot of formatting on save. I'll turn that off and keep the two separate, no problem!

PR #545, opened 2017-10-10
Very rough initial policy detail page

cmc333333:
It lists requirements! They can be clicked!

This gives us a structure to build off more than a full implementation.

Looks like
policy-detail


cmc333333:
Thanks @rememberlenny ! @jeremiak does anything stick out wrt the direction we're trying to take around code layout?

rememberlenny:
Ill leave this up to @jeremiak

ui/components/policy/page-counter.js
```diff
@@ -0,0 +1,18 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+
+export default function PageCounter({ count, page, pageSize }) {
```
jeremiak: This component looks good and tested, but it doesn't look like its policy specific so it should be in `components/`. I think the benefit of this is that it is easy to find all the components in one spot, leading to increased chance of reuse and refactor.

ui/pages/policy.js
```diff
@@ -0,0 +1,97 @@
...
+
+export class Policy extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      focusReq: props.url.query.reqId,
+    };
+  }
+
+  get reqs() {
```
jeremiak: I don't think we use the class getter/setter methods anywhere else in the code base so for consistency I would change this to just be `reqs`.

FWIW, I think using the getter/setters might add some readability to the code base but it should be a consistent pattern across the component if not also the codebase

jeremiak:
@cmc333333 ya, I think the general direction is fine. One thing I'd suggest is to put as many components in a single directory as you feel comfortable doing. At some point, this list gets unwieldy or requires long file, unique file names, but up until that point it makes it easier to know which components are available for re-use/refactoring

cmc333333:
Thanks @jeremiak , will do.

PR #547, opened 2017-10-11
Link component

jeremiak:
Tests are not yet in a state for merging as I have to better mock out the router to decouple our tests from the routes in the codebase.

Other than that, let me know what you think of this new `<Link />` component @cmc333333 and @rememberlenny (I've asked for a review but feel free to not approve/comment in the review until the tests are ready). Basically, its a one link component to rule them all type of deal.

It takes a few different props:
* `className` (string)
* `href` (string)
* `params` (object)
* `route` (string)

You use this component just like `<a>` or the link component from `next-routes`, ie `<Link href='/fake'>Fake link, click me!</Link>`. The children are passed right through.

The `className` functions just as you'd expect and passes through the classes to the underlying/resulting `<a>` element.

You can use either `href` or `params` and `route`. If you use `href`, then it works just like a regular link tag except it assumes that the link is an "external link" if it starts with `https`, `http` or `//` since those are likely going to point to other hosts/URLs.

If you use `route`, then you can just pass in the named route. The component will verify that the route exists on the router and then wrap the `<Link />` exported by `next-routes`/`routes.js`.

If any `aria-*` props are supplied they will be passed onto the underlying `<a>` element.

This component is a bit complicated, but it accounts for our three link use cases and DRYs up the importing of the routes from the router. This way, we can change the way that file is structured and only have to update the single shared `components/link.js`.


ui/components/homepage/new-policy-view.js
```diff
@@ -1,6 +1,6 @@
 import PropTypes from 'prop-types';
 import React from 'react';
-import { Link } from '../../routes';
+import Link from '../link';
```
cmc333333: Will we need to remove the `<a>` from all of the `Link`s?

ui/components/link.js
```diff
@@ -0,0 +1,50 @@
...
+    } : {};
+    return (<a className={className} href={href} {...attrs} >{children}</a>);
+  }
+}
+
+/* eslint-disable consistent-return, react/forbid-prop-types */
+Link.propTypes = {
+  children: PropTypes.node.isRequired,
+  className: PropTypes.string,
+  href: PropTypes.oneOfType([PropTypes.bool, PropTypes.string]),
```
cmc333333: When would this be a boolean?

ui/components/link.js
```diff
@@ -0,0 +1,50 @@
...
+  params: PropTypes.object,
+  route: (props, propName) => {
```
cmc333333: Curious: can we simplify to 
```js
route: PropTypes.oneOf(existingRoutes)
```
? A lot less custom logic if so.

ui/components/link.js
jeremiak: It won't anymore, but I had the default prop as `false` to make the conditional easy, but that conditional will still work with the new default `null`. I'll update this.

ui/components/homepage/new-policy-view.js
jeremiak: Yup, pushing an update with this

cmc333333:
I've rebased this from master and extended it a bit to use `Link` consistently throughout the app; it's ready for re-review.

ui/components/link.js
```diff
@@ -3,52 +3,36 @@ import React from 'react';
...
-      ...accum,
-      [next.name]: next.value,
-    }), {});
-
-  return otherProps;
-}
-
-export default function Link({ children, className, href, params, route, ...props }) {
-  const ariaProps = getAriaProps(props);
+export default function Link({ children, href, params, route, ...otherProps }) {
```
rememberlenny: 👍

cmc333333:
Thanks @rememberlenny

PR #560, opened 2017-10-13
Connection and serialization between Req and Doc

### LINKED ISSUE #544: Determine document structure/metadata format
Moving to a policy-centric approach means we need:

- [ ] A document structure for policies.
- [ ] A metadata structure for policies and requirements.
- [ ] An outline of how to relate policies and requirements.

These are all on the development side; this issue doesn't address the design side.

cmc333333:
In moving to a document-centric model, we still want to associate requirement data with specific paragraphs of the document. Based on work built in a pairing session with @tadhg-ohiggins, this adds a connection between the two and serializes it in the document endpoint.

This should resolve #544


rememberlenny:
Looks good!

cmc333333:
Thanks both! I'll merge in master rather than rebasing to keep from messing up any of Tadhg's work.

Issue #549, opened 2017-10-11
Policy detail: Better Metadata


cmc333333:
On the Policy detail page, we're currently displaying the requirement id, and an ordered list of topics. Let's clean that up based on this [early prototype](https://gsa.invisionapp.com/share/QUDR1NBCK#/screens/256032982):

* Relabel the Requirement id field
* Add the "Applies to" field ("impacted entity" in the data)
* Turn topics into links (to a list of policies with that filter)

Don't worry about sunset date at the moment -- that applies to the policy as a whole, so we likely won't want it on a specific requirement.

rememberlenny:
@cmc333333 On it!

rememberlenny:
<img width="1044" alt="screen shot 2017-10-11 at 5 58 13 pm" src="https://user-images.githubusercontent.com/1332366/31468909-cee8789e-aead-11e7-8c8e-f364999a5fd0.png">

PR #561, opened 2017-10-13
Add the metadata on the policy page

rememberlenny:
# Reference 

https://github.com/18F/omb-eregs/issues/549

# What

Move the `MetaData` components found in the requirements page over into the `Policy` page.

screen shot 2017-10-13 at 3 05 15 pm

### DIFF
--- a/ui/__tests__/components/policy/req.test.js
+++ b/ui/__tests__/components/policy/req.test.js
@@ -1,4 +1,4 @@
-import { shallow, mount } from 'enzyme';
+import { shallow } from 'enzyme';
 import React from 'react';
 
 import Req from '../../../components/policy/req';
@@ -12,24 +12,24 @@ describe('<Req />', () => {
   };
 
   it('includes meta data when highlighted', () => {
-    const normalText = mount(
+    const normalText = shallow(
       <Req href="" onClick={jest.fn()} req={req} />,
-    ).text();
-    const hlText = mount(
+    ).html();
+    const hlText = shallow(
       <Req highlighted href="" onClick={jest.fn()} req={req} />,
-    ).text();
+    ).html();
 
     expect(normalText).toMatch(/Some text here/);
-    expect(normalText).not.toMatch(/123.45/);
+    expect(normalText).not.toMatch(/>123.45</);
     expect(hlText).toMatch(/Some text here/);
-    expect(hlText).toMatch(/123.45/);
+    expect(hlText).toMatch(/Requirement ID: 123.45/);
   });
 
   it('includes topiclinks with data when highlighted', () => {
     req.topics = [{ id: 1, name: 'link 1' }, { id: 2, name: 'link 2' }];
-    const normalText = mount(
+    const normalText = shallow(
       <Req highlighted href="" onClick={jest.fn()} req={req} />,
-    ).find('.topics').text();
+    ).find('.topics').html();
 
     expect(normalText).toMatch(/link 1/);
     expect(normalText).toMatch(/link 2/);
--- a/ui/__tests__/components/policy/req.test.js
+++ b/ui/__tests__/components/policy/req.test.js
@@ -25,6 +25,16 @@ describe('<Req />', () => {
     expect(hlText).toMatch(/123.45/);
   });
 
+  it('includes topiclinks with data when highlighted', () => {
+    req.topics = [{id: 1, name: 'link 1'}, {id: 2, name: 'link 2'}];
+    const normalText = mount(
+      <Req highlighted href="" onClick={jest.fn()} req={req} />,
+    ).find('.topics').text();
+
+    expect(normalText).toMatch(/link 1/);
+    expect(normalText).toMatch(/link 2/);
+  });
+
   it('includes an anchor with the correct data', () => {
     const onClick = jest.fn();
     req.req_id = '1.1';
--- a/ui/__tests__/components/policy/req.test.js
+++ b/ui/__tests__/components/policy/req.test.js
@@ -2,7 +2,6 @@ import { shallow, mount } from 'enzyme';
 import React from 'react';
 
 import Req from '../../../components/policy/req';
-import Metadata from '../../../components/requirements/metadata';
 
 describe('<Req />', () => {
   const req = {
@@ -27,11 +26,8 @@ describe('<Req />', () => {
 
   it('includes an anchor with the correct data', () => {
     const onClick = jest.fn();
-    const req = {
-      req_id: '1.1',
-      req_text: 'Text goes here',
-      topics: [],
-    };
+    req.req_id = '1.1';
+    req.req_text = 'Text goes here';
     const anchors = shallow(
       <Req href="/some/location" onClick={onClick} req={req} />,
     ).find('a');
--- a/ui/__tests__/components/policy/req.test.js
+++ b/ui/__tests__/components/policy/req.test.js
@@ -26,9 +26,9 @@ describe('<Req />', () => {
   });
 
   it('includes topiclinks with data when highlighted', () => {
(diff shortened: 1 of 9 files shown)


ui/components/policy/req.js
```diff
@@ -1,15 +1,64 @@
...
+  'Cannot determine-Ask Mindy',
+].map(e => e.toLowerCase());
+
+export function filterAppliesTo(text) {
+  const normalized = (text || '').toLowerCase().trim();
+  if (badEntities.includes(normalized)) {
+    return null;
+  }
+  return text;
+}
```
cmc333333: What do you think about importing `filterAppliesTo` rather than copy-pasting it? We'll probably move logic around in the future, but it'd be better if we knew the function hadn't changed.

ui/components/policy/req.js
```diff
@@ -1,15 +1,64 @@
...
-        <ul>
-          { req.topics.map(t => <li key={t.id}>{t.name}</li>) }
-        </ul>
+      <div className="req col col-4">
+        <Metadata className="requirement-id" name="Requirement ID" value={req.req_id} />
+        <Metadata
+          className="omb-policy-id"
+          name="OMB Policy ID"
+          value={req.policy.omb_policy_id}
+        />
```
cmc333333: The "policy" fields won't be too useful here, as they apply to the whole document rather than an individual requirement. Would you mind removing them?

ui/components/policy/req.js
```diff
@@ -1,15 +1,64 @@
...
+          className="sunset-date"
+          name="Sunset date"
+          value={req.policy.sunset}
+          nullValue="Sunset date: none"
+          separator=" by "
+        />
+        <div className="topics metadata">
+          <span>Topics: </span>
+          <ul className="topics-list list-reset inline">
+            {req.topics.map(topic => <TopicLink key={topic.id} topic={topic} />)}
```
cmc333333: We should probably link to the `policies` list page rather than `requirements`, yes? It might make sense just to modify `TopicLink`'s implementation for now.

ui/components/policy/req.js
```diff
@@ -1,15 +1,64 @@
...
 export default function Req({ highlighted, href, onClick, req }) {
   let meta = null;
   if (highlighted) {
     meta = (
-      <div className="col col-4">
-        Number: { req.req_id }
-        <ul>
-          { req.topics.map(t => <li key={t.id}>{t.name}</li>) }
-        </ul>
+      <div className="req col col-4">
```
cmc333333: Mind adding a quick test that some of these fields are visible?

rememberlenny:
Yep! Not done yet. Want to transfer over the old requirement tests.

--
Leonard Bogdonoff
rememberlenny.com | @rememberlenny

ui/components/policy/req.js
rememberlenny: Love this idea.

rememberlenny:
@cmc333333 Should be good to go now

ui/__tests__/components/policy/req.test.js
```diff
@@ -1,19 +1,21 @@
...
-      req_id: '123.45',
-      req_text: 'Some text here',
-      topics: [],
-    };
-    const normalText = shallow(
+    const normalText = mount(
       <Req href="" onClick={jest.fn()} req={req} />,
     ).text();
-    const hlText = shallow(
+    const hlText = mount(
```
cmc333333: Also small: `mount` renders a component and all of its children where `shallow` will only render the component; that makes `shallow` a bit better for pure unit tests (as we aren't depending on the work of contained components). Where possible, we should prefer `shallow` for that reason.

ui/__tests__/components/policy/req.test.js
```diff
@@ -23,13 +25,20 @@ describe('<Req />', () => {
...
+
   it('includes an anchor with the correct data', () => {
     const onClick = jest.fn();
-    const req = {
-      req_id: '1.1',
-      req_text: 'Text goes here',
-      topics: [],
-    };
+    req.req_id = '1.1';
+    req.req_text = 'Text goes here';
```
cmc333333: What do you think about extending the `req` rather than modifying it?
```js
const modifiedReq = { ...req, req_id: '1.1', req_text: 'Text goes here' };
```
If we modify it in one test, the modified version will be shared by other tests, which makes the order of test execution matter.

Alternatively, we could use a `beforeEach`:
```js
describe('<Req />', () => {
    let req;
    beforeEach(() => {
        req = {
            req_id: '123.45',
            req_text: 'Some text here',
            topics: [],
            policy: {},
        };
    });

    it( ...
});
```

ui/__tests__/components/requirements/topic-link.test.js
```diff
@@ -0,0 +1,31 @@
...
+describe('<TopicLink />', () => {
+  let topic = { id: 1, name: 'hello link 1' };
+  const basic = shallow(
+    <TopicLink topic={topic} />,
+  );
+  it('renders', () => {
+    expect(basic.type()).toBe('li');
+  });
+  it('has the right route', () => {
+    expect(basic.find('LinkRoutes').props().route).toMatch('requirements');
```
cmc333333: Small detail: we can use `.prop('route')` as a shorthand.

ui/components/policy/req.js
```diff
@@ -1,15 +1,40 @@
...
+        <Metadata className="requirement-id" name="Requirement ID" value={req.req_id} />
+        <Metadata
+          className="issuance"
+          name="Policy issuance"
+          value={req.policy.issuance}
+        />
```
cmc333333: We should probably remove the `policy` fields as they apply to the whole document rather than individual requirements.

ui/components/policy/req.js
```diff
@@ -1,15 +1,40 @@
...
+          value={filterAppliesTo(req.impacted_entity)}
+        />
+        <Metadata className="issuing-body" name="Issuing body" value={req.issuing_body} />
+        <Metadata
+          className="sunset-date"
+          name="Sunset date"
+          value={req.policy.sunset}
+          nullValue="Sunset date: none"
+          separator=" by "
+        />
```
cmc333333: Same comment, re: policy fields.

ui/components/requirements/policy-link.js
```diff
@@ -3,13 +3,19 @@ import React from 'react';
 
 import { Link } from '../../routes';
 
+export function policyLinkTag(policy) {
+  return (
+    <Link route="policies" params={{ id__in: policy.id }}>
+      <a>{policy.title_with_number}</a>
+    </Link>
+  );
+}
```
cmc333333: Should this be a component? Components can be functions, but offer a lot of extra React functionality.

ui/components/requirements/requirement.js
```diff
@@ -57,13 +58,6 @@ export default function Requirement({ requirement }) {
             value={filterAppliesTo(requirement.impacted_entity)}
           />
           <Metadata className="issuing-body" name="Issuing body" value={requirement.issuing_body} />
-          <Metadata
-            className="sunset-date"
-            name="Sunset date"
-            value={requirement.policy.sunset}
-            nullValue="Sunset date: none"
-            separator=" by "
-          />
```
cmc333333: Did you mean to remove this from the `policy/req.js` component?

ui/__tests__/components/policy/req.test.js
rememberlenny: Updated

ui/components/requirements/policy-link.js
rememberlenny: @cmc333333 This was done to quiet codeclimate. I think if we want, we could turn this into a component, but I would wait to see where this goes.

PR #562, opened 2017-10-16
Account for bad policy id

### LINKED ISSUE #557: Policy detail: Account for bad urls
On the policy detail page, if a user types a policy id which doesn't exist in our database, they should receive a 404 page. They currently get a 500.

cmc333333:
If requesting a policy detail page for a policy which doesn't exist, we should return a 404 rather than a 500.

Resolves #557


cmc333333:
Thanks @rememberlenny

PR #563, opened 2017-10-16
Rework Search component to be route-based.

### LINKED ISSUE #558: Policy detail: Search adds extra params
The policy detail page is our first to use parameters from the url. Next.js mixes these in with the query parameters, which unfortunately means that they are included when we include those params in the full text search in the header.

For example, if we are on the policy detail page and we search for something in the upper-right search box, we'll get an extra `policyId` param (and possibly, `reqId`) in the URL.

To fix this, we probably want the search box to only include url parameters from the "pretty" url, rather than the Next.js page-based url.

cmc333333:
The Search component was written prior to our Next.js-based routes, which
meant that it was doing all of its logic via substring searches. This
refactors it to instead define a constant for the default route and a mapping
between routes and their parameter names.

In the process, it alters the logic to only keep the query params when the
user is on one of the two whitelisted pages (requirements or policies); query
parameters on other pages aren't relevant.

Should resolve #558

### DIFF
--- a/ui/__tests__/components/search/search.test.js
+++ b/ui/__tests__/components/search/search.test.js
@@ -8,17 +8,17 @@ const blankRouter = { pathname: '', query: {} };
 
 describe('<Search />', () => {
   describe('actionPath()', () => {
-    it('returns /policies/ by default', () => {
+    it('returns /policies by default', () => {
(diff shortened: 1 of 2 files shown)


cmc333333:
Thanks @rememberlenny

PR #567, opened 2017-10-20
Remove the navigation tabs from policies results.

cmc333333:
Per feedback from @carodew, the partial navigation tabs on the policies
results page were confusing and no longer provided use. This removes them by
moving their markup into the requirements listing page.

It also removes the RequirementsTab component as that is no longer in use.


cmc333333:
Thanks @tadhg-ohiggins, I'll rebase then merge.

PR #571, opened 2017-10-23
Script to import a document from XML

cmc333333:
While we begin thinking through what an import/authoring tool would look like,
we have pressing needs to display whole documents. This command acts as a
stop-gap, giving an interface for entering these documents (as XML).

This also includes some XML to represent M-16-19. It doesn't account for inline elements (like footnote citations) at all, but should make generating documents a bit easier.


cmc333333:
Thanks @tadhg-ohiggins

PR #573, opened 2017-10-25
Initial document-viewing page

cmc333333:
Includes renderers for:
* paragraphs
* headings
* sections

along with some minimal, inline styles to distinguish them. The goal in pushing this up is that we'll have a common base to start dividing up work.

## Looks like
document-detail


cmc333333:
Thanks @rememberlenny. I'll rebase and merge.

PR #577, opened 2017-10-26
Fix import error.

cmc333333:
We moved the redirectQuery function around during our last sprint; looks like
we missed a reference.


cmc333333:
Thanks @rememberlenny

Issue #551, opened 2017-10-11
Policy detail: Policy number in url


cmc333333:
The policy detail page is currently accessible at `/policy/{policy_id}`. Many (but not all) polices have a more meaningful policy number (like "A-130" or "M-16-19") which would be better to key off. Let's ignore the corner case around numbers like "A-136 Revised" and check for `[AM]-[0-9-]+` in the canonical URL. This is currently stored in the `omb_policy_id` field on `Policy`.

krues8dr:
We have policies that do not have numbers, unfortunately. E.g. [this](https://obamawhitehouse.archives.gov/sites/default/files/omb/assets/egov_docs/release_memo_-_planning_guide_roadmap_toward_ipv6_adoption_052009.pdf). I'd love to see us move to using a slug here instead of IDs, universally, which defaults to the policy number or a slugified title in absence of that.

cmc333333:
We'll aim to get to both this sprint, but I've split of the slug-related things into #575. For this ticket, we just want them accessible by their policy number (e.g. M-16-19) when present. We'll continue to allow access via policy id.

cmc333333:
Looks like we'll need to tweak the UI as well -- maybe just the routers file?

* https://omb-eregs-demo.app.cloud.gov/policy/M-16-19 gives a 404 but
* https://omb-eregs-api-demo.app.cloud.gov/policies/M-16-19/ works

rememberlenny:
On it!

rememberlenny:
This should be finished up with #598

PR #578, opened 2017-10-26
Add policy search through omb_policy_id

rememberlenny:
# Reference

https://github.com/18F/omb-eregs/issues/551

### DIFF
--- a/api/reqs/tests/views_policies_tests.py
+++ b/api/reqs/tests/views_policies_tests.py
@@ -156,3 +156,27 @@ def test_nonpublic_reqs():
     assert policy.requirements.count() == 5
     assert response['results'][0]['relevant_reqs'] == 4
     assert response['results'][0]['total_reqs'] == 4
+
+
+@pytest.mark.django_db
+def test_omb_policy_id():
+    client = APIClient()
+    omb_policy_id = "M-123-4"
+    path = "/policies/{0}".format(omb_policy_id)
+    response = client.get(path)
+    assert response.status_code == 301
+    mommy.make(Policy, omb_policy_id=omb_policy_id)
+    response = client.get(path + '.json').json()
+    assert response['omb_policy_id'] == omb_policy_id
+
+
+@pytest.mark.django_db
+def test_pk_id():
+    client = APIClient()
+    pk_id = 123
+    path = "/policies/{0}".format(pk_id)
+    response = client.get(path)
+    assert response.status_code == 301
+    mommy.make(Policy, pk=pk_id)
+    response = client.get(path + '.json').json()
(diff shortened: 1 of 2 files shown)


api/reqs/tests/views_policies_tests.py
```diff
@@ -156,3 +156,31 @@ def test_nonpublic_reqs():
...
+@pytest.mark.django_db
+def test_omb_policy_id():
+    client = APIClient()
+    omb_policy_id = 'M-123-4'
+    path = "/policies/{0}.json".format(
+        omb_policy_id)
+    response = client.get(path).json()
+    assert response['detail'] == "Not found."
+    policy = mommy.make(Policy, omb_policy_id=omb_policy_id)
+    mommy.make(Requirement, policy=policy, public=False)
```
cmc333333: Let's remove this line -- it's not relevant to the test, so may be misleading.

api/reqs/tests/views_policies_tests.py
```diff
@@ -156,3 +156,31 @@ def test_nonpublic_reqs():
...
+@pytest.mark.django_db
+def test_pk_id():
+    client = APIClient()
+    pk_id = 123
+    path = "/policies/{0}.json".format(
+        pk_id)
+    response = client.get(path).json()
+    assert response['detail'] == "Not found."
+    policy = mommy.make(Policy, pk=pk_id)
+    mommy.make(Requirement, policy=policy, public=False)
```
cmc333333: Similar here.

api/reqs/tests/views_policies_tests.py
```diff
@@ -156,3 +156,31 @@ def test_nonpublic_reqs():
...
     assert response['results'][0]['relevant_reqs'] == 4
     assert response['results'][0]['total_reqs'] == 4
+
+
+@pytest.mark.django_db
+def test_omb_policy_id():
+    client = APIClient()
+    omb_policy_id = 'M-123-4'
+    path = "/policies/{0}.json".format(
+        omb_policy_id)
```
cmc333333: Small detail, but f-strings are slightly more pythonic:
```python
path = f"/policies/{omb_policy_id}"
```

api/reqs/tests/views_policies_tests.py
```diff
@@ -156,3 +156,31 @@ def test_nonpublic_reqs():
...
+    response = client.get(path).json()
+    assert response['detail'] == "Not found."
```
cmc333333: What do you think about testing the response (prior to `.json()`)'s status code instead? Testing for a 404 is probably a bit more meaningful as we don't check `detail` in the frontend.

PR #581, opened 2017-10-26
Fix import error. (Hotfix)

cmc333333:
We moved the redirectQuery function around during our last sprint; looks like
we missed a reference.

---

This cherry-picks the one commit from #577 into the release branch.


cmc333333:
Thanks @tadhg-ohiggins !

PR #583, opened 2017-10-26
Add "content" field, including footnote citations

cmc333333:
This is easiest to review as two changesets.

The first lays out an eRegs like model to represent inline content, here dubbed `Annotation`s, which refer to a specific interval of text within a `DocNode`. It adds one specific type of annotation, a footnote citation (i.e. the text that links _to_ a footnote, not the footnote itself) to show how this might work.

The second breaks from the eRegs model a bit by moving the logic which combines the annotations into the API. Here, we're adding a `content` field, which is currently a list of JSON elements representing either a chunk of normal text or a chunk of footnote text. It should be pretty straight forward to convert those chunks of content into the appropriate templates. The approach here makes some naive decisions (using `RangeMap`s to handle overlap, not allowing nesting, predefining which annotations we'll need, etc.), but should be enough to build on.


cmc333333:
Thanks @tadhg-ohiggins!

Issue #535, opened 2017-10-10
Can we change the feedback email at the top to the policy library email address? PolicyLibrary@omb.eop.gov


carodew:

tadhg-ohiggins:
The fix has been merged into master.

PR #584, opened 2017-10-27
Change email address

tadhg-ohiggins:
Sometimes these things change
And the next time it happens
We will be ready.

Create a ContactEmail element, move all existing mailto: references over to it, and add tests for it and a couple of elements that were missing tests.

Address #535

### DIFF
--- a/ui/__tests__/components/contact-email.test.js
+++ b/ui/__tests__/components/contact-email.test.js
@@ -1,0 +1,30 @@
+import { mount } from 'enzyme';
+import React from 'react';
+
+import { email, ContactEmail } from '../../components/contact-email';
+
+const expr = `a[href="mailto:${email}"]`;
+
+describe('<ContactEmail />', () => {
+  describe('when given no props', () => {
+    const mailtext = ContactEmail.defaultProps.text;
+    const result = mount(<ContactEmail />);
+    it('Has the expected href value', () => {
+      expect(result.find(expr).length).toEqual(1);
+    });
+    it('Has the expected text value', () => {
+      expect(result.find(expr).text()).toEqual(mailtext);
+    });
+  });
+  describe('when given a text prop', () => {
+    const result = mount(<ContactEmail text='something'/>);
+    it('Has the expected href value', () => {
+      expect(result.find(expr).length).toEqual(1);
+    });
+    it('Has the expected text value', () => {
+      expect(result.find(expr).text()).toEqual('something');
+    });
+  });
+
+});
+
--- a/ui/__tests__/components/contact-email.test.js
+++ b/ui/__tests__/components/contact-email.test.js
@@ -17,7 +17,7 @@ describe('<ContactEmail />', () => {
     });
   });
   describe('when given a text prop', () => {
-    const result = mount(<ContactEmail text='something'/>);
+    const result = mount(<ContactEmail text="something" />);
     it('Has the expected href value', () => {
       expect(result.find(expr).length).toEqual(1);
     });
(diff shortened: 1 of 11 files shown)


ui/__tests__/components/contact-email.test.js
```diff
@@ -0,0 +1,30 @@
+import { mount } from 'enzyme';
+import React from 'react';
+
+import { email, ContactEmail } from '../../components/contact-email';
+
+const expr = `a[href="mailto:${email}"]`;
+
+describe('<ContactEmail />', () => {
+  describe('when given no props', () => {
+    const mailtext = ContactEmail.defaultProps.text;
```
cmc333333: Mind adding a quick assert that this is non-empty? Right now we're asserting against a variable, which could be unintentionally empty and throw off the test.

ui/__tests__/components/disclaimer.test.js
```diff
@@ -0,0 +1,23 @@
+import { mount } from 'enzyme';
+import React from 'react';
+
+import Disclaimer from '../../components/disclaimer';
+import { email } from '../../components/contact-email';
+
+describe('<Disclaimer />', () => {
+  describe('returns the right contents', () => {
+    it('has the right number of div elements', () => {
```
cmc333333: I think tests that are tied too closely to the markup are a bit fragile (since they're easy to break). What do you think about testing for some of the content within divs instead? Something like:
```js
expect(result.text()).toMatch(/some string/);
```

ui/components/contact-email.js
```diff
@@ -3,7 +3,7 @@ import React from 'react';
 
 import Link from './link';
 
-export const email = 'ofcio@omb.eop.gov';
+export const email = 'PolicyLibrary@omb.eop.gov';
```
cmc333333: Love how simple this change is.

ui/components/contact-email.js
```diff
@@ -0,0 +1,21 @@
...
+import React from 'react';
+
+import Link from './link';
+
+export const email = 'PolicyLibrary@omb.eop.gov';
+const mailto = `mailto:${email}`;
+const mailtext = `Email ${email}`;
+
+/* Has the same href value every time, making the global email address easy to change. */
+export function ContactEmail({ text }) {
```
cmc333333: Something to consider in terms of interface design -- do we want to use React children instead of a prop? If we did, we could call:
```jsx
<ContactEmail>contact us</ContactEmail>
```
It'd also let us use non-strings:
```jsx
<ContactEmail><img src="mailbox.png" /></ContactEmail>
```
Getting the default value would be a bit more complicated -- I'm not sure we could use `defaultProps`.

ui/__tests__/components/contact-email.test.js
tadhg-ohiggins: 👍

ui/__tests__/components/disclaimer.test.js
```diff
@@ -0,0 +1,18 @@
+import { mount } from 'enzyme';
```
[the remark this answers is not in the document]
tadhg-ohiggins: I wanted it to render the nested ContactEmail element inside of Disclaimer here, so went with `mount`.

ui/__tests__/components/disclaimer.test.js
tadhg-ohiggins: Good point.

ui/components/contact-email.js
tadhg-ohiggins: I went with the easy option here, because I wasn't sure how to handle the default value, but this seems like a good enhancement. I might leave it until we hit a point where we know we want something that's not flat text, though.

PR #586, opened 2017-10-27
Refactor DocCursor out of models.

cmc333333:
The DocNode model was getting a bit heavy with logic, all of which could be
shifted to the DocCursor (/tree). This moves that logic and the DocCursor out
to lighten the models file.


cmc333333:
Thanks @tadhg-ohiggins and @rememberlenny

PR #594, opened 2017-10-30
Fix value of API_URL in docker-compose.yml to work on Windows.

toolness:
This is a follow-up to #592 that fixes a problem on Windows where 0.0.0.0 isn't a valid IP address on Windows browsers (Edge, Chrome, and Firefox all report it as an invalid address).  So this changes `API_URL` to use `localhost` instead, so the front-end's XHR calls on Windows browsers work.


toolness:
Umm sure, like this?

PR #595, opened 2017-10-30
Modify import script to allow footnote citations

cmc333333:
This reworks the XML import script to operate over `<content/>` tags, which may contain inline elements (starting with `footnote_citation`s). It'll be easiest to review changeset by changeset (re-organizing code is separate from new logic).

This doesn't handle nested, inline tags at the moment, but should be enough to get us started.


cmc333333:
Thanks, @tadhg-ohiggins!

PR #596, opened 2017-10-31
Display inline content in the UI

cmc333333:
Now that the API is returning an array of `content` elements to represent inline text, we need to render it. This begins that process with renders for plain text (which become spans) and footnote citations (which become a super-script anchor).


cmc333333:
Thanks @tadhg-ohiggins !

Issue #575, opened 2017-10-26
Add slug to policies; use it in urls


cmc333333:
Building off @krues8dr's suggestion in #551:

* Add a slug field to policy models
* Back-port values for this field (based on the policy title)
* Allow policies to be fetched using the slug

cmc333333:
#599 implemented this

PR #599, opened 2017-10-31
Add slug to policies

toolness:
This is an initial attempt at fixing #575 by adding a `slug` field
to the `Policy` model and making it accessible via URLs.

Notes:

* It seems the `/policies/:foo` route first checks to see if
  `foo` is an OMB Policy ID and then checks to see if it's a primary key. I
  added an additional fallback to check if it's a slug; I'm not sure if
  this prioritization is right, or if this is even what was intended by
  the requirement "Allow policies to be fetched by the slug" in #575's
  description.

* The `Policy.save()` method has been overridden to set the `slug`
  from the `title` if it's currently empty.  I'm not 100% sure if this
  is needed, since we're also pre-populating the `slug` field from the
  `title` in the admin.

* Currently the `slug` doesn't seem to be exposed by the API, e.g.
  when fetching the JSON serialization of a policy. This seems like it
  could make it difficult for the front-end to actually use the slug.
  However, since it wasn't listed as an explicit requirement in
  #575, I didn't add the functionality. (Also, since I'm brand-new to
  this project and this is my first PR, I lack the judgement to decide
  whether it's actually needed or not.)

* The data migration that sets the slug automatically from the title
  has a dependency on django's `slugify` function, so it
  assumes that won't change in the future.

To do:

* I'd really like to add a test for the data migration if possible. 

* I'm not currently sure what the behavior is if two different policies
  have the same slug. Should this not be allowed?


tadhg-ohiggins:
I guess we should decide explicitly what the priority order for checking `foo` against should be? Do we want slug to be last, as it is here?

tadhg-ohiggins:
FWIW I think we should return slug in the JSON, too.

tadhg-ohiggins:
IMO slug should be unique; two policies should not be allowed to have the same slug. If we're also looking for OMB Policy ID and primary key, then slug shouldn't be allowed to have the same value as any present values of those fields either…

api/reqs/tests/model_tests.py
```diff
@@ -28,3 +31,12 @@ def test_original_url():
...
     policy.document_source = Mock(url='http://example.com/uploaded')
     assert policy.original_url == 'http://example.com/uploaded'
+
+
+@pytest.mark.django_db
+def test_slug_is_created_from_title_on_save_if_slug_is_empty():
+    policy = mommy.make(models.Policy)
+    policy.title = 'hello there'
+    policy.slug = ''
+    policy.save()
```
cmc333333: A few small details that might make this slightly easier to read in the future:
* `mommy.make` will _save_ the record to the database, which might not be intended. `mommy.prepare` will not
* `mommy.make` and `.prepare` allows us to set model properties via kwargs

I think that could leads us to something like:
```python
policy = mommy.prepare(models.Policy, title='hello there', slug='')
policy.save()  # not using `make` to emphasize the save() method
assert policy.slug == 'hello-there'
```
What do you think?

api/reqs/tests/views_policies_tests.py
```diff
@@ -180,3 +180,15 @@ def test_pk_id():
...
     assert response['id'] == pk_id
+
+
+@pytest.mark.django_db
+def test_slug():
+    client = APIClient()
+    slug = "hello-there"
+    path = "/policies/{0}".format(slug)
+    response = client.get(path)
+    assert response.status_code == 301
```
cmc333333: Should this be a 404? Maybe it's not following a redirect to `/policies/hello-there/`?

api/reqs/tests/views_policies_tests.py
```diff
@@ -180,3 +180,15 @@ def test_pk_id():
     mommy.make(Policy, pk=pk_id)
     response = client.get(path + '.json').json()
     assert response['id'] == pk_id
+
+
+@pytest.mark.django_db
+def test_slug():
+    client = APIClient()
+    slug = "hello-there"
+    path = "/policies/{0}".format(slug)
```
cmc333333: Mind using f-strings, pretty please?
```python
path = f"/policies/{slug}/"
```

api/reqs/models.py
```diff
@@ -86,6 +87,7 @@ class Meta:
         max_length=32, choices=[(e.name, e.value) for e in PolicyTypes],
         blank=True
     )
+    slug = models.SlugField(max_length=title.max_length, blank=True)
```
cmc333333: What do you think about disallowing blank here? We should expect it to always be set, yes?

cmc333333:
I agree with @tadhg-ohiggins and will expand his suggestions a bit, but I also think this is a step up from what we have now, so don't think we need to decide before merging.

* priority: we could also do a sql `OR` and have them all running conceptually parallel. While not _impossible_, in practice, there shouldn't be overlap. It might make sense to combine the slug and omb_policy_id into a single field at some point to make the logic more apparent. I think we'll need to keep the integer pk for a while, though, as it's a pretty deep-seated default across Django.
* `save()`: I think this is a good safety net; we may be skipping over Django's admin soon
* `slug` in API: Agreed that in the next round, we'll want to surface the `slug` field in the API. We don't currently link to any of the pages that'd use this endpoint in the UI, yet, so this is more of a proof-of-concept.
* `slugify` variation: Good point. I suspect this is likely to remain constant due to the number of Django users that'd be upset if it didn't, however.
* testing migrations: We don't have any infrastructure for this, but I think it'd be worthwhile. It looks pretty straight forward to add a unit test of the `populate_slug` method, but we might want to delay writing the more general case until we prioritize it as a group -- I suspect it'd be large.
* same slug: Agreed that we shouldn't allow multiple policies with the same slug. I don't have a good sense how often this will arise in practice (OMB _does_ release revisions of previous policies, but that's another can of worms). In the meantime, the `first()` approach will prevent an explosion (though it _will_ silently fail).

cmc333333:
Heads up that #601 moved the "priority" logic into a separate function. Shouldn't be hard to merge.

api/reqs/tests/views_policies_tests.py
toolness: Ummm to be honest I was just copy-pasting the previous test, which tested to make sure that retrieving policies by OMB policy ids worked. I would've expected it to be a 404 myself... maybe the path should end with a `/` and it should verify that it 404's?

api/reqs/tests/views_policies_tests.py
toolness: Oh sure. Should I change the test above it to use f-strings too, since this new test is just a copy-paste of that one?

api/reqs/tests/model_tests.py
toolness: Oh great idea! I'm unfamiliar w/ model mommy so I didn't know about the `make` vs. `prepare` distinction, thanks for explaining it.

api/reqs/models.py
toolness: Yep--I had originally set it to always be set, but then `makemigrations` wanted me to provide a hard-coded default value, which I couldn't provide, since each policy's slug is based on each policy's title. I can think of two options here:

1. Read up on django's docs and see if there's a way to e.g. pass in a function for the default value, so that it can be dynamically set for each policy. This would essentially mean combining the two current migrations into one combined migration.

2. Add a third migration that simply changes `blank=False` and provides a nonsensical default value, since we know from the previous data migration that no blank slugs can possibly exist in the db (at least, not if all the migrations are run in one batch).

cmc333333: Option 2 has worked well for me in the past. Option 1 would be tricky due to the need to mix Python and SQL.

api/reqs/tests/views_policies_tests.py
cmc333333: No obligation, but that would be appreciated.

cmc333333:
@toolness says he's okay with merging this as is, and I'd like to get it in before demo, so I'm going to merge this in his stead.

PR #601, opened 2017-11-01
Allow documents to be queried by M-number, add policy meta data

cmc333333:
This re-uses the policy-id-or-m-number logic @rememberlenny wrote to do the same for whole documents. It also adds policy details to the root of the serialized tree.


cmc333333:
Thanks @rememberlenny . I'll re-base and merge.

PR #603, opened 2017-11-01
Warn about mismatching data when importing XML.

cmc333333:
We (currently) have two sources of metadata for a document: the document
itself and a Policy object in the database. Unfortunately, this allows for the
same data to be stored in two places (and potentially conflict). We should at
least warn the user when we see this happening.


cmc333333:
Thanks @tadhg-ohiggins

PR #604, opened 2017-11-01
Be slightly more consistent with variable names.

cmc333333:


cmc333333:
S'all good. Thanks @tadhg-ohiggins and @rememberlenny

Issue #591, opened 2017-10-28
Style full policy view based on design


rememberlenny:
# Task

Based on the designs from https://github.com/18F/omb-eregs/issues/527#issuecomment-340073841 create the policy designs.

---

#### hey gang! here be me first pass ☠️ 

I assume all of this will change or be tweaked, but it's a great start! If anyone sees something we need to add or if you need any specificity, don't hesitate to ask! 

[sketch file here](https://github.com/18F/omb-eregs/files/1423318/VisualDesign0.3.6-WIP.zip) for those that want it

### text and color
![omb styles](https://user-images.githubusercontent.com/12564977/32123106-9b57b80a-bb28-11e7-85cf-0ffa45626dad.png)

### mobile typography
![omb mobile type](https://user-images.githubusercontent.com/12564977/32123202-f65e5a10-bb28-11e7-86fa-9c0ec63ba39d.png)

### main policy page, with a table
![full policy page](https://user-images.githubusercontent.com/12564977/32123210-fc7e6192-bb28-11e7-956f-abf8279fba0e.png)

### footnote in opened state (hover at the top of the page)
![footnote](https://user-images.githubusercontent.com/12564977/32123230-05e44a26-bb29-11e7-8b1f-370f4288729d.png)

PR #605, opened 2017-11-01
Adding policy styling for document view

rememberlenny:
screen shot 2017-11-02 at 2 34 41 pm

# Context

https://github.com/18F/omb-eregs/issues/591

# What

- Adding the css folder structure to better organize the stylesheets.
- Add the high level typography recommendations for mobile and desktop

# What it isnt

- Footnotes styling/components
- Table styling


cmc333333:
@rememberlenny are we sure we want these changes on _all_ pages? It appears to break the padding on most, but also makes headings gold where they probably shouldn't be.

screen shot 2017-11-02 at 9 36 34 am
screen shot 2017-11-02 at 9 36 22 am
screen shot 2017-11-02 at 9 35 50 am

ui/css/module/_footnotes.scss
```diff
@@ -0,0 +1,18 @@
+.node-footnote {
+  display: none;
+}
```
cmc333333: Let's not hide these in CSS. If we don't want to display them, we shouldn't render them.

ui/css/state/_print.css
```diff
@@ -5,15 +5,15 @@
   }
 
   .usa-disclaimer, .navbar-title, .req .metadata, .req .req-text {
-    color: #111 !important;
+    color: $color-gray-darker !important;
```
cmc333333: If using variables, this should probably be an `scss` file, hey?

cmc333333:
With these updates the rest of the site is running pretty smoothly. I think the only regression is on the policy listing page; the new font causes some text to wrap when it shouldn't.

screen shot 2017-11-03 at 10 04 45 am

We can probably just fiddle with the col widths to fix it.

cmc333333:
Spinning off the Code Climate issues into a separate, followup ticket: #613; I'm going to merge this as is so it'll be ready for demo.

Issue #597, opened 2017-10-31
Policy detail: Preamble/heading


cmc333333:
From the designs in https://github.com/18F/omb-eregs/issues/527#issuecomment-340073841, we need to pull out and re-order some meta data.

<img width="668" alt="screen shot 2017-10-31 at 12 52 21 pm" src="https://user-images.githubusercontent.com/326918/32237146-8c0d3950-be3a-11e7-8567-4ee3a06c4714.png">

Unfortunately, this data is currently split across the `Policy` object and the document text

| Text | Field on `Policy` | Tag in XML Doc |
|---|---|---|
| M-16-19 | omb_policy_id | policyNum |
| Memorandum for... | - | policyTitle |
| Data Center... | title | subject |
| See original | uri | - |
| From: ... | - | from |
| Issued ... | issuance | published |

cmc333333:
This needs additional styling love before we close.

PR #606, opened 2017-11-02
Initial preamble display

cmc333333:
Serving as a first step of the solution for #597, this adds a set of metadata fields to the top of a policy and hides the document's "preamble" (if present). This isn't well styled, but should be semantically okay. Easiest to review changeset-by-changeset

Along the way, it also:
* Allows a url like `/document/M-16-19`
* Fixes the pdf url from the document API
* Unstyles `sec` nodes

Looks like
screen shot 2017-11-02 at 7 00 04 pm

### DIFF
--- a/api/document/serializers.py
+++ b/api/document/serializers.py
@@ -29,8 +29,8 @@ class PolicySerializer(serializers.ModelSerializer):
         fields = (
             'issuance',
             'omb_policy_id',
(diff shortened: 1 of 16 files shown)


cmc333333:
Thanks @rememberlenny and @tadhg-ohiggins! I'll rebase and merge.

Issue #600, opened 2017-10-31
Policy detail: List elements


cmc333333:
Ensure we have a reasonable rendering for list elements, numbered lists, etc.

Remaining tasks

- [ ] First indent should be 32px
- [ ] ~~Second indent should be 80px~~ @rtwell has said we can do 32px per level.
- [ ] It would be great if we could scale these indents on mobile (i can see them not needing quite as much space)
- [ ] Numbers bold and black please
- [ ] Slightly larger bullets, if possible (6px would be perf!)

cmc333333:
There's an implementation now, but should check in w/ design before closing.

rtwell:
## Mark up here for easy viewing:
![lists](https://user-images.githubusercontent.com/12564977/32563272-72568ce0-c476-11e7-9eff-ccd24f8743a4.png)
## List o things

- [ ] First indent should be 32px
- [ ] Second indent should be 80px
- [ ] It would be great if we could scale these indents on mobile (i can see them not needing quite as much space)
- [ ] Numbers bold and black please
- [ ] Slightly larger bullets, if possible (6px would be perf!)

cmc333333:
Resolved with #647

PR #607, opened 2017-11-02
Initial list display

cmc333333:
Per #600, we want to have some reasonable formatting for lists/list items. We generally need to display the `marker` (e.g. "1." or "(c)") separately from the text of the paragraph, so this creates a separate field to represent that information (see the individual changesets for more detail around philosophy).

I'm not keen on the amount of padding, but we can work on that in a followup styling pass.

Looks like
screen shot 2017-11-02 at 3 04 32 pm
screen shot 2017-11-02 at 3 04 20 pm


cmc333333:
Thanks @tadhg-ohiggins, and I agree. Two thoughts come to mind, though:
* The two major AKN author tools I've seen parse markers from what the user types, so the burden is less (on authors, at least).
* I think content can _also_ be incrementally improved. So maybe the first round of a document is a bunch of paragraphs. Maybe a month later, an editor notices it's rendering ugly and so makes it make semantic by making some lists. Afterwards, someone realizes a link to paragraph b from some other document doesn't work because the data doesn't include that info yet, so they go through and fix it.

I'll rebase and merge this PR.

tadhg-ohiggins:
@cmc333333 Good points, particularly about incremental improvement.

Issue #565, opened 2017-10-20
view math equations within policy text (?)


carodew:
is this what we mean by maths?

cmc333333:
Heads up that one strategy here is to rely on [KaTeX](https://khan.github.io/KaTeX/), which renders most of TeX's math instructions. Might look like:

<img width="282" alt="screen shot 2017-10-23 at 1 22 15 pm" src="https://user-images.githubusercontent.com/326918/31903175-5a925e00-b7f5-11e7-8a01-e603db52c952.png">

cmc333333:
@tadhg-ohiggins will take a first pass at implementing these equations. Major decisions include:

1. We'll move the footnote _out_ of the math script to make this a little easier

I'd recommend https://khan.github.io/KaTeX/. There are a few react-katex libraries, but I think they're all outdated.

cmc333333:
@tadhg-ohiggins implemented a first pass in #609, but we've noticed some wonkiness around text sizes.

cmc333333:
I think #634 and #645 finish this.

cmc333333:
All merged; closing.

PR #609, opened 2017-11-03
Add math support

tadhg-ohiggins:
Two plus two is four.
We think. At least now the equations
Are easier to read.

Add preliminary support for math display using KaTeX. I'm not sure this is the best way to integrate it with React and next.js, but it appears to work, which is a good start.

Addresses https://github.com/18F/omb-eregs/issues/565.


ui/components/header-footer.js
```diff
@@ -70,6 +70,8 @@ export default function HeaderFooter({ children, showSearch }) {
       <Head>
         <title>OMB Policy Library (Beta)</title>
         <link rel="stylesheet" href="/static/styles.css" />
+        <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.9.0-alpha1/katex.min.css" integrity="sha384-8QOKbPtTFvh/lMY0qPVbXj9hDh+v8US0pD//FcoYFst2lCIf0BmT58+Heqj0IGyx" crossOrigin="anonymous" />
```
cmc333333: We may be able to [import this](https://github.com/18F/omb-eregs/blob/da8c484254201380ea2a577152642d30089525f6/ui/css/main.scss#L7-L8) into our styles from the npm library, which would allow us to bundle the CSS rather than make another HTTP call.

ui/components/header-footer.js
```diff
@@ -70,6 +70,8 @@ export default function HeaderFooter({ children, showSearch }) {
...
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.9.0-alpha1/katex.min.js" integrity="sha384-GR8SEkOO1rBN/jnOcQDFcFmwXAevSLx7/Io9Ps1rkxWp983ZIuUGfxivlF/5f5eJ" crossOrigin="anonymous" />
```
cmc333333: Is this needed? We're compiling KaTeX into the JS, so I don't think we need this separate, client-only asset.

ui/components/nodeRenderers/math.js
```diff
@@ -0,0 +1,26 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import katex from 'katex';
+
+
+export default function TeXMath({ children, docNode }) {
+  return (
+    <div id={docNode.identifier}>
+      { children }
```
cmc333333: Do you think we should be accounting for children of `math` nodes? I'm thinking they'll always be a leaf.

If we're keeping this, should it be swapped down a line so that the content of this node is rendered above its children?

ui/package.json
```diff
@@ -52,6 +52,7 @@
     "express": "^4.15.4",
     "helmet": "^3.8.1",
     "http-status": "^1.0.1",
+    "katex": "^0.8.0",
```
cmc333333: Awesome -- is there a corresponding change to `npm-shrinkwrap.json`? Try running
```
docker-compose run -rm npm shrinkwrap
```
to generate it, if not.

ui/components/header-footer.js
tadhg-ohiggins: I ran into font import errors when trying this, so I'm going to leave it as is for the moment.

ui/components/header-footer.js
tadhg-ohiggins: 👍

ui/components/nodeRenderers/math.js
tadhg-ohiggins: Good point; they shouldn't have child elements.

ui/package.json
tadhg-ohiggins: 👍

PR #611, opened 2017-11-03
Add temporary, simple footnote style.

cmc333333:
We won't be able to get the footnotes at the bottom of the page prior to demo.
In its stead, render a display that's at least evocative of the original
design.

Looks like
screen shot 2017-11-03 at 10 58 27 am


cmc333333:
Thanks @tadhg-ohiggins. I'll rebase and merge.

PR #614, opened 2017-11-03
Allow use of non-ints when looking up documents.

cmc333333:
The API now allows documents to be looked up by their policy's M-number (and
soon, slug); allow that on the frontend as well.

This pulls out the only part of #606 that feels essential for demo


cmc333333:
Thanks @tadhg-ohiggins!

PR #616, opened 2017-11-06
Rework component rendering strategy

cmc333333:
Previously, the `Document` component was responsible for recursively rendering
all children within a node tree. The textual content of a node would be passed
along as its `renderedContent` property. This has been problematic for a few
reasons: not only is the `renderedContent` field very awkward, but rendering
every node in the tree is quite inefficient if we're not going to render a
parent. As an immediate need, too, it's very difficult to do something like
rendering footnotes out of sequence; the Policy component can't easy access
the recursive rendering infrastructure due to recursive imports.

This reworks the approach so that each node is responsible for rendering its
(docNode) children (or not) and the `renderedContent` is now passed as the
body (React's `children` prop). Combined this should feel more React-y and
allow us to avoid rendering nodes that don't need it. It also solves the need
around footnotes as components now have access to the `renderNode` function to
use at will.


cmc333333:
Thanks @rememberlenny and @tadhg-ohiggins

PR #626, opened 2017-11-07
Fix n+1 problem around footnote citations.

cmc333333:
When serializing a footnote citation, we grab that footnotecitations'
footnote_node (to fetch its identifier). Unfortunately, we hadn't prefetched
any of that data, which meant we were executing an extra query per
footnotecitation.


cmc333333:
Thanks @tadhg-ohiggins !

PR #627, opened 2017-11-07
Remove the need to repeat policy in tree children.

cmc333333:
This resolves a minor annoyance around creating document trees. All nodes
within a single document should be associated with the same policy, yet we
needed to specify the policy for every node that we added.

This removes the redundancy by adding the policy as an attribute of the
underlying graph and using that whenever adding a new child node. It's still
possible to manually set the policy differently in a child (though I don't see
a use case).


cmc333333:
Thanks @tadhg-ohiggins

Issue #624, opened 2017-11-06
Display footnotes at bottom of page


yowill:

cmc333333:
Design is in [this comment](https://github.com/18F/omb-eregs/issues/566#issuecomment-341149927)

rtwell:
## Quick markup with tweaks:
![footnotes](https://user-images.githubusercontent.com/12564977/32560474-5f26a8aa-c46f-11e7-9721-897b4633f5e7.png)

## Changes listed here for easy reference:
- [ ] Line at the top should be our blue (#1067A6)
- [ ] Lessen the indent 50%
- [ ] The footnotes themselves should be sans serif, and smaller @ `14px (.875em)` and the line-height should be` 18px (1.125em)` 
- [ ] The numbers are the same size as the text, except they are `bold` and `black`
- [ ] it looks as if the baseline between the numbers and the footnotes is slightly off, if we could make sure they are aligned, that would be fantastic!
- [ ] Text should not break the right margin
- [ ] URLs should wrap
- [ ] If possible, can we take URLs and roll them into the text before them? And if we could get sassy,  **if the link points to an external web page** use [this icon](https://github.com/18F/omb-eregs/files/1454628/external-link.svg.zip). **If the link points to a file of any kind**, use [this icon](https://github.com/18F/omb-eregs/files/1454621/external-file.svg.zip). Here's an example:
![image](https://user-images.githubusercontent.com/12564977/32560639-c6047502-c46f-11e7-921c-b33f5143d57f.png)

cmc333333:
I've broken out the bits about external url linking to #656 and #657.

cmc333333:
Closing per Micah's okay in #650

PR #628, opened 2017-11-07
Display footnotes at bottom of page

cmc333333:
This modifies the API to include footnotes in the `meta` data for the root of a document and any `table`s. It then uses that new data to display footnotes at the bottom of the `Policy` component. This should resolve a majority of #624, though we'll want a design review before closing.

## Looks like
(note that footnotes aren't displaying inline anymore)
screen shot 2017-11-07 at 6 07 37 pm 

---

screen shot 2017-11-07 at 6 07 48 pm

---

screen shot 2017-11-07 at 6 08 04 pm


cmc333333:
Thanks @tadhg-ohiggins and @rememberlenny ! I'll rebase and merge.

Issue #620, opened 2017-11-06
Interface around displaying a footnote when clicked


yowill:

cmc333333:
Design is in [this comment](https://github.com/18F/omb-eregs/issues/566#issuecomment-340080892)

cmc333333:
#629 did a good chunk of this work, thanks @rememberlenny!

Some remaining pieces that I see:
* Use real data. We're currently using a mock footnote, but real data is now available on `content.footnode_node`
* Add a close link to the bottom of the footnote.
* Style the inline footnote to match the design as closely as possible. This should not affect footnotes elsewhere on the page.
* Triangle cutout for the speech bubble. I'd bet this will be difficult, so I'm calling it out separately from the remaining styles.

cmc333333:
@rtwell says that the last (the triangle cutout) isn't essential for shipping -- if we can't implement this sprint, let's split into another issue to prioritize.

rememberlenny:
Finalized changes on the `content.footnode_node`, with exception to the triangle cutout.

cmc333333:
I've moved the triangle cutout work to #651

Issue #566, opened 2017-10-20
view policy footnotes


carodew:

carodew:
Regarding footnotes, we want to be sure we don't dump screen reader and keyboard users at the end of a page without providing them a way to easily get back to where they left off in the text. 

![screenreader - footnotes](https://user-images.githubusercontent.com/10144074/31823204-4c1490a6-b579-11e7-8ad7-bc0d0eb0175d.png)

(comment copied from #527 )

carodew:
one bit of inspiration – https://harvardlawreview.org/2017/10/bureaucracy-and-distrust-landis-jaffe-and-kagan-on-the-administrative-state/

rtwell:
lil bit of crossover with #527 

![image](https://user-images.githubusercontent.com/12564977/32124588-e8aea1cc-bb2d-11e7-98df-559efe438e43.png)

cmc333333:
@rememberlenny will take on implementation. Major decisions include:

1. Only one footnote should be open at a time
1. We'll start by shifting content around (rather than overlaying the footnote on existing content)
1. Sequential footnotes should be merged together, e.g. "Footnote 4, 5, 6" should be a single button if they're adjacent.
1. Footnotes should also appear at the bottom of the document

carodew:
@rememberlenny @rtwell do we have or need to have text styles for the list of footnotes at the bottom of the document?

rtwell:
@cmc333333 ask and ye shall receive

Sketch file [footnote-specs.zip](https://github.com/18F/omb-eregs/files/1434728/footnote-specs.zip)

![footnotes-compiled](https://user-images.githubusercontent.com/12564977/32283839-7ddc87b4-bef3-11e7-82e1-e068427aedf2.png)

cmc333333:
We're a far way from this on the implementation front. I'd recommend we split this into smaller issues. Maybe:
1. Display footnotes at bottom of page
2. Interface around displaying a footnote when clicked
3. Only allow one footnote to be opened at a time (depends on 2)
4. Combine sequential footnotes (depends on 2)

rememberlenny:
Closing in place of the newly created subtasks.

PR #629, opened 2017-11-08
Rendering footnote inline function

rememberlenny:
# Reference

- https://github.com/18F/omb-eregs/issues/620
- https://github.com/18F/omb-eregs/issues/566

# What

Creating the component behavior that saves footnote citation state and displays footnote content inline.

click

### DIFF
--- a/ui/components/content-renderers/footnote-citation.js
+++ b/ui/components/content-renderers/footnote-citation.js
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 
-import {renderNode, renderContent} from '../../util/render-node';
+import renderContent from '../../util/render-node';
 
 import Footnote from '../node-renderers/footnote';
 import Link from '../link';
--- a/ui/components/content-renderers/footnote-citation.js
+++ b/ui/components/content-renderers/footnote-citation.js
@@ -33,10 +33,6 @@ export default class FootnoteCitation extends React.Component {
     this.setState({ expanded: !this.state.expanded });
   }
 
-  renderFootnote() {
-    return renderNode(demoData);
-  }
-
   render() {
     let citation;
     const expanded = this.state.expanded;
@@ -50,8 +46,8 @@ export default class FootnoteCitation extends React.Component {
     if (expanded) {
       citation = (
         <span>
-          { this.renderFootnote() }
-          <span className="clearfix"></span>
+          { renderNode(demoData) }
+          <span className="clearfix" />
         </span>
       );
     }
--- a/ui/components/content-renderers/footnote-citation.js
+++ b/ui/components/content-renderers/footnote-citation.js
@@ -1,8 +1,9 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 
-import renderNode from '../../util/render-node';
+import {renderNode, renderContent} from '../../util/render-node';
 
+import Footnote from '../node-renderers/footnote';
 import Link from '../link';
 
 /* eslint-disable */
@@ -46,7 +47,9 @@ export default class FootnoteCitation extends React.Component {
     if (expanded) {
       citation = (
         <span>
-          { renderNode(demoData) }
+          <Footnote key={demoData.identifier} docNode={demoData}>
+            { renderContent(demoData.content) }
+          </Footnote>
           <span className="clearfix" />
         </span>
       );
--- a/ui/components/content-renderers/footnote-citation.js
+++ b/ui/components/content-renderers/footnote-citation.js
@@ -1,14 +1,15 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 
+import renderNode from '../../util/render-node';
+
 import Link from '../link';
 
+/* eslint-disable */
+const demoData = {"children":[],"content":[{"content_type":"__text__","text":"The FDCCI was first established by OMB â€œMemo for CIOs: Federal Data\n        Center Consolidation Initiative,â€ issued on February 26, 2010, and\n        modified by subsequent memoranda."}],"depth":3,"identifier":"policy_1__sec_1__para_1__footnote_1","marker":"1","node_type":"footnote","requirement":null,"text":"The FDCCI was first established by OMB â€œMemo for CIOs: Federal Data\n        Center Consolidation Initiative,â€ issued on February 26, 2010, and\n        modified by subsequent memoranda.","type_emblem":"1","policy":{"issuance":"2016-08-01","omb_policy_id":"M-16-19","original_url":"https://www.whitehouse.gov/sites/default/files/omb/memoranda/2016/m_16_19_1.pdf","title":"Data Center Optimization Initiative (DCOI)"}};
+/* eslint-enable */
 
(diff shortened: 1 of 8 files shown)


rememberlenny:
@cmc333333 Im getting stuck on where these errors are coming from. Locally, its different than the circleci. Could I borrow your hand to can take a look tomorrow?

ui/components/content-renderers/footnote-citation.js
```diff
@@ -1,17 +1,77 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 
+import renderNode from '../../util/render-node';
+
 import Link from '../link';
 
+const demoData = {"children":[],"content":[{"content_type":"__text__","text":"The FDCCI was first established by OMB â€œMemo for CIOs: Federal Data\n        Center Consolidation Initiative,â€ issued on February 26, 2010, and\n        modified by subsequent memoranda."}],"depth":3,"identifier":"policy_1__sec_1__para_1__footnote_1","marker":"1","node_type":"footnote","requirement":null,"text":"The FDCCI was first established by OMB â€œMemo for CIOs: Federal Data\n        Center Consolidation Initiative,â€ issued on February 26, 2010, and\n        modified by subsequent memoranda.","type_emblem":"1","policy":{"issuance":"2016-08-01","omb_policy_id":"M-16-19","original_url":"https://www.whitehouse.gov/sites/default/files/omb/memoranda/2016/m_16_19_1.pdf","title":"Data Center Optimization Initiative (DCOI)"}};
```
cmc333333: Eslint'll probably complain about this line, so we may want to turn it off. I think that'd look like:
```js
/* eslint-disable */
const demoData = ...
/* eslint-enable */
```

ui/components/content-renderers/footnote-citation.js
```diff
@@ -1,17 +1,77 @@
...
+  return {
+    expanded: false,
+  };
+}
+
+export default class FootnoteCitation extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = stateSetter();
```
cmc333333: What do you think about removing a layer of indirection and doing the initialization in the constructor:
```js
this.state = { expanded: false }
```
I suspect this file will grow soon, so any effort to minimize complexity seems like a win.

ui/components/content-renderers/footnote-citation.js
```diff
@@ -1,17 +1,77 @@
...
+    this.handleCitationClick = this.handleCitationClick.bind(this);
+  }
+
+  componentDidMount() {
+  }
+
+  componentWillUnmount() {
+  }
```
cmc333333: Since these methods are empty, let's remove them. We'll inherit the noop defaults from `React.Component`.

ui/components/content-renderers/footnote-citation.js
```diff
@@ -1,17 +1,77 @@
...
+
+  handleCitationClick(e) {
+    e.preventDefault();
+    if (this.state.expanded) {
+      this.setState({ expanded: false });
+    } else {
+      this.setState({ expanded: true });
+    }
```
cmc333333: Should we make this more concise by using negation?
```js
e.preventDefault();
this.setState({ expanded: !this.state.expanded });
```

ui/components/content-renderers/footnote-citation.js
```diff
@@ -1,17 +1,77 @@
...
+  }
+
+  renderCitation() {
```
cmc333333: I think we'll want to rename this from `renderCitation` to something like `renderFootnote`. A citation is the bit of text that _refers_ to content, not the content itself; misnaming this seems prone to confusion going forward.

ui/components/content-renderers/footnote-citation.js
```diff
@@ -1,17 +1,77 @@
...
+    return renderNode(demoData);
+  }
+
+  render() {
+    let citation = '';
```
cmc333333: Similar note re: naming.

Also: what do you think about leaving this undefined? Giving it a string default has two drawbacks in my mind:
1. it implies that the variable will contain a string (which isn't true if we're expanded)
2. I'm not positive, but I'd bet it causes React to jump through a few more hoops to track and render the string rather than ignoring the content

ui/components/content-renderers/footnote-citation.js
```diff
@@ -1,17 +1,77 @@
...
+    const href = `#${this.props.content.footnote_node}`;
+    const link = (
+      <Link onClick={ (e) => { this.handleCitationClick(e); }}
```
cmc333333: Can this be simplified to:
```jsx
<Link onClick={this.handleCitationClick}>
```

ui/components/content-renderers/footnote-citation.js
```diff
@@ -1,17 +1,77 @@
...
+        className="footnote-link"
+        href={href}>
+        <sup>Footnote { this.props.content.text }</sup>
+      </Link>
+    );
+    if (this.state.expanded) {
+      citation = (
+        <span>
+          { this.renderCitation() }
+          <span className="linebreaker"></span>
```
cmc333333: Could we get the same effect by using the `clearfix` class? Fewer utility classes like this means less we have to remember.

ui/components/content-renderers/footnote-citation.js
```diff
@@ -1,17 +1,77 @@
...
+    return (
+      <cite data-citation-target={href} >
+        { link }
+        { citation }
+      </cite>
+    );
+  }
+}
+
+FootnoteCitation.propTypes = propTypes;
```
cmc333333: What do you think about adding these as a static member on the class? We hadn't been doing that before because we didn't have babel configs to handle it, but I _think_ Next.js' config does. It'd look something like

```js
class FootnoteCitation extends React.Component {
  static propTypes = {
    ...
  };
  ...
}
```

ui/components/content-renderers/footnote-citation.js
```diff
@@ -1,17 +1,77 @@
...
+    e.preventDefault();
+    if (this.state.expanded) {
+      this.setState({ expanded: false });
+    } else {
+      this.setState({ expanded: true });
+    }
+  }
+
+  renderCitation() {
+    return renderNode(demoData);
```
cmc333333: Heads up that `renderNode` is actively hiding footnotes now; you'll want to render that first component (`Footnote` or otherwise) yourself. See [this chunk](https://github.com/18F/omb-eregs/blob/c5b691ba2a516a96e7974c2d74c1f5a10f823228/ui/components/node-renderers/policy.js#L24-L26) of the Policy component for an example.

rememberlenny:
@cmc333333 
Finally got this green. Adding unit tests.

ui/css/layout/_layout.scss
```diff
@@ -38,3 +38,7 @@ body {
     padding-right: 1.5em;
   }
 }
+
+.node-policy {
+  max-width: 100%;
+}
```
cmc333333: I think we'll be able to remove this as part of #636

PR #630, opened 2017-11-08
Rework "meta" field serialization.

cmc333333:
This leverages DRF's serialization framework to separate out the logic
necessary to generate the "meta" field from the DocCursor. This should make it
easier to test and allow us to focus future changes.


cmc333333:
Thanks @tadhg-ohiggins

PR #631, opened 2017-11-08
Include footnote content in citation annotation

cmc333333:
This extends the FootnoteCitation serialization to include the referenced node, which should make it easier to display (and lines up well with #629).

I hope @tadhg-ohiggins will enjoy the use of `singledispatch` over `abc`.

### DIFF
--- a/api/document/tests/views_test.py
+++ b/api/document/tests/views_test.py
@@ -87,10 +87,16 @@ def test_query_count(client):
             req.topics.add(mommy.make(Topic))
         req.docnode = req_node
         req.save()
-    # select 3 nodes as footnotes
-    for citing, footnote in zip(random.sample(subtree_nodes, 3),
-                                random.sample(subtree_nodes, 3)):
-        citing.footnotecitations.create(start=0, end=1, footnote_node=footnote)
+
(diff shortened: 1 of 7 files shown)


api/document/tests/views_test.py
```diff
@@ -87,10 +87,16 @@ def test_query_count(client):
...
+
+    # select 3 nodes to add footnote citations
+    citing_nodes = random.sample(list(root.walk()), 3)
+    footnotes = [citing.add_child('footnote') for citing in citing_nodes]
+    root.nested_set_renumber()
+    for node in root.walk():
+        node.model.save()
+    for citing, footnote in zip(citing_nodes, footnotes):
+        citing.model.footnotecitations.create(start=0, end=1,
+                                              footnote_node=footnote.model)
```
cmc333333: This refactor is a little hinky -- it's trying to move away from a situation where there's recursion between the citing node and the cited footnote. Since the footnote citation now includes its full tree of the footnote, we have to guarantee that said tree doesn't also contain the citation.

cmc333333:
Thanks @tadhg-ohiggins; I'll rebase and merge.

PR #632, opened 2017-11-08
Refactor redundant node saves.

cmc333333:
90% of the time when we re-number nodes in our document tree (to agree with
the nested-set model), we want to create them immediately afterwards. The
removes a chunk of redundancy by having the renumber method also save the
DocNodes.


cmc333333:
Thanks @tadhg-ohiggins. I'll rebase and merge.

PR #635, opened 2017-11-08
Activate eslint import ordering plugin.

cmc333333:
The plugin *doesn't* make demands about import order within a "group" (e.g.
builtins, external libs, parent, etc.), but does demand those groups be in a
particular order. It also discourages newlines between groups.

### DIFF
--- a/ui/.eslintrc
+++ b/ui/.eslintrc
@@ -18,6 +18,7 @@
     "no-unsafe-innerhtml/no-unsafe-innerhtml" : 2,
     "react/jsx-filename-extension": [1, {"extensions": [".js", ".jsx"]}],
     "import/no-named-as-default": 0,
+    "import/order": 2,
     "scanjs-rules/assign_to_hostname" : 2,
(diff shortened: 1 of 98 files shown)


toolness:
Huh, that's surprising to me that it discourages newlines between groups. What is our general group order, btw? Is it like python, where it's builtins then external libs then first-party libs?

cmc333333:
Yeah, I think there's a choice between "make it operate like isort's default" and "make it follow [AirBnb's config](https://github.com/airbnb/javascript/blob/44dbd0bdc41d08eb5de8ad698099ae44240f4b0d/packages/eslint-config-airbnb-base/rules/imports.js#L139)". I went with the latter, but AirBnB's not even enforcing their config, so it might not be worth following.

That link also describes the order of the groups, but a [full listing](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md#groups-array) is mentioned in the plugin's docs.

cmc333333:
I'm going to go ahead and close this seeing that no one wants to commit to it.

PR #640, opened 2017-11-09
Enable webpack dev middleware polling if USE_POLLING is set.

toolness:
Some docker setups, such as Windows, don't notify processes of changes to the filesystem like Linux does, so we need to use polling.

This detects if `USE_POLLING` is set in the host environment, and if so, configures webpack's dev middleware to use it (both in next.js and the CSS webpack process).

To do:
- [x] Document `USE_POLLING` in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,12 +101,23 @@ There are two types of entry points:
 
 ### Resolving common container issues
 
+#### Bundles aren't being rebuilt when I change them
+
+Try setting `USE_POLLING=true`, either in your host shell environment, or
+via an [`.env` file](https://docs.docker.com/compose/env-file/). This will
+force all the watchers to use filesystem polling instead of OS notifications,
+which works better on some platforms, such as Windows.
+
+#### Conflicting ports
+
 If you see an error about a conflicting port, try spinning down the running
 services (including those associated with integration tests).
 ```sh
 docker-compose down
 docker-compose -p integration_tests down
 ```
+
+#### Restarting from scratch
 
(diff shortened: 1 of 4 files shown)


toolness:
@cmc333333, do you have a preference for how to have users define environment variables?

I can think of a few options:

1. Rely on users to define them on a per-service basis in a `docker-compose.override.yml` file.
2. Have users run `cp .env.sample .env` during initial setup; they can then edit any of the variables directly in their `.env`.  We can also annotate the `.env.sample` with examples/docs/etc.
3. Don't advise anything, relying on users to define the environment variables in their Docker host however they like. The only downside to this is that `docker-compose` will warn them that `USE_POLLING` is undefined whenever they run `docker-compose`, which is kind of annoying.
4. I suspect there are other ways of doing this that I haven't thought of yet.

cmc333333:
I like option 3 the most for now as it requires the least cognitive overhead to get started. If we start to need more env vars (e.g. if we split the ui and api into two sets of docker-compose setups and need to pass an IP between them), I'd vote for 2.

It [looks like](https://docs.docker.com/compose/compose-file/#variable-substitution) we can use defaults for our env vars now, so we might be able to avoid that warning message.

toolness:
Ok, I've implemented option 3, though using default values for environment variables appears to require the 2.1 compose file format, so I've "upgraded" our `docker-compose.yml` to use that... Hopefully that shouldn't cause any weird side effects...

toolness:
Ok this is good to go, assuming circle is happy!

PR #642, opened 2017-11-09
Styling footnotes and remove demo data

rememberlenny:
# Reference

- https://github.com/18F/omb-eregs/issues/620
- https://github.com/18F/omb-eregs/issues/566

# What

Finishing up styles on the footnotes and remove demo data.

### DIFF
--- a/ui/css/module/_footnotes.scss
+++ b/ui/css/module/_footnotes.scss
@@ -42,16 +42,18 @@
     @extend .col;
     @extend .col-1;
     font-style: normal;
-    width: 1em;
+    width: 2.5em;
     font-size: .9em;
     line-height: 1.5em;
+    text-align: right;
+    padding-right: .5em;
   }
 
   .footnote-text {
     @extend .col;
     @extend .m0;
     @extend .col-11;
-    width: calc(100% - 1.5em);
+    width: calc(100% - 3em);
     color: $color-gray-medium;
   }
 
--- a/ui/css/module/_footnotes.scss
+++ b/ui/css/module/_footnotes.scss
@@ -12,6 +12,45 @@
   }
 }
 
+.node-section {
+  .node-footnote-content {
+    display: block;
+    width: 95%;
+    margin-left: auto;
+    margin-right: auto;
+  }
+
+  .node-footnote {
+    font-size: 0.9em;
+    display: block;
+    margin-top: 1rem;
+    margin-bottom: 2rem;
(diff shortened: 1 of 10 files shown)


rememberlenny:
Finished update on requirements from issue.

ui/css/module/_footnotes.scss
```diff
@@ -5,13 +5,77 @@
   padding: 2px 4px;
   font-size: 0.6em;
   margin-left: .3em;
+  font-style: normal;
```
cmc333333: Rather than undoing the `bold`, should we be including `font-san-serif()`?

ui/css/module/_footnotes.scss
```diff
@@ -5,13 +5,77 @@
...
 
   &.active {
-    background: $color-cool-blue;
+    border-color: $color-primary-darker;
+    background: $color-primary-darker;
     color: $color-white;
   }
 }
 
+.node-section {
```
cmc333333: This document has sections throughout, but I don't think it'd be wise to use that as a differentiator between inline footnotes and others. What do you think about adding a new class for inline footnotes instead?

Relatedly, we may need to use a separate component; I _think_ that right now we're including multiple elements with the same `id` due to the component reuse.

ui/css/module/_footnotes.scss
```diff
@@ -5,13 +5,77 @@
...
+    line-height: 1.5em;
+    text-align: right;
+    padding-right: .5em;
+  }
+
+  .footnote-text {
+    @extend .col;
+    @extend .m0;
+    @extend .col-11;
+    width: calc(100% - 3em);
```
cmc333333: Very smart. Let's remove the `col-11` extension, as this replaces it.

rememberlenny:
Sounds good!

I noticed that Marker was empty in the data examples, but type emblem has the footnote index.

Is that something to be concerned with?

cmc333333:
Try re-importing the XML document -- it could be that it's changed since your last import; it should definitely have the `marker` field populated for footnotes.

ui/css/module/_footnotes.scss
rememberlenny: @cmc333333 I'm keeping the `col-11` because it acts as a fallback when the `calc` doesnt work.

cmc333333: Ah, good call!

Issue #623, opened 2017-11-06
Foonotes in tables (we're combining them at the bottom of the table)


yowill:

cmc333333:
Latest design is noted in [this comment](https://github.com/18F/omb-eregs/issues/564#issuecomment-342892641)

cmc333333:
This still needs styling love.

cmc333333:
We also need a different ux and graphic design for the footnote citations. Per slack, clicking the footnote citation should scroll to the relevant footnote in the table footer.

PR #643, opened 2017-11-09
Convert node fns to OO, add table footnotes

cmc333333:
As part of #623, this adds footnotes to the bottom of tables. It's not styled well, but should get us started. It also converts some library functions into methods for ease of use.

Looks like
screen shot 2017-11-09 at 6 52 35 pm

### DIFF
--- a/ui/__tests__/components/node-renderers/policy.test.js
+++ b/ui/__tests__/components/node-renderers/policy.test.js
@@ -6,6 +6,7 @@ import {
   itIncludesTheIdentifier,
   itRendersChildNodes,
 } from '../../test-utils/node-renderers';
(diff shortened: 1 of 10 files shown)


tadhg-ohiggins:
Does this duplicate the bottom-of-document footnotes that are also in the table, or are they at the bottom of the table instead of at the bottom of the document?

cmc333333:
Thanks @tadhg-ohiggins and @rememberlenny.

@tadhg-ohiggins the footnotes are duplicated. This feels like proper behavior, but I could be dissuaded. It's also easier to implement ;)

@rememberlenny let's chat about `DocumentNode`.

PR #644, opened 2017-11-10
Use white-space: nowrap in the footnote citation.

cmc333333:
This prevents the citation from breaking across lines.

## Before
footnote-split

## After
footnote-fixed


cmc333333:
A css solution is definitely more flexible; I'll refactor to use that style.

cmc333333:
Since this has changed quite a bit, @rememberlenny: would you mind re-reviewing?

cmc333333:
@rememberlenny says this is still good. I'll rebase and merge.

PR #645, opened 2017-11-10
Add simple test for KaTeX math component

tadhg-ohiggins:
If a render fails
When nobody can hear it
Now we might find out.

Add a basic test for the KaTeX math component.

### DIFF
--- a/ui/__tests__/components/node-renderers/math.test.js
+++ b/ui/__tests__/components/node-renderers/math.test.js
@@ -12,12 +12,12 @@ describe('<TeXMath />', () => {
   it('generates something math-like', () => {
     const docNode = {
       identifier: 'aaa_1__bbb_2__ccc_3_math',
-      text: '\\frac{10}{11}',
+      text: '\\frac{12345654321}{10987678901}',
     };
     const result = mount(<TeXMath docNode={docNode} />);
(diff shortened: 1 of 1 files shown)


ui/__tests__/components/node-renderers/math.test.js
```diff
@@ -0,0 +1,23 @@
...
+  itIncludesTheIdentifier(TeXMath, { text: 'abcd' });
+  it('generates something math-like', () => {
+    const docNode = {
+      identifier: 'aaa_1__bbb_2__ccc_3_math',
+      text: '\\frac{10}{11}',
+    };
+    const result = mount(<TeXMath docNode={docNode} />);
+    expect(result.render().find('span.katex')).toHaveLength(1);
+    expect(result.html()).toContain('10');
+    expect(result.html()).toContain('11');
```
cmc333333: I see that these numbers are coming from line 15, but I wonder if there's not a risk that they're also part of the generated css classes? Would it make sense to use `result.text()` instead? Or maybe less "common" input?

tadhg-ohiggins: You're right, I should change these to rarer numbers. (I'll also test the text option.)

PR #647, opened 2017-11-13
Update list item styling

cmc333333:
This should resolve the elements on #600 by tweaking the list item styles and replacing the bullet marker in the source XML.

## Bullets
screen shot 2017-11-13 at 1 31 49 pm

## Small screen
screen shot 2017-11-13 at 1 31 32 pm

## Large screen
screen shot 2017-11-13 at 1 31 17 pm

### DIFF
--- a/api/example_docs/m_16_19_1.xml
+++ b/api/example_docs/m_16_19_1.xml
@@ -261,7 +261,7 @@
             community
           </footnote>
           <list>
-            <listitem marker="•">
+            <listitem marker="●">
               <para>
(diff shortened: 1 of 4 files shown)


cmc333333:
Thanks @micahtaylor and @rememberlenny. I'll rebase and merge.

@rememberlenny I hoped we could avoid too much inspection of the data when determining which styles to use. If we make all `list-item-marker`s bigger, for example, we'll harm the numbered and lettered lists.

PR #649, opened 2017-11-13
Fix footnote accessibility issues

### LINKED ISSUE #633: Ensure pop-up footnotes are accessible (depends on #620)
Ensure that the footnote interaction is accessible, particularly to a screen reader. One potential solution is described in [this comment](https://github.com/18F/omb-eregs/issues/566#issuecomment-338236395), but that doesn't account for the popup footnotes.

toolness:
Fixes #633.

Notes:

* `role="alertdialog"` for the pop-up footnote might not be ideal, but it's the closest thing I could find that seemed to work on screen readers.
* I replaced the footnote `<Link>` with an `<a>` because I couldn't figure out how to get a reference to the DOM node for the link otherwise (when I put a `ref` on the `<Link>` it was called back with `null` as its first argument). I thought it would be okay since we're not actually using this link for any next.js-specific functionality, e.g. next.js isn't intercepting clicks and diverting them to `pushState` here or anything.
* I considered adding an `aria-haspopup="true"` to the footnote link, so screen-reader users would know what would happen when they activate the link, but apparently [`aria-haspopup` should only be used for menus and sub-menus](https://stackoverflow.com/a/20402871). Huh.

To do:

- [ ] Add tests.
- [x] Convert `<Link>` to a React class so it can support `ref`.
- [x] Use `<Link>` instead of `<a>`.

### DIFF
--- a/ui/components/content-renderers/footnote-citation.js
+++ b/ui/components/content-renderers/footnote-citation.js
@@ -1,6 +1,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 
+import Link from '../link';
 import { renderContent } from '../../util/render-node';
 
 import Footnote from '../node-renderers/footnote';
@@ -55,14 +56,14 @@ export default class FootnoteCitation extends React.Component {
     const href = `#${this.props.content.footnote_node.identifier}`;
     const link = (
       <sup>
-        <a
+        <Link
           className={klass}
           onClick={this.handleCitationClick}
           href={href}
-          ref={(el) => { this.citationLink = el; }}
+          ref={(component) => { this.citationLink = component; }}
         >
           Footnote { this.props.content.text }
-        </a>
+        </Link>
       </sup>
     );
     if (expanded) {
--- a/ui/components/content-renderers/footnote-citation.js
+++ b/ui/components/content-renderers/footnote-citation.js
@@ -4,7 +4,6 @@ import React from 'react';
 import { renderContent } from '../../util/render-node';
 
 import Footnote from '../node-renderers/footnote';
-import Link from '../link';
 
 const propTypes = {
   content: PropTypes.shape({
@@ -26,6 +25,14 @@ export default class FootnoteCitation extends React.Component {
     this.handleCitationClick = this.handleCitationClick.bind(this);
   }
 
+  componentDidUpdate(prevProps, prevState) {
+    if (!prevState.expanded && this.state.expanded && this.citationWrapper) {
+      // Focus the footnote so screen-readers announce it, and to ensure
+      // that keyboard focus moves to it.
+      this.citationWrapper.focus();
+    }
+  }
+
   handleCitationClick(e) {
     e.preventDefault();
     this.setState({ expanded: !this.state.expanded });
@@ -40,14 +47,6 @@ export default class FootnoteCitation extends React.Component {
     this.setState({ expanded: false });
   }
 
-  componentDidUpdate(prevProps, prevState) {
-    if (!prevState.expanded && this.state.expanded && this.citationWrapper) {
-      // Focus the footnote so screen-readers announce it, and to ensure
-      // that keyboard focus moves to it.
-      this.citationWrapper.focus();
-    }
-  }
-
   render() {
     let footnoteContent;
     const footnote = this.props.content.footnote_node;
@@ -56,17 +55,19 @@ export default class FootnoteCitation extends React.Component {
     const href = `#${this.props.content.footnote_node.identifier}`;
     const link = (
       <sup>
-        <a className={klass} onClick={this.handleCitationClick} href={href}
-          ref={(el) => { this.citationLink = el; }}>
+        <a
+          className={klass}
+          onClick={this.handleCitationClick}
+          href={href}
+          ref={(el) => { this.citationLink = el; }}
+        >
           Footnote { this.props.content.text }
         </a>
       </sup>
     );
     if (expanded) {
       footnoteContent = (
-        <span className="citation-wrapper" role="alertdialog"
-          aria-label={"Footnote " + footnote.marker} tabIndex="-1"
-          ref={(el) => { this.citationWrapper = el; }}>
+        <span className="citation-wrapper" role="alertdialog" aria-label={`Footnote ${footnote.marker}`} tabIndex="-1" ref={(el) => { this.citationWrapper = el; }}>
           <Footnote docNode={footnote}>
             { renderContent(footnote.content) }
           </Footnote>
--- a/ui/components/content-renderers/footnote-citation.js
+++ b/ui/components/content-renderers/footnote-citation.js
@@ -18,7 +18,7 @@ export class FootnoteCitation extends React.Component {
     this.footnoteIdentifier = props.content.footnote_node.identifier;
   }
 
-  componentDidUpdate(prevProps, prevState) {
+  componentDidUpdate(prevProps) {
     if (!prevProps.expanded && this.props.expanded && this.citationWrapper) {
       // Focus the footnote so screen-readers announce it, and to ensure
       // that keyboard focus moves to it.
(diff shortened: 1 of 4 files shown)


ui/css/module/_footnotes.scss
```diff
@@ -31,6 +31,12 @@
     padding-bottom: 0.2em;
   }
 
+  /* It's ok to add outline: none here for accessibility purposes.
+   * For more details, see: https://youtu.be/ifW_oy9hajU */
+  .citation-wrapper[tabindex="-1"] {
+    outline: none;
+  }
+
```
toolness: Technically I could've added this CSS to the `.citation-wrapper` rule above, but I thought adding the `[tabindex="-1"]` made it more explicit that the `outline: none` is intended _because_ the element itself is non-interactive, and we're focusing on it for reasons explained in the aforementioned [video](https://youtu.be/ifW_oy9hajU).

ui/components/content-renderers/footnote-citation.js
```diff
@@ -44,14 +55,19 @@ export default class FootnoteCitation extends React.Component {
     const href = `#${this.props.content.footnote_node.identifier}`;
     const link = (
       <sup>
-        <Link className={klass} onClick={this.handleCitationClick} href={href}>
+        <a
```
cmc333333: Is the problem with `Link` that setting its `ref` won't fire (since it's not an html tag)?

toolness: So the weird thing is that according to the React docs (well, the React v16 docs) `ref` _is_ fired for components, it just passes the component instance rather than the DOM element. But what's strange is that for `Link`, the callback _is_ triggered but it's being passed neither a component instance nor a DOM element--it's just getting passed `null`.

That said, though, I do have an additional concern, which is that i'm uncertain whether a `Link` is even appropriate here.  My (admittedly novice) understanding of next.js's `Link` component is that it's meant for links that navigate to other pages on the site, rather than other `id`s on the current page; `Link` is required over `<a>` in such cases so that next can override the `onClick` handler to use `pushState` instead of the browser performing its own navigation.  However, in the case of links that navigate to different `id`s, the browser's default navigation _does_ work fine, so it seems to me like a `Link` shouldn't be needed.  But aside from that, there's also the fact that we're providing our _own_ `onClick()` behavior here, which renders `Link`'s `onClick` behavior particularly irrelevant. However, I could be mistaken given that I have very limited experience with next.js.

toolness:
@cmc333333 I don't think aria tags are the solution here, and `autofocus` only applies to first page loads.  For one thing, I'm not just fixing accessibility issues with screen-readers and other ATs (which is what ARIA is for)--I'm also fixing accessibility issues with basic keyboard navigation, for which ARIA is not a solution.

Based on the [Google Udacity Web Accessibility class](https://www.udacity.com/course/web-accessibility--ud891), manually managing input focus is a common task when making custom HTML5 widgets accessible.  In one case, we're literally removing the currently-focused element from the DOM, and browsers don't have well-defined behavior on what's supposed to happen then; there is no kind of HTML tag for "when a specific element in the DOM disappears, re-focus on this other one", so we need to explicitly tell the browser what to do.  (In a lot of ways, our footnote widget is similar to a modal dialog, and examining [WAI Aria's modal dialog example](https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html) reveals how much manual focusing is required in that particular case.)

Finally, [React's `ref` documentation](https://reactjs.org/docs/refs-and-the-dom.html) explicitly calls out "managing focus" as one of the good use cases for `ref`s, so it seems like the right React-based solution to accomplish what needs to be done.

ui/components/content-renderers/footnote-citation.js
cmc333333: Indeed, we don't need the extra functionality provided by Next's `Link`, but we've [tried](https://github.com/18F/omb-eregs/pull/547) to standardize on our wrapper over different types of anchors. If we can make this work with (our) `Link`, we should, if only to keep the code base consistent.

cmc333333:
Sounds good @toolness -- just wanted to press on you in case there was a cleaner alternative.

ui/components/content-renderers/footnote-citation.js
toolness: D'oh! LOL, I just saw the name `Link` and assumed it was next.js's... I even looked at next.js's source to make sure `Link` was a full `React.Component` class and not a function, because `ref`s don't work with components that are functions!

Looking at our own `Link`, though, it appears that it _is_ a function, and that is most likely why `ref` isn't working.  Should I convert our `Link` into a standard `React.Component` subclass in this PR?

cmc333333: Sounds good -- please add a comment about why when you do. I bet the linter will complain; adding such a comment when you disable the rule would be perfect.

ui/components/link.js
```diff
@@ -3,24 +3,35 @@ import React from 'react';
...
+        href,
+      };
+      if (href.search(/^https|http|^\/\//) === 0) {
+        props.target = '_blank';
+        props.rel = 'noopener noreferrer';
+        props['aria-label'] = 'Link opens in a new window';
+      }
+      return <a {...props} ref={(el) => { this.el = el; }}>{children}</a>;
+    }
+    throw new Error('Link created without an href or route');
```
toolness: Interesting... The function-based version of this component passed eslint fine, but when converted to a `render()` method eslint complained about the following:

```
/usr/src/app/ui/components/link.js
  13:3  error  Expected to return a value at the end of method 'render'  consistent-return
```

I chose to make it throw an error because this seems like that kind of situation, but alternatively I could change it to log a warning and return `null`, or return a React element that displays a warning on the webpage, or something else. Or, if that branch really _is_ a valid use case, we could just return `null` and not log anything.

Any preferences?

ui/components/link.js
```diff
@@ -3,24 +3,35 @@ import React from 'react';
...
+      );
+    } else if (href) {
+      const props = {
+        ...otherProps,
+        href,
+      };
+      if (href.search(/^https|http|^\/\//) === 0) {
+        props.target = '_blank';
+        props.rel = 'noopener noreferrer';
+        props['aria-label'] = 'Link opens in a new window';
```
toolness: This is pre-existing code but it's bad code--see #655. I could quickly fix it in this PR by just removing the line if you want, let me know!

cmc333333: Sounds good to me; I commented a bit more in that ticket.

ui/components/link.js
```diff
@@ -3,24 +3,35 @@ import React from 'react';
...
+      }
+      return <a {...props} ref={(el) => { this.el = el; }}>{children}</a>;
```
cmc333333: Would it be clearer to move the `ref` definition up to line ~26, where we're creating `props`?

ui/components/link.js
cmc333333: Nice. I'd bet the linter not catching this before was a bug. I like the exception approach. We could instead remove the final guard (`else if (href)`), but that'd generate some silent fail cases. I think the same is true for `null`.

ui/components/link.js
toolness: Oh sure, I can do that. It's funny, in some ways I think of `ref` as a sort of "meta-prop" or something, so it felt OK specifying it in the JSX rather than in the `props` dict.  But I think it could also make the code more readable by not making the JSX too lengthy.

toolness:
Yeah, I think that the reason the linter didn't catch it before was because with function-based components, the linter doesn't have an easy way to determine whether it's actually a function-based component or just a regular function, but I'm not sure.

Anyways, cool, I think I will go ahead and pre-emptively merge this PR with your redux PR before writing the tests, since your PR seems to change tests enough that I'd rather build atop them than write my tests separately and then deal with merge conflicts.

toolness:
@cmc333333 actually do you mind if we merge this now and I submit tests in a separate PR?  There's been some churn in this area of the code and it might actually take me a little bit to write proper tests for this, as I'm pretty rusty on jest and DOM testing, so I'd rather get the implementation in sooner rather than later to avoid further merge conflicts... It's ok if not though.

Issue #648, opened 2017-11-13
Discuss site-wide color palette


cmc333333:
There's a bit of a mismatch between the pallet described in [this comment](https://github.com/18F/omb-eregs/issues/527#issuecomment-340073841) and our current USWDS color scheme. Let's select a single set of variables and use them site wide.

Some additional requirements:
* [x] accessible
* [ ] relatively easy to rebrand (meaning var names shouldn't be tied to color)

cmc333333:
Take aways from meeting:
1. the pallet @rtwell should already include accessible combinations
2. we'll delete all of the uswds color vars
3. we'd still like a strategy for swapping colors/themes out. Ryan's contacting @donjo to get his strategy
4. we'd similarly like to have a consistent unit for spacing, font-sizes, etc. It may make sense to split that out into a separate issue.

yowill:
Done!

PR #650, opened 2017-11-14
Style bottom policy footnotes

cmc333333:
As part of #624, this

* Changes color of border
* Changes indent for marker
* Shrinks font size a bit
* Ensures long URLs are wrapped
* Makes marker text bold
* Attempts to line up marker text with footnote text

It also moves the `cite` tag to be more semantic and prevent italic text from appearing in our inline footnotes.

It also adds a new color variable since the requested color wasn't present. I've opened #648 to try to resolve that more permanently.


cmc333333:
Thanks @micahtaylor -- I'll rebase and merge.

PR #653, opened 2017-11-14
Ensure only one footnote is opened at a time

### LINKED ISSUE #621: Only allow one footnote to be opened at a time (depends on #620)

cmc333333:
This resolves #621 by pushing the which-footnote-is-expanded logic into redux. It's overkill for our single bit of state, but should bootstrap this style of state management.

CC @tadhg-ohiggins for visibility

### DIFF
--- a/ui/components/app-wrapper.js
+++ b/ui/components/app-wrapper.js
@@ -1,8 +1,10 @@
 import Router from 'next/router';
+import withRedux from 'next-redux-wrapper';
 import NProgress from 'nprogress';
 import PropTypes from 'prop-types';
 import React from 'react';
 
+import createStore from '../store';
 import ErrorView from './error';
 import HeaderFooter from './header-footer';
 
@@ -45,5 +47,8 @@ export default function wrapPage(Page, dataFn, headerFooterParams) {
       return { err };
     }
   };
-  return WrappedPage;
+  return withRedux({
+    createStore,
+    debug: process.env.NODE_ENV === 'development',
+  })(WrappedPage);
 }
--- a/ui/components/content-renderers/footnote-citation.js
+++ b/ui/components/content-renderers/footnote-citation.js
@@ -4,7 +4,7 @@ import { bindActionCreators } from 'redux';
 import { connect } from 'react-redux';
 
 import { renderContent } from '../../util/render-node';
-import { openFootnote } from '../../store/actions';
+import { closeFootnote, openFootnote } from '../../store/actions';
 
 import Footnote from '../node-renderers/footnote';
 import Link from '../link';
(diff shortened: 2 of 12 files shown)


ui/components/app-wrapper.js
```diff
@@ -45,5 +47,8 @@ export default function wrapPage(Page, dataFn, headerFooterParams) {
       return { err };
     }
   };
-  return WrappedPage;
+  return withRedux({
+    createStore,
+    debug: process.env.NODE_ENV === 'development',
```
toolness: Something I've noticed with `NODE_ENV` is that folks seem to usually compare it to `production` rather than checking to see if it's `development`, meaning that this conditional might instead be expressed as `process.env.NODE_ENV !== 'production'`. However, if our docker setup is correctly setting `NODE_ENV` to `development` then I guess it doesn't matter!

ui/components/content-renderers/footnote-citation.js
```diff
@@ -1,45 +1,37 @@
...
-    };
     this.shrinkFootnote = this.shrinkFootnote.bind(this);
     this.handleCitationClick = this.handleCitationClick.bind(this);
+    this.footnoteIdentifier = props.content.footnote_node.identifier;
   }
 
   handleCitationClick(e) {
     e.preventDefault();
-    this.setState({ expanded: !this.state.expanded });
+    this.props.openFootnote(this.footnoteIdentifier);
```
toolness: Does this (along with the corresponding logic in the reducer) technically change the behavior of the citation link from _toggling_ the footnote to specifically ensuring that it's open, but not closing it if it's already visible?

ui/store/initial-state.js
```diff
@@ -0,0 +1 @@
+export default { openedFootnote: null };
```
toolness: Do we want to add a comment explaining why some elements of the state, such as the parts of it reflected in the URL, aren't currently represented here?  Just in case folks come here expecting to see a bunch more stuff?

ui/components/app-wrapper.js
cmc333333: It's funny you mention that -- I originally had `!== 'production'`, but realized we actually have three environments: production, development, test. Jest sets the latter by default (though I think we make it explicit, too) to trigger a slightly different babel config. When I set to `!== 'production'`, I saw some rather noisy tests, so I opted out. What do you think?

ui/components/content-renderers/footnote-citation.js
cmc333333: Excellent catch; I did _not_ intend to change that logic. Will update.

ui/store/initial-state.js
cmc333333: Can do.

cmc333333:
Thanks @toolness. I'll push up a fix for the toggle, rebase and merge.

ui/components/app-wrapper.js
toolness: Oh, interesting... I am used to testing frameworks that suppress stdout unless the test fails, which I like because that makes failing tests easier to debug without making successful tests super noisy.  If there's no way to enable that in jest, though, I think it's fine to leave your code as-is!

PR #654, opened 2017-11-14
Alter footnote citations in a table

cmc333333:
As part of #623, this renders footnote citations within tables differently than the rest of the document by:
1. allowing individual node renderers to override how their content is rendered
2. adding a custom renderer for footnote citations for use within tables.

The big change about the former is a change in the way we render content (specifically, inline text, citations, etc.) Rather than having `renderNode` pass the rendered content along, this pushes the render logic into the components:

```jsx
<Footnote docNode={docNode}>
  { renderContents(docNode.content) }
</Footnote>

function Footnote({ children, docNode }) {
  return <div id={docNode.identifier}>{ children }</div>;
}
```

becomes:

```jsx
<Footnote docNode={docNode} />

function Footnote({ docNode }) {
  return <div id={docNode.identifier}>{ renderContents(docNode.content) }</div>
}
```

### DIFF
--- a/ui/__tests__/components/content-renderers/footnote-citation-in-table.test.js
+++ b/ui/__tests__/components/content-renderers/footnote-citation-in-table.test.js
@@ -1,0 +1,21 @@
+import { shallow } from 'enzyme';
+import React from 'react';
+
+import FootnoteCitationInTable from '../../../components/content-renderers/footnote-citation-in-table';
+
+describe('<FootnoteCitationInTable />', () => {
+  const content = {
+    footnote_node: { identifier: 'some__identifier' },
+    text: 'some text, like "1"',
+  };
+  const result = shallow(<FootnoteCitationInTable content={content} />);
+  const link = result.find('Link');
+
+  it('includes the correct link', () => {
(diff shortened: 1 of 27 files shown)


cmc333333:
I'll rebasing to resolve file conflict.

cmc333333:
Thanks @tadhg-ohiggins! The alternative I was considering (and I'm hoping it's your "a") was to have the `FootnoteCitation` inline component look to see if it was within a table. If so, it'd generate one type of link and if not, our normal version. That can definitely work (it's used in #662), but yeah, it pushes the testing down the tree. It's more flexible in the sense that we can check "are _any_ of my ancestors a table?" rather than the table explicitly defining the content renderers at only one level, but less flexible in that the "leaf" Components (e.g. FootnoteCitations) have to be aware of the different node types, which feels fragile. I'm pretty sure we could pass the content-renderer configuration down between nodes via React's `context` to resolve the any-ancestor problem, but that's even more complicated...

In any event, we can start with this and switch to a different style if this proves annoying. I'll rebase and merge.

PR #661, opened 2017-11-15
[WIP] Attempt to link to policy detail

### LINKED ISSUE #556: Policy detail: Incoming link
Per Slack [conversation](https://gsa-tts.slack.com/archives/C3HKL8MS7/p1507754840000293?thread_ts=1507754115.000618&cid=C3HKL8MS7), we want the circles on the policy listing page to link to their respective policy detail pages.

toolness:
One day, this might fix #556, but for now

Interesting things that have happened so far:

* ~~I originally tried adding a property to the policy API response called `docnode_count`, but this caused next.js to fail while rendering the homepage, claiming that it was being asked to JSONify a self-referential object. I have no idea why it felt that way, but renaming the field to `boop` somehow fixed things.~~ **Update:** CM explained the error and I've filed #665 separately.

* ~~Even once I called the field `boop` and it showed up in unit tests, it still wasn't showing up in API responses.  Eventually I did `docker-compose up dev-api` and intentionally broke the API by adding a nonsensical field, and when I fixed it and the server restarted itself, `boop` was in the API responses. So I guess something weird was being cached somewhere.~~ Apparently the app has a LRU cache for API results both in the browser and on the server, so I need to restart the server for stuff to take effect properly.

## To do

- [x] Figure out how to rename `boop` to `has_docnode` and change the data type to a boolean instead of an int.
- [x] Add tests for the front-end.
- [x] Add a test to the back-end that verifies that `has_docnode` is `true` when it should be.
- [ ] If the OMB policy ID doesn't exist for the document, we need to fall back to its slug.

### DIFF
--- a/api/reqs/serializers.py
+++ b/api/reqs/serializers.py
@@ -24,7 +24,7 @@ class GroupWithAgenciesSerializer(serializers.ModelSerializer):
 
 
 class PolicySerializer(serializers.ModelSerializer):
-    boop = serializers.IntegerField(read_only=True)
+    boop = serializers.BooleanField(read_only=True)
     total_reqs = serializers.IntegerField(read_only=True)
     relevant_reqs = serializers.IntegerField(read_only=True)
     title_with_number = serializers.CharField(read_only=True)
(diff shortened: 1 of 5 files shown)


toolness:
@cmc333333 do you mind if I do a separate PR for the "If the OMB policy ID doesn't exist for the document, we need to fall back to its slug" thing?  Other than that, I think this PR is good to go...

cmc333333:
Sounds like a good plan @toolness.

PR #662, opened 2017-11-15
Footnote citation links in maths!

### LINKED ISSUE #625: Footnotes in math

cmc333333:
This hacks around some missing functionality in KaTeX (the ability to include
links) by inserting markdown syntax for a link and replacing it with an `a`
tag _after_ running it through the KaTeX transform.

This should get simpler once KhanAcademy/KaTeX#923 is merged, but until then, I think it'll do.

Note that the anchor this links to won't be functional until #654 is merged.

This should resolve #625 

## Looks like
screen shot 2017-11-15 at 5 53 35 pm

### DIFF
--- a/api/example_docs/m_16_19_1.xml
+++ b/api/example_docs/m_16_19_1.xml
@@ -708,6 +708,7 @@
                           \textit{Total Active}
                           \\
                           \textit{Rack Count}
+                          <footnote_citation>28</footnote_citation>
                         \end{matrix}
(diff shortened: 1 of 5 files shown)


cmc333333:
Thanks @tadhg-ohiggins, and you got it! I think that once the KaTeX PR is merged, we'll be able to squish 5-8 into a single step.

I'll rebase and merge.

PR #669, opened 2017-11-16
Add styling class to make list bullets larger

rememberlenny:
# What

Based on design, make list markers larger without affecting the numbers.

screen shot 2017-11-16 at 4 24 04 pm
screen shot 2017-11-16 at 4 24 07 pm

### DIFF
--- a/ui/components/node-renderers/list-item.js
+++ b/ui/components/node-renderers/list-item.js
@@ -4,8 +4,15 @@ import React from 'react';
 import renderNode from '../../util/render-node';
 
 export default function ListItem({ docNode }) {
+  let markerType;
+  if (docNode.marker == "•") {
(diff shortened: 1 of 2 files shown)


cmc333333:
I think doing this sort of logic in the frontend is going to be pretty error prone. The markers for lists will vary depending on our document; sometimes they'll have a parenthesis, sometimes they'll be letters, other times they'll be triangles. If it's really critical that we distinguish between numbered lists and bulleted lists, we'll want to change that at the structural level (i.e. create different node types).

Given that the "black circle" unicode is larger than what we had before (and passed visual design review), I think we can skip this PR.

rememberlenny:
Sounds good!

rememberlenny:
Sounds good!

PR #672, opened 2017-11-17
Remove focus outline.

cmc333333:
This style was accidentally lost in some class renaming and PR splitting.

## Before

screen shot 2017-11-16 at 9 48 09 pm

## After

screen shot 2017-11-16 at 9 48 21 pm


cmc333333:
Thanks @rememberlenny

PR #675, opened 2017-11-17
Re-add vertical padding on nav bar

rememberlenny:
# What

Add vertical margin on navbar after removing Chrome default settings.


cmc333333:
I'm going to go ahead and merge this as we'll want it for demo.

PR #677, opened 2017-11-17
Add initial importer, model, display for external links

cmc333333:
We'd like to allow editors to select certain chunks of text as external links, which means we need to have that data model ready for them. This PR creates that (simple) model and defines how it serializes. It then uses that within the import scripts to include several links in M-16-19. Finally, it creates a renderer on the UI side to generate the links and a simple CSS rule to include them in print output.

This doesn't include a full css implementation (notably, it's missing underline and icon). Thanks for pairing with me on this @tadhg-ohiggins !

## Looks like
screen shot 2017-11-17 at 11 28 08 am
screen shot 2017-11-17 at 11 29 31 am


toolness:
The front-end part of this is splendid. I'm still getting a feel for the back-end document models, so I can't say I understand 100% of what's going on there, but what I do understand seems reasonable to me. Excellent work!

cmc333333:
Thanks @tadhg-ohiggins!

PR #679, opened 2017-11-17
Minor style fixes from footnotes and document

rememberlenny:
# What

Correcting some style artifacts.

screen shot 2017-11-17 at 12 41 28 pm
screen shot 2017-11-17 at 12 38 21 pm
screen shot 2017-11-17 at 12 44 29 pm
screen shot 2017-11-17 at 12 53 45 pm

### DIFF
--- a/ui/css/layout/_markers.scss
+++ b/ui/css/layout/_markers.scss
@@ -3,15 +3,23 @@ $marker-whitespace-xs: 4px;
 $marker-width-md: $marker-width-xs + 8;
 $marker-whitespace-md: $marker-whitespace-xs + 4;
 
-@mixin paragraph-marker() {
+@mixin paragraph-marker($width: normal) {
   @extend .col;
   @extend .right-align;
-  width: $marker-width-xs;
+  @if $width == 'auto' {
+    width: $width;
+  } @else {
+    width: $marker-width-xs;
+  }
   margin-right: $marker-whitespace-xs;
   display: inline-block;
 
   @media #{$breakpoint-md} {
-    width: $marker-width-md;
+    @if $width == 'auto' {
+      width: $width;
+    } @else {
+      width: $marker-width-md;
+    }
     margin-right: $marker-whitespace-md;
   }
 }
--- a/ui/css/module/_listitem.scss
+++ b/ui/css/module/_listitem.scss
@@ -10,6 +10,10 @@
   .list-item-text {
     @include paragraph-with-marker();
   }
+
+  .node-paragraph {
+    margin-bottom: 0;
+  }
 }
 
 .node-paragraph {
--- a/ui/css/module/_listitem.scss
+++ b/ui/css/module/_listitem.scss
@@ -11,3 +11,14 @@
     @include paragraph-with-marker();
   }
 }
+
+.node-paragraph {
+  .list-item-marker,
+  .list-item-text {
+    line-height: 1.69rem;
+  }
+
+  .list-item-marker {
+    font-size: .9rem;
+  }
+}
--- a/ui/css/module/_listitem.scss
+++ b/ui/css/module/_listitem.scss
@@ -11,7 +11,11 @@
     @include paragraph-with-marker();
   }
 
-  .node-paragraph {
+  .node-paragraph{
+    margin-bottom: 8px;
+  }
+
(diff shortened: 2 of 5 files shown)


ui/css/module/_listitem.scss
```diff
@@ -10,4 +10,19 @@
   .list-item-text {
     @include paragraph-with-marker();
   }
+
+  .node-paragraph {
+    margin-bottom: 0;
+  }
```
cmc333333: This feels a bit risky as a list item could contain multiple paragraphs. Is the goal to modify just the last one?

ui/css/module/_listitem.scss
```diff
@@ -10,4 +10,19 @@
...
+}
+
+.node-paragraph {
```
cmc333333: I don't understand the intent here -- why do we want different line heights for lists inside paragraphs?

ui/css/layout/_markers.scss
```diff
@@ -3,15 +3,23 @@ $marker-whitespace-xs: 4px;
 $marker-width-md: $marker-width-xs + 8;
 $marker-whitespace-md: $marker-whitespace-xs + 4;
 
-@mixin paragraph-marker() {
+@mixin paragraph-marker($width: normal) {
   @extend .col;
   @extend .right-align;
-  width: $marker-width-xs;
+  @if $width == 'auto' {
+    width: $width;
```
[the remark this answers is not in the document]
rememberlenny: The design for the table footnotes are flush against the left column. If we manually set the width, then we are at odds to the number of characters in the marker. For example, setting a width for `1.` or `2.` is very different than `1203.` or `201`. By making it auto, we allow the marker to fit to the left, while also keeping a consistent distance from the following list item text.

ui/css/module/_listitem.scss
rememberlenny: Currently the list item has a `margin-bottom: 8px` based on the design. With the `.node-paragraph` inside, it is making the list item's `margin-bottom` closer to `24px`. The goal was to correct this. Given the multiple paragraph case, I'll make a fix.

ui/css/module/_listitem.scss
rememberlenny: If the line-height of the `list-item-marker` and the content inside `list-item-text` are different, then we end up with the unaligned baseline problem. If the `node-paragraph` has a larger/smaller line-height, the `marker` gets thrown off. Do you have any thoughts on a good way to resolve this?

cmc333333: The basic problem is that we'd like to have each of these node types be compose-able -- with some exception, any type of node can be inside any other type. Given that, any time we're testing for nested css classes, we're in a bad situation.

Would a potential solution be to remove top-margins and _only_ add bottom margins? We probably shouldn't have padding as that will be additive rather than collapsing. Does that make sense?

rememberlenny: I agree about the nested nodes.

I think Im missing something around the top-margin changes.

ui/css/layout/_markers.scss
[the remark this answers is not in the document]
cmc333333: I think we're seeing the mockup differently -- I see the mockup as aligning the footnotes to the right (i.e. "growing" to the left), which would be consistent with the footnotes and markers in the rest of the document.

Regarding fixed width: I agree that there's a significant risk with large markers -- we can only accommodate ~5-character markers currently. The tradeoff is that defining a fixed width ensures the paragraphs are all aligned; with `auto` we get this:
screen shot 2017-11-17 at 3 39 10 pm

ui/css/module/_listitem.scss
cmc333333: Apologies around top-margin -- I was poorly multi-tasking and thought we were talking about paragraphs aligning too low due to a top margin. Let's try a better question: should the line-heights of markers and paragraphs be different?

cmc333333:
Given the new information around the table footnotes (that they were meant to be right-aligned), seems like we should boot that. I'm torn on whether adding the nested style rules is worth adding for now or will bite us soon enough that we should delay. What do you think, @rememberlenny?

cmc333333:
#712 implemented the non-contentious parts of this, so I'm going to close.

Issue #646, opened 2017-11-13
Document the "document" model


cmc333333:
Find a good place to store this documentation. Maybe in the API? Maybe duplicated in frontend's `DocumentNode`?

---

* identifier is a unique (with a single document) string that's composed of `__`-divided components
* node_type is a string that gives some semantic meaning to each node. https://github.com/18F/omb-eregs/wiki/Content-Types describes most of these, but we can expect to add more in the future
* type_emblem is a way to distinguish children, to differentiate paragraph a and paragraph b. Parent identifier + node_type + type_emblem == identifier of this node.
* text is the plain text for a node (sans any markers). We generally don't want to access this as it doesn't deal with inline elements
* marker contains things like "a)" or "Section 22"; not essential to understanding a list item, footnote, etc. but essential to correct display

Then there are the two compound fields:
* children - a recursive array of nodes
* content - an array of inline elements (links, plain text, footnote citations, that sort of thing)

PR #687, opened 2017-11-20
Explicitly define most of DocumentNode's interface.

cmc333333:
Previously, we had been pretty loose about what a DocumentNode's fields were,
effectively referring to the API's fields. This makes us a bit more explicit
(and follows JS conventions around camel case) in using DocumentNode objects
everywhere, rather than an object with a particular shape.

This should also provide a bit more documentation, resolving #646.

As a separate benefit, this unifies a few helper functions around constructing DocumentNodes.


cmc333333:
Thanks @toolness! I'll rebase and merge.

PR #688, opened 2017-11-20
Prefer .toHaveLength() over .length).toEqual().

cmc333333:
eslint-plugin-jest added a rule around preferring toHaveLength() in 21.3.0;
let's follow it.


cmc333333:
Thanks @tadhg-ohiggins! I'll rebase and merge.

Issue #678, opened 2017-11-17
Link to policy detail if it has a slug but not an OMB policy id


toolness:
I left this out of #661, so it still needs to be done.

I have some architectural thoughts on this:

1. It seems like it'd be useful to move the React propTypes for a "policy" object out to a separate module, and import it into any component that happens to need it.  Is this a good idea?

2. I feel like putting the logic for this functionality into the `PolicyView` component isn't a great idea--it makes it a bit harder to unit test, for one thing, and it also seems like the kind of functionality that other components might want to use. If a Policy were a regular JS class, I'd recommend adding an `asUrl()` method to it, which performed this functionality. But Policy is currently just a plain JSON-serializable object, as I understand it... So one option is to make a bunch of functions that take this object as its first parameter, while another might be to "embellish" Policy into its own full JS class. I also might be getting a bit architecture-astronauty here so let me know if I'm thinking about this too much.

cmc333333:
I think making the policy into as full-blown class would make sense here, particularly if we'll be using it in multiple places (which seems likely). This approach was taken for document nodes, to some success.

I certainly appreciate your hesitancy to introduce OO, though. I struggle a lot with that balance.

toolness:
Yeah! Ok cool, I will take a look at the document node class and try to replicate it for the policy, and use that to implement a solution for this issue.

toolness:
Ok, the `Policy` class is added, but I just noticed that it looks like our serializer on the back-end doesn't actually expose the slug, so fixing this issue will require adding that one too.

Issue #673, opened 2017-11-17
Account for randomness in test failures


cmc333333:
Occasionally, a test will fail due to randomization. Specifically, we're testing for "fbi" when we search; there's a non-zero chance one of the randomly populated fields will contain that string.

See [this CI build](https://circleci.com/gh/18F/omb-eregs/739) for an example.

PR #690, opened 2017-11-20
Add a Policy class

toolness:
As per the discussion in #678, this factors out a new `Policy` class to contain business logic for policies, so that it can easily be reused by different views.

Note that this is only a **refactoring**; it doesn't add any new functionality to the site.


ui/__tests__/util/policy.test.js
```diff
@@ -0,0 +1,42 @@
...
+    id: 1,
+    has_docnode: false,
+    omb_policy_id: 'M-16-19',
+    ...props,
+  });
+}
+
+describe('Policy', () => {
+  it('is frozen', () => {
+    expect(Object.isFrozen(policy())).toBe(true);
```
toolness: I might be going overboard with this freezing thing, let me know what you think!

ui/components/policies/policies-view.js
```diff
@@ -14,17 +15,15 @@ export default function PoliciesView({ policies, count, topicsIds }) {
       <ThingCounterContainer count={count} singular={singular} plural={plural} />
       <ul className="policy-list list-reset">
         { policies.map(policy =>
-          <Policy key={policy.id} policy={policy} topicsIds={topicsIds} />,
+          <PolicyView key={policy.id} policy={new Policy(policy)} topicsIds={topicsIds} />,
```
toolness: I'm rusty on my React, so I'm not sure if constructing a brand-new `Policy` every render is going to irritate React's innards and make it think that props are changing all the time, thus causing lots of extra pointless work. We should figure this out before merging, methinks.

ui/util/policy.js
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  getDocumentLinkProps() {
+    return {
+      route: 'document',
+      params: {
+        policyId: this.omb_policy_id,
+      },
+    };
+  }
```
toolness: These methods *could* be computed properties, but I felt that making them explicit methods might make it easier for readers to distinguish between what are original properties of policy structs vs. extra business logic that we add. Let me know if you disagree though, we can always change them.

toolness:
Um, CircleCI is failing on an API test, so that's confusing. Oh, it looks like it's a recurrence of #673.

ui/__tests__/util/policy.test.js
```diff
@@ -0,0 +1,48 @@
...
+  });
+
+  it('copies properties from passed-in object', () => {
+    expect(policy({ id: 4 }).id).toBe(4);
+  });
+
+  it('Serializes (via JSON) to the original passed-in object', () => {
```
toolness: Honestly there's no strong reason that I added this; I just wanted the `Policy` class to be as close to the struct in behavior as possible, I guess.

ui/util/policy.js
cmc333333: I'd actually lean heavier on the properties side; when I see a method call, I assume work is being done. You raise a good point about distinguishing between API values and new logic. Maybe we make the decision about whether or not to duplicate the API's interface and, if not, keep these as methods?

ui/util/policy.js
```diff
@@ -0,0 +1,37 @@
...
+}
+
+Policy.shape = PropTypes.shape({
+  id: PropTypes.number,
+  title_with_number: PropTypes.string,
+  relevant_reqs: PropTypes.number,
+  total_reqs: PropTypes.number,
+  has_docnode: PropTypes.bool,
+  omb_policy_id: PropTypes.string,
+});
```
cmc333333: Neat idea!

toolness:
Cool! Ok, I think your suggestions are good but for now I am gonna do them in separate PRs, and commit this one as-is.

PR #692, opened 2017-11-21
Set random seed in test_filter_agency_by_abbr to avoid random failures

toolness:
It appears [model mommy uses `random` for its randomness](https://github.com/vandersonmota/model_mommy/blob/development/model_mommy/random_gen.py), which means that calling [`random.seed()`](https://docs.python.org/2/library/random.html#random.seed) with a consistent value should make individual test results reproducible.

Currently this fix just calls `random.seed(1)` at the beginning of `test_filter_agency_by_abbr`, which is the test that is particularly susceptible to random failures, as mentioned in #673 (and which repeated itself in #690 for me).

An alternative fix might be to add some kind of fancy pytest fixture that *always* sets the random seed at the beginning of every test function, ensuring that *all* tests produce consistent and reproducible results. Do we have a preference?

### DIFF
--- a/api/reqs/tests/views_simple_tests.py
+++ b/api/reqs/tests/views_simple_tests.py
@@ -1,3 +1,5 @@
+import random
+
 import pytest
 from model_mommy import mommy
 from rest_framework.test import APIClient
@@ -7,6 +9,9 @@ from reqs.models import Agency
 
 @pytest.mark.django_db
 def test_filter_agency_by_abbr():
+    # This is a workaround for https://github.com/18F/omb-eregs/issues/673.
+    random.seed(1)
+
     mommy.make(Agency)
     fbi = mommy.make(Agency, abbr='FBI')
     gsa = mommy.make(Agency, name='General Services Administration')


cmc333333:
I think we _want_ randomness in our tests, so setting a fixed seed feels like a mistake. I worry that it's also a rather indirect solution for this particular problem. Can we correct the issue by spelling out the agencies?
```python
mommy.make(Agency, abbr='non', name='Not an agency')
fbi = mommy.make(Agency, abbr='FBI', name='Federal Bureau of Investigation')
gsa = mommy.make(Agency, abbr='GSA', name='General Services Administration')
```

Alternatively, if randomness seemed important to this test:
```python
mommy.make(Agency, abbr=non_fbi_str(), name=non_fbi_str())
fbi = mommy.make(Agency, abbr='FBI', name=non_fbi_str())
gsa = mommy.make(Agency, abbr=non_fbi_str(), name='General Services Administration')
```

toolness:
Hmm, so one concern I have is regarding what happens if we end up adding more fields to the `Agency` model, e.g. `description`.  Just adding those fields could then cause this test to break, which seems funky.

Some other alternatives:

1. Change the abbreviated/searched-for agency from `FBI` to an acronym that's less susceptible to randomly occurring in our data, e.g. `FBIZZZZZ`.

2. Change model mommy to actually use different kinds of random data for each field, depending on what the field is supposed to contain.  This is a feature that factory boy, an alternative to model mommy, has; I'm not sure if model mommy supports it or not.

Another concern I have is with regards to the _reproducibility_ of tests. It's a slightly different concern, but it seems useful to at least e.g. set the random seed to the current timestamp before each test (thus ensuring actual randomness), but then output that seed if the test fails, so that developers can consistently reproduce it if needed.  In the case of `test_filter_agency_by_abbr`, for instance, it's relatively easy to determine what the cause of the random test failures was, but I'm concerned that might not be the case for more complex tests.

cmc333333:
This particular test is probably poorly named as it's really testing that the "search" feature is including the abbreviation. I'd think that if we added another field to search we shouldn't be surprised if it broke a search test.

Using a more archaic string will reduce the frequency of these errors, for sure. Seems like a reasonable partial fix.

Regarding model mommy, we can use a custom [recipe](http://model-mommy.readthedocs.io/en/latest/recipes.html) for this situation. We could either use random strings with non-overlapping characters or the `seq` function. Perhaps?

Regarding repeatability, looks like [pytest-randomly](https://pypi.python.org/pypi/pytest-randomly) will set a fixed seed at the beginning of a test (and spit it out for reproducibility). It also randomized test order, which would be a win. What do you think?

toolness:
Cool!  Ok, so for now I've just changed the agency from FBI to PCLOB, which should be harder to randomly occur in the data.

I like pytest-randomly!  Will file a separate PR to add that.

PR #698, opened 2017-11-21
Use checkPropTypes() in Policy constructor

toolness:
So one thing I'm realizing as we move some of these structs into classes (see e.g. #687) is that we're potentially losing some of the nice development affordances and self-documentation that React provides via `propTypes`.  I noticed, though, that there is a [`PropTypes.checkPropTypes`](PropTypes.checkPropTypes) function that allows properties to be checked manually. According to the documentation, "this function is safe to call in production, as it will be replaced by an empty function", so we don't need to guard it with an `if (process.env.NODE_ENV != 'production')` statement.

Thoughts?

### DIFF
--- a/ui/util/policy.js
+++ b/ui/util/policy.js
@@ -9,6 +9,7 @@ import PropTypes from 'prop-types';
 
 export default class Policy {
   constructor(policy) {
+    PropTypes.checkPropTypes(Policy.propTypes, policy, 'property', 'Policy');
     Object.assign(this, policy);
     Object.freeze(this);
   }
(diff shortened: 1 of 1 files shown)


ui/util/policy.js
```diff
@@ -9,6 +9,7 @@ import PropTypes from 'prop-types';
 
 export default class Policy {
   constructor(policy) {
+    PropTypes.checkPropTypes(Policy.propTypes, policy, 'property', 'Policy');
```
cmc333333: Do we want to assert that the policy _parameter_ has this proptype or do we want to assert the _Policy_ has the proptype? Since we're using `Object.assign`, they equate to the same thing, they are semantically different, in my mind.

toolness:
Oh, I didn't realize that flow/typescript were possibilities here! I'd love to use flow!

PR #702, opened 2017-11-27
Refactor content/annotation serialization

cmc333333:
This bundle of changesets isolates "content" serialization a bit better and converts it to use Django Rest Framework's serializers. This refactor clears the runway for allowing _nested_ inline content.


api/document/serializers/content.py
```diff
@@ -0,0 +1,93 @@
...
+            raise NotImplementedError(f'Bad annotation: {type(instance)}')
+        return content_type
+
+    def to_representation(self, instance: Annotation):
+        """Different annotations will have different attributes to display. To
+        allow that, we check for a corresponding method name."""
+        representation = super().to_representation(instance)
+        method_name = f"{representation['content_type']}_attrs"
+        if hasattr(self, method_name):
+            representation.update(getattr(self, method_name)(instance))
```
tadhg-ohiggins: I might be missing something here, but I'm unsure about this approach to finding the appropriate operations to apply. Since we're checking for the presence of a method anyway, would it be simpler to have a `handle_attrs` method for each content type and to put the various necessary operations into that method?

cmc333333: Agreed -- I'm reworking this now to use proper serializers _per annotation type_, which, I think, addresses this indirection. Will post soon.

cmc333333:
@tadhg-ohiggins updated this to include serializers for the individual annotation classes. I *think* this is easier to follow; what about you?

tadhg-ohiggins:
@cmc333333 I also think it's easier to follow, although it's a shame that we have to use `ContentListSerializer.serializer_mapping.update` rather than setting those values inline.

cmc333333:
Thanks @tadhg-ohiggins ! What do you think about using a decorator to register these? Something like:
```python
@ContentListSerializer.register(PlainText)
class PlainTextSerializer(BaseAnnotationSerializer):
    ...
```
If that conveys our meaning more, I can submit it in a later PR?

PR #703, opened 2017-11-28
Allow nested content

### LINKED ISSUE #689: Allow inline content to be nested
As part of #668, we need to deal with inline content overlapping with other inlines (in this case footnote citations). For example, requirement 21.02 contains footnote citation 7.

For this ticket, let's just focus on serializing overlapping inlines and rendering them in the UI. We'll make a separate ticket around importing.

cmc333333:
This alters the API to include nested inline elements as a tree structure, with PlainText/__text__ leaf nodes. Parent content's "text" attribute continues to include the full text, so there's no immediate need for a corresponding change in the UI.

This should resolve #689

### DIFF
--- a/api/document/serializers/content.py
+++ b/api/document/serializers/content.py
@@ -1,7 +1,6 @@
 from collections import defaultdict
-from typing import Iterator, List
+from typing import Iterator, List, Optional
 
-from collections_extended import RangeMap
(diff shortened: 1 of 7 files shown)


cmc333333:
Thanks @tadhg-ohiggins!

PR #706, opened 2017-11-29
Inline Requirement API data

cmc333333:
This creates a model to represent inline requirements and includes them when serializing. It also removes the existing one-requirement-per-paragraph logic.


cmc333333:
Thanks @tadhg-ohiggins! I'll rebase then merge.

PR #712, opened 2017-11-29
Minor style updates

cmc333333:
Pulls in some of the content from #679 and adds a tweak to footnote citation line height:

## Before
screen shot 2017-11-29 at 3 24 41 pm

## After
screen shot 2017-11-29 at 3 24 50 pm

PR #713, opened 2017-11-29
UI for inline requirements

cmc333333:
This is the mirror of #706, taking the data from the API and displaying it in the user interface. It also allows external links to include other types of annotations and fixes an issue with the nesting logic.

Looks like (with #706 and #712)
screen shot 2017-11-29 at 5 00 50 pm


cmc333333:
Thanks @rememberlenny! Per our discussion during standup, we might remove this feature soon, but it won't hurt to have it (and the related nested inlines) now.

PR #714, opened 2017-11-30
Add styles for footnote triangle

rememberlenny:
footnote

### DIFF
--- a/ui/css/module/_footnotes.scss
+++ b/ui/css/module/_footnotes.scss
@@ -1,9 +1,21 @@
+$one-pixel-value: 0.063;
+$two-pixel-value: $one-pixel-value * 2;
+$one-pixel: $one-pixel-value + rem;
+
+$footnote-vertical-margin-value: 1.5;
+$footnote-vertical-margin: $footnote-vertical-margin-value + rem;
+
+$triangle-dimensions-value: 1.125;
+$triangle-dimensions: $triangle-dimensions-value + rem;
+
+$triangle-border-vertical-offset: ($triangle-dimensions-value + $footnote-vertical-margin-value - $two-pixel-value) + rem;
+
 @mixin footnote-border-triangle-up() {
   content: '';
   position: absolute;
   background: $color-white;
-  height: 18px;
-  width: 18px;
+  height: $triangle-dimensions;
+  width: $triangle-dimensions;
   margin-left: -9px;
   left: 50%;
   border-top-right-radius: 3px;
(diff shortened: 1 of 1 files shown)


ui/css/module/_footnotes.scss
```diff
@@ -5,14 +5,38 @@
   padding: 2px 4px;
   font-size: 0.6em;
   margin-left: .3em;
+  position: relative;
 
   &.active {
-    border-color: $color-primary-darker;
-    background: $color-primary-darker;
-    color: $color-white;
+    @extend .active-footnote-style;
```
cmc333333: Does a mixin make sense here? The concern I have with `extend` is that we have to create a "real" css class, which might trip us up later.

ui/css/module/_footnotes.scss
```diff
@@ -5,14 +5,38 @@
...
+      position: absolute;
+    }
   }
 }
 
+.active-footnote-style {
+  border-color: $color-primary-darker;
+  background: $color-primary-darker;
+  color: $color-white;
+}
```
cmc333333: Not for now, but something to consider going forward: is this a building block we'd want to use elsewhere? Will we have other white on blue things? If so, should we make a more generic mixin?

ui/css/module/_footnotes.scss
```diff
@@ -5,14 +5,38 @@
...
+
+.rotated-triangle {
+  top: 39px;
```
cmc333333: Is there a way to compute this from other values (maybe based on the margin value of the inline footnote?) so it is less of a magic number?

ui/css/module/_footnotes.scss
```diff
@@ -5,14 +5,38 @@
...
+  background: $color-white;
+  height: 18px;
+  width: 18px;
+  left: 14px;
```
cmc333333: Is there a way to do this (in CSS) based on a percentage rather than a fixed offset? With the potential for different fonts we might have very different "middle"s for a footnote citation (e.g. "Footnote 88" is much wider than "Footnote 1")

rememberlenny:
@cmc333333 Held back on merging, in case you have any thoughts on the sass variables.

ui/css/module/_footnotes.scss
```diff
@@ -1,16 +1,49 @@
...
+  border-right: 1px solid $color-primary-darker;
+}
+
 .footnote-link {
   @include font-san-serif();
   border: 1px solid $color-cool-blue-lighter;
   border-radius: 4px;
   padding: 2px 4px;
-  font-size: 0.6em;
+  font-size: 0.7em;
```
rememberlenny: I reviewed Ryans designs and noticed our implementation was definitely too small. This looks more like the design.

cmc333333:
I think we can be a bit more semantic with these variables, but I'm happy to merge this now. I'll follow up with a PR today to show what I mean.

Thanks @rememberlenny!

PR #715, opened 2017-11-30
Propagate 404s from the API for paging errors.

### LINKED ISSUE #691: Invalid page numbers causing 500s
https://omb-eregs-demo.app.cloud.gov/policies?page=11 works just fine but https://omb-eregs-demo.app.cloud.gov/policies?page=12 explodes. This is because the API is giving a 404 and we're not bubbling that up.

I believed the API would return an empty (but successful) response in this scenario; my bad! We'll need to update the ui to handle that 404.

cmc333333:
Django Rest Framework returns a 404 when the user requests a non-existent page
number in a listing. Previously, we didn't account for that in the UI and just
raised the exception. This resolves by reusing our existing propagate404
function on the listing pages.

It might make sense to apply this to _all_ listing endpoints, but I've not
thought through the ramifications enough to be confident that that's a good
idea.

This should resolve #691

### DIFF
--- a/ui/__tests__/util/api/queries.test.js
+++ b/ui/__tests__/util/api/queries.test.js
@@ -3,12 +3,17 @@ import {
   documentData,
   formatIssuance,
   homepageData,
+  policiesData,
   policyData,
   redirectQuery,
+  requirementsData,
 } from '../../../util/api/queries';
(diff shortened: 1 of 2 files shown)


cmc333333:

PR #716, opened 2017-11-30
Copy derived fonts to container in CI.

cmc333333:
We pull some fonts from a node library and place them in the "static"
directory *within a remote container*. We need to bring them to the local file
system so they can be used when deploying.

This [successfully deploys](https://circleci.com/gh/18F/omb-eregs/828).


cmc333333:
Thanks @tadhg-ohiggins !

PR #717, opened 2017-11-30
Bump up moment.

cmc333333:
This is a security release. We aren't in danger as we only use trusted input,
but updating won't hurt, either.


cmc333333:

PR #719, opened 2017-12-01
Palette reduction

cmc333333:
@rtwell and I worked through this migration of ~50 color variables down to a dozen. As he's already reviewed the color selections, I'm hoping this review can be just a sanity check.


cmc333333:
Thanks for reviewing, @rememberlenny and @tadhg-ohiggins and to @rtwell for pairing with me!

PR #723, opened 2017-12-01
Suggestions for #714

cmc333333:
I had a few ideas for tweaks on #714 but couldn't describe them in words too well, so I broke them into a few commits. Please take a look (reviewing changeset by changeset would be helpful) to see if you agree.

### DIFF
--- a/ui/css/module/_footnotes.scss
+++ b/ui/css/module/_footnotes.scss
@@ -33,7 +33,7 @@ $footnote-vertical-margin: 24px;
 
     &::before {
       @include footnote-border-triangle-up();
-      top: $triangle-dimensions + $footnote-vertical-margin + 2 * $footnote-border;
+      top: $triangle-dimensions + $footnote-vertical-margin;
       z-index: 10;
     }
   }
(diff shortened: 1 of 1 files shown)


ui/css/module/_footnotes.scss
```diff
@@ -40,8 +31,9 @@ $triangle-border-vertical-offset: ($triangle-dimensions-value + $footnote-vertic
     color: $color-white;
 
     &::before {
-      @include footnote-border-triangle-up();
-      top: $triangle-border-vertical-offset;
+      $triangle-dimensions: 18px;
+      @include footnote-border-triangle-up($triangle-dimensions);
+      top: $triangle-dimensions + $footnote-vertical-margin;
```
rememberlenny: Im guessing you tested this, but we need to also subtract for the two pixel borders.

cmc333333: I removed the borders in the second commit because they _aren't_ lined up now.

## Current Demo
screen shot 2017-12-01 at 5 40 44 pm

## With change
screen shot 2017-12-01 at 5 42 02 pm

That said, it's slightly different in firefox:

screen shot 2017-12-01 at 5 42 49 pm

rememberlenny: 👍

cmc333333: I wonder if the recent change to the citation text's font size caused the problem?

cmc333333:
Thanks @rememberlenny

PR #724, opened 2017-12-04
Upgrade React, Next, Enzyme

cmc333333:
In updating to the new version of React (which allows useful features like
components returning _fragments_), we must also update all of the dependencies
which peer with React. We'll next fix up any broken tests, etc. Upgrade docs:

* React: https://reactjs.org/blog/2017/09/26/react-v16.0.html#upgrading
* Next: https://zeit.co/blog/next4
* Enzyme:
* https://github.com/airbnb/enzyme/blob/master/docs/guides/migration-from-2-to-3.md


cmc333333:
Thanks @rememberlenny!

PR #731, opened 2017-12-04
Importer for requirements

cmc333333:
Though it seems likely that we'll be reworking inline requirements, this includes two bug fixes that seemed worth pushing up.


cmc333333:
@tadhg-ohiggins would replacing `__call__()` with `derive()` make this more clear, do you think?

tadhg-ohiggins:
@cmc333333 Yes, I think so, even though that's not as succinct.

cmc333333:
Thanks @tadhg-ohiggins

PR #737, opened 2017-12-05
Add data for in-document navigation

### LINKED ISSUE #732: Implement in-document navigation data
I recommend adding a "title" field to nodes and adding headings to section titles so we compute the abstract titles once, at import time.

cmc333333:
This adds a field to track the "title" of nodes throughout the document. We won't use the header nodes directly as having an abstraction lets us point to give _semantic_ titles to non-header elements (like sections) rather than pointing to a heading within them.

This adds data to the example M-16-19 and serializes it in API responses. This should resolve #732


cmc333333:
Thanks @tadhg-ohiggins

Issue #728, opened 2017-12-04
Add another policy so we have the M-16-19 view for it


tadhg-ohiggins:
Not sure yet which is the best one to tackle.

yowill:
Done!

PR #738, opened 2017-12-05
Add Cite as a supported element and add M-15-16 to the example docs 

tadhg-ohiggins:
Works are distinct things.
So represent them as such.
We like semantics.

Adds the HTML concept of `cite`, denoting a specific work or reference, to the system.

Also adds M-15-16 as another document, so this is part of https://github.com/18F/omb-eregs/issues/728 but doesn't finish it because we need to figure out what to do about emphasis and/or “paragraph title” concepts. Note that while I use `underline` in the XML for M-15-16, this is likely to change due to the concerns just mentioned.

This PR might also be useful as a reference for how to add new element support; thanks @cmc333333 for helping me through that.


tadhg-ohiggins:
It's an even better reference PR once I remember to also commit the Django migration and to account in the tests for the extra queries caused by adding a new component to look for.

api/document/tests/views_test.py
```diff
@@ -103,4 +103,4 @@ def test_query_count(client):
         # 7: fetch footnote citations _and_ referenced node for child nodes
         # 8: fetch external links for child nodes
         # 9: fetch inline requirements _and_ referenced req for child nodes
-        assert len(capture) == 9
+        assert len(capture) == 11
```
cmc333333: Would you mind updating the comment above as well? This'll be analogous to external links

api/example_docs/m_15_16.xml
```diff
@@ -0,0 +1,367 @@
...
+        Scientific discovery, technological breakthroughs, and innovation are
+        the primary engines for expanding the frontiers of human knowledge and
+        are vital for responding to the challenges and opportunities ofthe
+        2P1 century.  The Nation depends
+        on science, technology, and innovation to promote economic growth and
+        job creation, maintain a safe and sufficient food supply, improve the
+        health of Americans, move toward a clean energy future, address global
+        climate change, manage competing demands on environmental resources,
+        and ensure the Nation's security. 
+      </content>
```
cmc333333: Shortcut for the future: you don't need to include the `content` tag if there's no inline elements.

api/example_docs/m_15_16.xml
```diff
@@ -0,0 +1,367 @@
...
+      <content>
+        In the FY 2017 Budget, agencies should balance priorities to ensure
+        that resources are adequately allocated for agency-specific,
+        mission-driven research, including fundamental research, while focusing
+        resources, where appropriate, on the following multi-agency research
+        activities that cannot be addressed effectively by a single agency. 
+      </content>
+    </para>
+    <list>
+        <listitem emblem="1" marker="•">
```
cmc333333: These'll be numbered automatically, so we probably don't need to spell out their emblem here.

Issue #704, opened 2017-11-28
Footnotes should include links back to footnote citations


tadhg-ohiggins:
Otherwise, getting to the citation in the document (either after clicking on the citation in an environment without the required JS support, or while reading through the footnotes at the end) is too difficult.

PR #739, opened 2017-12-05
Add link from footnote back up to citation

rememberlenny:
# Reference 

#704


ui/components/content-renderers/footnote-citation-in-table.js
```diff
@@ -6,7 +6,7 @@ import Link from '../link';
 export default function FootnoteCitationInTable({ content }) {
   const href = `#${content.footnote_node.identifier}-table`;
   return (
-    <sup className="footnote-citation-in-table">
+    <sup className="footnote-citation-in-table" id={`citation-${content.text}`}>
```
cmc333333: The text of a footnote citation and the marker of the footnote it refers to might be slightly different. Can we use `content.footnote_node.identifier` in both, similar to line 7?

ui/components/node-renderers/footnote.js
```diff
@@ -9,7 +9,9 @@ export default function Footnote({ docNode }) {
     <span className="clearfix node-footnote" id={docNode.identifier}>
       <span className="node-footnote-content">
         <span className="citation-marker">
-          { docNode.marker }
+          <a href={`#citation-${docNode.marker}`}>
```
cmc333333: Let's use `Link` for consistency. It should handle anchors correctly.

ui/css/module/_footnotes.scss
```diff
@@ -128,6 +128,10 @@ $footnote-vertical-margin: 24px;
     line-height: 1rem; // like font-text, except adjusting for bold
   }
 
+  a {
+    color: $color-black;
+  }
```
cmc333333: Do we want this to apply to _all_ anchors at the bottom of the page, or just the markers? Designs for external links are blue, for example.

ui/components/content-renderers/footnote-citation-in-table.js
rememberlenny: Definitely agree on this. Updated.

ui/components/node-renderers/footnote.js
rememberlenny: Added!

ui/css/module/_footnotes.scss
rememberlenny: Not at all. Good catch.

rememberlenny:
@cmc333333 Updated with changes.

PR #741, opened 2017-12-06
Add a very preliminary style guide

toolness:
This is an extremely preliminary style guide, inspired by the [CALC style guide](https://calc.gsa.gov/styleguide/). It's just a page at `/styleguide` that tries putting all our styles and components in one place, so that @rtwell and others can easily verify that styling looks OK.

> styleguide


ui/pages/styleguide.js
```diff
@@ -0,0 +1,107 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import HeaderFooter from '../components/header-footer';
+import Link from '../components/link';
+import LabeledText from '../components/labeled-text';
+
+const GITHUB_URL = 'https://github.com/18F/omb-eregs';
+
+function Example({ children }) {
```
cmc333333: I _think_ this component is meant to wrap app components so we can see clearly where they start and end. Is that right? Should it take a title in that case?

It might also be worth noting why inline styles are the way to go here.

ui/pages/styleguide.js
```diff
@@ -0,0 +1,107 @@
...
+      <p>
+        {`In the future, we should have examples of our policy document
+          styles here. It might actually be easiest to just create a
+          fake, development-only policy document that just contains a
+          "kitchen sink" of anything that can possibly appear in a
+          policy document.`}
+      </p>
+      <p>
+        For now, most policy document styles can be seen on {' '}
+        <Link href="/document/M-16-19">M-16-19</Link>.
```
cmc333333: Let's use the `route` prop to indicate this is an internal link.

ui/pages/styleguide.js
```diff
@@ -0,0 +1,107 @@
...
+      </p>
+    </div>
+  );
+}
+
+export default () => <HeaderFooter><Styleguide /></HeaderFooter>;
```
cmc333333: We should probably use `wrapPage` function rather than `HeaderFooter`. The former sets up analytics, redux, and catches errors.

Unrelated, but the AirBnB guide [recommends](https://github.com/airbnb/javascript/tree/master/react#class-vs-reactcreateclass-vs-stateless) using normal functions over arrows for components.

toolness: Oh heh, this was actually just copypasta'd from the privacy policy page... should we make these fixes over there too?  Or maybe not, since that page really is just static content...

ui/pages/styleguide.js
toolness: Oh, uh, that's mostly lazyness on my part.  Although I also figured this is the only place the styles will actually be used, so maybe it's just easier to inline them? Eh, I'm happy to factor them out into a SCSS thingy.

toolness: Hmm, the example component is really just meant as a sort of "template partial" to show an example snippet of code--it can be just one component, or it can be a bunch of them, or it can also be things that aren't very component-like at all, such as the first example which just shows typography.  I think the equivalent in CALC did have an *optional* title, but it wasn't required...

cmc333333: Re styling: I could see an argument around inlining these so they aren't overridden by one of the imported stylesheets... or something along those lines. If that's not the reason, SCSS (or basscss classes) would be better.

Re title: Gotcha. If you can think of a way to document that sentiment, it might be helpful. Having more examples may resolve the uncertainty, too, though, so we might just wait.

ui/pages/styleguide.js
cmc333333: Oh no, we should `wrapPage` everywhere, otherwise redux won't get initialized; add that to ajaxy page loading and we'll have an issue. I can make a separate issue unless you want to fixup the privacy policy page now.

toolness: Actually can you make a separate issue? I'm not sure if the privacy policy page is the only exception, and I'd prefer to keep this PR focused on the styleguide stuff.

ui/pages/styleguide.js
toolness: Ok, I added a styleguide SCSS file. I think that ideally it would be a separate SCSS file that's only loaded at `/styleguide` so that it doesn't bloat the core CSS bundle, but that would take some more work and it's a really tiny SCSS file... maybe we could file that as a separate issue to tackle later?

cmc333333: Created #743 to track, though I made it a bit more general than this one page.

ui/css/module/_styleguide.scss
```diff
@@ -0,0 +1,26 @@
+.styleguide-container {
+  padding: 1rem;
+  max-width: 800px;
```
cmc333333: Why do we need to limit the width here?

ui/css/module/_styleguide.scss
```diff
@@ -0,0 +1,26 @@
...
+    text-transform: uppercase;
+    color: $color-gray;
+    letter-spacing: 1px;
+    font-size: 1em;
+  }
+}
+
+.example-title {
+  margin-top: 2rem;
+  margin-bottom: 1rem;
```
cmc333333: Rather than defining everything in constant rems, could we use the new `$spacing` variables?

ui/css/module/_styleguide.scss
```diff
@@ -0,0 +1,26 @@
...
+
+.styleguide-example {
+  border: 1px solid $color-lt-gray;
+  padding: 1rem 2.5rem;
+
+  h4 {
+    text-transform: uppercase;
+    color: $color-gray;
+    letter-spacing: 1px;
+    font-size: 1em;
```
cmc333333: Will this override the nested h4 in `styleguideTypographyHeading`? If so, can we just remove these rules? They don't seem super important.

ui/css/module/_styleguide.scss
```diff
@@ -0,0 +1,26 @@
...
+  }
+}
+
+.example-title {
+  margin-top: 2rem;
+  margin-bottom: 1rem;
+}
+
+.styleguide-example>.document-header {
```
cmc333333: Should this rule be nested in `.styleguide-example` rather than separate?

ui/pages/styleguide.js
```diff
@@ -0,0 +1,150 @@
...
+          "kitchen sink" of anything that can possibly appear in a
+          policy document.`}
+      </p>
+      <p>
+        For now, most policy document styles can be seen on {' '}
+        <Link route="document" params={{ policyId: 'M-16-19' }}>M-16-19</Link>.
+      </p>
+    </div>
+  );
+}
```
cmc333333: Would it make sense for these functions to be constants instead? That'd give a better sense of their intended use.

ui/css/module/_styleguide.scss
rememberlenny: @cmc333333 For the time being, the page is way too wide on a large screens. @rtwell is looking at a better way to style this, but it helps.

rememberlenny:
@cmc333333 Applied updates!

ui/css/module/_styleguide.scss
```diff
@@ -1,26 +1,22 @@
+$basic-spacing: 1rem;
+$medium-spacing: 2.5rem;
```
cmc333333: Probably better to use the [existing](https://github.com/18F/omb-eregs/blob/9b0b4e8302662b2dcb729ec7f549281b18f5c0bb/ui/css/base/_spacing.scss) spacing vars, yes?

Issue #718, opened 2017-11-30
visual tweaks to M-16-19


rtwell:
## Random visual tweaks 

- [x] 1) The footnotes appear to be altering the line height 
- [ ] 2) let's kill the gold in the H4s and change to black 
- [ ] 3) for numbered lists, let's simplify and use sans-serif for the numbers as well. I thought it looked easier to read with the sans in there for numbers, but the difference in the base-size of the fonts is throwing the alignment off. 
- [ ] 4) The spacing between the headers and body copy should be adjusted:
### live site
![image](https://user-images.githubusercontent.com/12564977/33456966-a4e8980c-d5e6-11e7-8143-d714d0e86a83.png)
### no markup
![image](https://user-images.githubusercontent.com/12564977/33457287-a36b35a6-d5e7-11e7-94d5-54e54c1d508c.png)
### markup
![image](https://user-images.githubusercontent.com/12564977/33457463-2ecde7ec-d5e8-11e7-9599-7fd4ba178a69.png)

cmc333333:
Heads up that #712 should resolve 1 (and perhaps 4).

cmc333333:
Resolved by #786

PR #742, opened 2017-12-06
Add desktop navigation

### LINKED ISSUE #733: Implement desktop in-doc navigation styling
#585 includes a design for navigation within the desktop. Implement that without the position-in-document logic and stickiness (which are separate issues).

Depends on #732

cmc333333:
This adds components to represent the document navigation menu. It uses the data from #737 for population (meaning we'll need to re-import that doc).

This does _not_ implement mobile, navigation stickiness, or highlighting the user's current position (which are all separate tickets).

Looks like
screen shot 2017-12-06 at 3 43 47 pm

Should resolve #733

### DIFF
--- a/ui/__tests__/components/document/navigation-heading.test.js
+++ b/ui/__tests__/components/document/navigation-heading.test.js
@@ -1,0 +1,51 @@
+import { shallow } from 'enzyme';
+import React from 'react';
+
+import NavigationHeading from '../../../components/document/navigation-heading';
+
+describe('<NavigationHeading />', () => {
+  it('works with no children', () => {
+    const tocNode = { children: [], identifier: '', title: '' };
+    const result = shallow(<NavigationHeading tocNode={tocNode} />);
+    const links = result.find('Link');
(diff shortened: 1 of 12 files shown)


rtwell:
this is looking great! a couple tweaks:

1. looks like our line-height for the nav text is too generous: let's try ~`1.3 ems`
2. the space around the type should be equal, so if we could shift the right indent to match the top/bottom, that would be great!

cmc333333:
@rtwell tried to address with some style tweaks. The left padding is now 1.5rem (slightly bigger), right is 1rem (wasn't defined before), and line height's at 1.3rem (slightly smaller). Now looks like:
screen shot 2017-12-06 at 5 58 44 pm

rtwell:
looks great! i have some different colors i'd like to use, but I can include those edits with https://github.com/18F/omb-eregs/issues/718

cmc333333:
Thanks @rtwell and @rememberlenny -- I'll rebase and merge

PR #744, opened 2017-12-06
Ensure all /pages are wrapped.

cmc333333:
The app-wrapper sets up redux (along with some error handling, etc.), so it's
important it be called at all entrypoints. Notably, if a user can do AJAX, but
initially loads the privacy page (which wasn't wrapped), they'll receive an
error when trying to use redux on the document page.

To do this, we needed to allow pages to have a no-op data fetching function.


cmc333333:
Thanks @rememberlenny . I'll rebase and merge

PR #745, opened 2017-12-06
Remove extraneous divs/spans.

cmc333333:
We had been wrapping these elements in a div or a span because React <16
didn't know how to render "partial" elements: plain strings or array.

This isn't a huge change in terms of file size (we drop ~2k off M-16-18), but
it does make the markup much cleaner.


cmc333333:
Thanks @tadhg-ohiggins

PR #746, opened 2017-12-06
Remove "policy" endpoint.

cmc333333:
We had built this while deciding what to do about "partial" document data.
That's not needed anymore, so we can remove it entirely.


cmc333333:
Thanks @rememberlenny - I'll rebase and merge

Issue #735, opened 2017-12-05
Track user's position in document, highlight in ToC


cmc333333:
#585's design includes indicating the user's current position within the document. Implement that.

Depends on #733

cmc333333:
Resolved with #750, #752, #759

PR #750, opened 2017-12-08
Store table of contents in Redux

cmc333333:
While working on #735, it seemed helpful to put the table of contents in our Redux store instead of needing to repeatedly derive it in different components. This also
1. Clears the open footnote on document load
2. Let's us use [Redux Devtools](https://github.com/zalmoxisus/redux-devtools-extension)

The latter looks like
screen shot 2017-12-08 at 9 24 45 am


rememberlenny:
@cmc333333 This looks good. Im curious, mainly out of ignorance, if there are any pitfalls with loading large amounts data through Redux.

cmc333333:
Thanks @rememberlenny. I'll wait a little bit to merge in case there are other comments. 

There are two drawbacks I can think of about putting too much data in the redux store:
1. We serialize the store when sending the initial page down. More data means more to serialize. This is also true for our-top-level props (which are serialized by Next).
2. Whenever the store is updated, logic is ran to figure out what parts of the React component tree need to be updated. More stuff in the store could imply more changes, meaning more of those tests. In the pathological case, we could have some very complicated component that requests the entire store. Every change to the store would re-render.

PR #752, opened 2017-12-11
Highlight current section in ToC

cmc333333:
This tracks the user's current section by tracking which sections are on screen and picking the bottom-most. It doesn't know what to do when we scroll to the footnotes at the bottom of the page yet.

Looks like
scroll-tracking


cmc333333:
Thanks @tadhg-ohiggins! I'll rebase and merge. I'll add your suggestion around variable naming to #759.

PR #754, opened 2017-12-11
Bring in the ombpdf project via 'git read-tree'.

toolness:
This was done using the instructions here:

  https://git-scm.com/book/en/v1/Git-Tools-Subtree-Merging

Specifically, the commands I ran were:

```
  git remote add ombpdf_remote https://github.com/18F/omb-pdf
  git fetch ombpdf_remote
  git checkout -b ombpdf master
  git read-tree --prefix=ombpdf/ -u ombpdf_remote/master
```

Interestingly, the instructions indicate that we can keep merging in changes from [omb-pdf](https://github.com/18F/omb-pdf), and even merge-in changes we make here upstream to it. That might actually be the easiest thing to do in the short-term, but we'll see!

### DIFF
--- a/ombpdf/.circleci/config.yml
+++ b/ombpdf/.circleci/config.yml
@@ -1,0 +1,26 @@
+version: 2
+jobs:
+  build:
+    working_directory: ~/omb-pdf
+    docker:
+      - image: circleci/python:3.6.2
+    steps:
+      - checkout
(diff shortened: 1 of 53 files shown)


toolness:
Hmm, I think I am gonna close this in favor of #755.

PR #761, opened 2017-12-12
Dockerize ombpdf development server

### LINKED ISSUE #760: Dockerize development server for ombpdf
In #757, I opted to not dockerize the development server for `ombpdf` for various reasons. It would be nice to dockerize it, though, so that developers don't have to juggle both docker and virtualenv/npm-based development for the project.

This could be done in a number of ways. For example, the development server uses Flask, but if we convert it to use Django instead, it could just run off the API server when `DEBUG` is enabled.  That still leaves the question of how to generate the JS bundle used by the server, but at worst we could add another node container to be responsible for that.

A test suite for the development server is also currently skipped, so we should bring that back as well.

toolness:
This is an attempt to fix #760.

To do:

- [x] Remove the `ombpdf-dev-server` directory, but ensure that the kinds of instructions it provides (on how to develop `ombpdf`) are merged into the root `README.md` or somewhere else.
- [x] Convert `cli.py download` into a Django management command.
- [x] Convert any other `cli.py` commands we want into Django management commands (maybe `fonts` but that's the only one I can think of).
- [x] Convert the webapp from Flask to Django, and expose it on the API server only when `DEBUG` is set.
- [x] Convert the webapp test suite.
- [x] Update all instances of `http://localhost:5000` in the source code to be rooted at the analogous endpoint on the API server.


docker-compose.yml
```diff
@@ -74,6 +74,17 @@ services:
...
+      - ombpdf-js-bundler
+
+  ombpdf-js-bundler:
+    image: node:6
+    volumes:
+      - .:/usr/src/app:delegated
+    environment:
+      USE_POLLING: ${USE_POLLING:-false}
+    working_dir: /usr/src/app/api/ombpdf
+    command: .docker/deps_ok_then node_modules/.bin/webpack --watch
```
toolness: @cmc333333 is this the kind of thing you were thinking of?  It's completely decoupled from the `ui` app and containers, and it's only available during development, since that's the only time the views that use the JS bundles will be exposed...

toolness:
Hmmmm, CircleCI seems to be infinitely waiting for Postgres, which is weird...

>

docker-compose.yml
```diff
@@ -74,6 +74,17 @@ services:
       DEBUG: "true"
       # Uncomment the following line for MAX logins
       # MAX_URL: https://login.test.max.gov/cas/login
+    depends_on:
+      - ombpdf-js-bundler
```
cmc333333: Note that this is overriding the `depends_on` in `PROD_API`. You'll probably want to add back `persistent_db` and `minio`

toolness:
Ok, I think this is ready for review @cmc333333 !

api/ombpdf/views.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def get_pdfmap():
+    pdfmap = {}
+    for path in DATA_DIR.glob('**/*.pdf'):
+        pdfmap[to_urlpath(path)] = path
+    return pdfmap
+
+
+_ltpages_cache = {}
```
cmc333333: This is just copy-pasted now, so we can address later, but a module-level cache here feels a bit hinky. Could we use Django's caching framework instead? Then we can configure it as needed. If that seems like overkill, what do you think about the built-in [lru cache](https://docs.python.org/3/library/functools.html#functools.lru_cache)? Both prevent unbound memory leaks.

api/ombpdf/views.py
```diff
@@ -0,0 +1,83 @@
...
+            _ltpages_cache[path] = get_ltpages(fp)
+    return OMBDocument(
+        _ltpages_cache[path],
+        filename=path.relative_to(DATA_DIR),
+    )
+
+
+def index(request):
+    return render(request, 'ombpdf/index.html', {
+        'name': __name__,
```
cmc333333: I'm confused by this -- why do we want the module name in our template?

docker-compose.yml
cmc333333: This is better than what I was thinking of! Having this as a dependency is a great idea and should ensure webpack'll keep running.

api/ombpdf/management/commands/pdf_fonts.py
```diff
@@ -0,0 +1,19 @@
...
+
+import ombpdf.fontsize
+import ombpdf.util
+
+
+class Command(BaseCommand):
+    help = 'Show font usage statistics for a PDF.'  # NOQA
+
+    def add_arguments(self, parser):
+        parser.add_argument('filename')
```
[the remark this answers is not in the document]
toolness: Oh cool, I was looking for something like that but couldn't find it when skimming the docs--will switch to it!

api/ombpdf/views.py
toolness: Sure, I think lru-cache is good. Using Django's caching framework concerns me a bit b/c if we end up using an out-of-process cache, it could actually mess with my intended development use case, which is that I want the cache to be cleared whenever I make a change to the code that generates the cache's content (e.g. the core `ombpdf` python code), but *not* when I make a change to the front-end code that uses the cache's data (e.g. the HTML template and/or JS code).

api/ombpdf/views.py
toolness: Oh lol that's just weird leftover code, I can remove it.  I think it started out as code to just make sure the Flask/Jinja2 template context worked properly, and then it just stuck.

api/ombpdf/views.py
toolness: (The former will happen because of the way django's development server reboots itself whenever it detects a change to its python code--so it's convenient for the cache to have the same lifetime as the django dev server process.)

docker-compose.yml
toolness: Oh cool. I would like to talk at some point about what you *were* thinking of, because this was literally the *only* thing I could think of 😆

PR #762, opened 2017-12-12
Work around react-stickynode bug.

cmc333333:
Our "Sticky" library has a bug around activation on initial page load. This
works around that by triggering a mock scroll event on DOM mount.


cmc333333:
Thanks @tadhg-ohiggins . I'll rebase and merge.

PR #765, opened 2017-12-13
Only log warnings from pdfminer

toolness:
pdfminer has a *lot* of informational logging output, which spams the logs during development, so this changes it to only log warnings/errors.

The change in behavior can be seen by running e.g. `manage.py pdf_fonts data/2016/m_16_19_1.pdf`.


toolness:
Thanks!

As an aside, I had no idea it was possible to have a `''` key in `LOGGING['loggers']` which catches all loggers. That completely blew my mind and Django logging is now a lot less confusing to me.

Issue #734, opened 2017-12-05
Implement in-doc navigation for mobile


cmc333333:
#585 includes a design for navigation within mobile devices. Implement that without the position-in-document logic and stickiness (which are separate issues).

Depends on #732

cmc333333:
Resolved by #766 and #768

PR #766, opened 2017-12-13
Tweak mobile header, begin to standardize on media

cmc333333:
As part of #734, this adds some scss to differentiate mobile from desktop views. It then uses those to tweak the mobile header. @rtwell's agreed that these changes are a step in the right direction (though we won't implement a full redesign this sprint).

Looks like:
screen shot 2017-12-13 at 11 03 28 am
screen shot 2017-12-13 at 11 03 12 am


cmc333333:
Thanks @tadhg-ohiggins! @rtwell was on board with the screenshots, so I'm going to rebase and merge.

Issue #722, opened 2017-12-01
Basic PDF import imports to the database


tadhg-ohiggins:
Currently we have a few forms of output, but nothing that goes directly to the database. We should add that capability.

cmc333333:
Resolved with #767 and #799

Issue #682, opened 2017-11-17
MVP policy editor feature list


tadhg-ohiggins:

yowill:
Closing. Linked to this https://github.com/18F/omb-eregs/issues/727

PR #767, opened 2017-12-13
Import PDFs into the database

toolness:
This attempts to convert PDFs to XML and then import them into the database, fixing #722.

For example, running the following:

```
docker-compose run manage.py pdf_to_db data/2016/m_16_19_1.pdf --policy M-16-19
```

will overwrite any existing document for M-16-19 in the database and replace it with the following, which has been imported via PDF:

> m-16-19-imported

## Implementation notes

Originally I started implementing a PDF-to-XML converter via [`ElementTree`](https://docs.python.org/3/library/xml.etree.elementtree.html) but stopped because that library confuses me, and because outputting directly to the database would be more efficient and architecturally simpler.

However, this also proved to be cumbersome because the only current way to serialize the document in a database is as JSON, and it's much harder to read than the existing XML format, which hampered debugging.

Ultimately it was easier for me to revert back to the strategy of using XML as an intermediary, but this time using [`xml.dom.minidom`](https://docs.python.org/3/library/xml.dom.minidom.html), which is much easier for me to understand than `ElementTree`.

In the future, I think we can potentially write directly to the DB without using the XML intermediary, but for now I found the `mindom` XML approach to be more expedient.

## Other notes

One particular thing I liked about serializing to XML rather than going directly to the database is that I was able to insert comment nodes to acknowledge content in the PDF that didn't necessarily make sense, such as text we couldn't classify or footnotes lacking citations. It might be nice to extend our document model to include the concept of such content, which we could e.g. show to staff in the policy editor (#682) but hide when presenting the document to non-staff users.

## Issues to be filed later

- [ ] The emblem/marker attributes for ordered lists are currently auto-generated and always numbered. Ideally, however, we should take them from the original PDF. Fixing this will require `semtree` to convey that information to writer objects.


toolness:
Oh snap, it looks like my fix to #780 actually caused a race condition in the test suite, lemme fix that.

toolness:
Ok I think this is good for review @cmc333333 !

api/ombpdf/semdom.py
```diff
@@ -0,0 +1,186 @@
...
+        self.root = self.document.childNodes[0]
+        self.cursor = self.root
+
+    def end_document(self, doc):
+        pass
+
+    def begin_heading(self, heading):
+        while self.depth >= heading.level:
+            self._pop_child()
+        self._push_child('sec', {'title': ''})
```
cmc333333: I misunderstood when reading this and assumed the `title` attr wasn't ever set. Given that it isn't set until `create_text`, should we even mention it here?

api/ombpdf/semdom.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def end_list_item(self):
+        self.writer._pop_child('content')
+        self.writer._pop_child('para')
+        self.writer._pop_child('listitem')
+
+
+class DOMWriter(semtree.Writer):
+    def __init__(self):
+        self.muted = True
```
cmc333333: It might be helpful to document this flag; I had guessed incorrectly that it skipped over chunks of the document rather than emitting xml comments.

api/ombpdf/semdom.py
```diff
@@ -0,0 +1,186 @@
...
+    def begin_footnote_list(self, fl):
+        self.muted = True
+
+    def end_footnote_list(self, fl):
+        if self.footnote_citations:
+            unresolved = str(list(self.footnote_citations.keys()))
+            self._add_comment(
+                'Warning, unresolved footnote citations exist: '
+                f'{unresolved}'
+            )
```
cmc333333: Should this be a logged warning rather than part of the XML?

api/ombpdf/semdom.py
```diff
@@ -0,0 +1,186 @@
...
+                'marker': str(f.number),
+            })
+            self._add_text(f.text)
+            self._pop_child('footnote')
+            self._pop_cursor()
+        else:
+            self._add_comment(f'Uncited footnote #{f.number}: {f.text}')
+
+    def create_text(self, text):
+        text = text.replace('\n', '')
```
cmc333333: What do you think about replacing newlines with spaces instead? Otherwise `some\nwords` will become `somewords`.

api/ombpdf/tests/test_pdf_to_db.py
```diff
@@ -0,0 +1,31 @@
...
+@pytest.mark.django_db
+def test_error_raised_if_policy_does_not_exist():
+    with pytest.raises(CommandError, message="Policy 'M-14-10' not found."):
+        call_command('pdf_to_db', str(pdf_path))
+
+
+@pytest.mark.django_db
+def test_it_works():
+    policy = mommy.make(Policy, omb_policy_id='M-14-10')
+    assert len(policy.docnode_set.all()) == 0
```
cmc333333: Can we make this a bit more straight forward by using `exists()`?

api/ombpdf/tests/test_semdb.py
```diff
@@ -0,0 +1,39 @@
...
+        Policy, issuance=date(2001, 2, 3), omb_policy_id='M-18-18',
+        title='Some Title', uri='http://example.com/thing.pdf',
+    )
+    result = semdb.to_db(m_14_10_doc, policy).serialize()
+
+    assert_has_paragraph(
+        result,
+        start='Office of Management and Budget (OMB) Memorandum M-12-16, ',
+        end='cash flow and prompt payment to small business subcontractors.'
+    )
```
cmc333333: We can use the DocCursor API to make this a bit more concise:
```python
start = 'Office of Management and Budget (OMB) Memorandum M-12-16, '
end = 'cash flow and prompt payment to small business subcontractors.'
matches = root.filter(lambda n: n.text.startswith(start) and n.text.endswith(end))
assert list(matches)
```

api/ombpdf/management/commands/pdf_to_db.py
```diff
@@ -0,0 +1,35 @@
...
+        if policy is None:
+            raise CommandError(f"Policy '{policy_name}' not found.")
+
+        self.stdout.write(f'Deleting any current document for "{policy}".')
+        DocNode.objects.filter(policy=policy).delete()
+
+        doc = OMBDocument.from_file(fp)
+        to_db(doc, policy)
+
+        self.stdout.write(f'Imported document for "{policy}" into database.')
```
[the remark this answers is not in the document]
toolness: Oh, my understanding was that `self.stdout` was the standard way of outputting stuff in Django commands, in part b/c it's [easy to redirect for testing](https://docs.djangoproject.com/en/2.0/topics/testing/tools/#topics-testing-management-commands).  That said, I'm not actually inspecting the output during my testing...

api/ombpdf/semdom.py
toolness: Heh, that is actually what it originally did--but I eventually changed the behavior to log comments b/c I found that more helpful for debugging purposes.  I think there might be a better name for what it does now, but I'm not sure what that is, so I guess I'll just add a comment documenting its purpose for now.

api/ombpdf/semdom.py
toolness: Oh, I had originally done that so I could use `+=` on the attribute's value without getting an "I can't append `None` to a string" error from Python, but I'll see if we can get away without that.

api/ombpdf/semdom.py
toolness: Hmm, at the time that I wrote this, the XML output was the main output I was paying attention to, so it made the most sense to leave it as a comment, but I'm not sure if a comment *or* a logging message is the best solution right now... For example, it might be best to add some kind of "validation" step later in the pipeline that checks for these sorts of things, because it's ultimately something we want to surface to the end-user who's importing a PDF.

api/ombpdf/semdom.py
toolness: Ah yeah gud point, I will bust that up.

api/ombpdf/tests/test_pdf_to_db.py
toolness: Oops good idea, will do.

api/ombpdf/tests/test_semdb.py
toolness: Ah nice, I will do that and get rid of the ol' `assert_paragraph_exists` thingus.

toolness:
Totally hear you on the snapshot tests bit--a lot of this PR consisted of experimenting until I made something that felt OK, and given the (likely) temporary nature of the XML solution, I decided to just leave it at snapshot tests for the time being.

Regarding the XML intermediary format, I think we are _mostly_ on the same page but would like to note a few things:

1. The XML format is _really_ easy to read and seems to convey most/all of the semantics of the document, so I think it has a lot of value for development and debugging.  Even if the XML format is ultimately abandoned as an _input_ source, I think it's still useful as an _output_ source due to its readability, so e.g. creating a db-to-XML serializer might still be of value.  It might even be useful in the future if someone wants to export a policy with full semantic fidelity (though obviously such a need would have to be validated first).

2. There are a number of aspects of the DB format I found non-obvious, and in the absence of any comprehensive documentation, I found it much easier to simply imitate the XML format instead.  This isn't an argument for the XML format, though, it's rather to say that a bit more documentation on the DB format and how to manipulate and introspect it would be really useful. I'm happy to create some if you think it'd be useful.

cmc333333:
:+1: I think we'll want an Akoma Ntoso (a semi-schema for legal documents that the XML input was based on) serializer at some point. Couldn't hurt to start on it earlier rather than later. Regarding docs, agreed that they're not unified. What do you think the best location for that sort of thing would be? Right now, the model describes its fields in comments and there's a bit more in the corresponding UI `DocumentNode` class.

toolness:
Oh yeah, someone told me about Akoma Ntoso when I first joined this project, but then I forgot about it. It would be cool to use that.

Regarding documentation, how about if I write a text/markdown document with doctests in it?  I really like writing those, possibly too much.

Here's an example of the kind of thing I'm thinking of from CALC, though it happens to be in a docstring rather than its own file:

https://github.com/18F/calc/blob/8ac4c0ac87ae3639cc8a025bfbb129e6262282ae/frontend/steps.py#L9-L111

I could also write the docs in a docstring like that, though then the question becomes where to put it.  I suppose `document.models` seems like a fairly central and discoverable place. Even if we don't end up putting it in a docstring, it might be a good idea to mention the location of the documentation in that file, so that developers can find it easily.

PR #768, opened 2017-12-13
Mobile, in-document navigation

cmc333333:
Building on on #766 and #769, this adds a first pass at mobile, in-document navigation. This component has an open/closed state and is progressively enhanced out of the "open" state for modern JS users.

Some caveats:
* It doesn't try to resolve the heading-under-floating-title issue.
* This'll need to use the fixed sticky library from #762 once it's merged.
* Code climate rightfully points out some deeply-nested styles. I'd prefer if we didn't have these, but I also don't see a nice way out.

Looks like
## Before open
screen shot 2017-12-14 at 10 09 53 am

## Opened
screen shot 2017-12-14 at 10 10 08 am

## No-JS/old browser
screen shot 2017-12-14 at 10 10 38 am

### DIFF
--- a/ui/__tests__/components/document/floating-nav.test.js
+++ b/ui/__tests__/components/document/floating-nav.test.js
@@ -32,4 +32,12 @@ describe('<FloatingNav />', () => {
     expect(button.hasClass('active')).toBe(true);
     expect(button.text()).toBe('✕');
   });
+  it('passes a "close" onClick handler to the DocumentNav', () => {
+    const result = shallow(<FloatingNav />);
+    result.setState({ open: true });
+    const onClick = result.children().find(DocumentNav).prop('onClick');
+    expect(result.state('open')).toBe(true);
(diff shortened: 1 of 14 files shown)


cmc333333:
Updated and ready for review! Heads up @rtwell -- I may need to do any cleanup in a separate PR so we can get a shared base faster.

rtwell:
any chance we could hide the lil vertical bar on open? It's the separator for "Jump to section" and "Search this policy", seen here: 
image

cmc333333:
Note that the commits specific to this feature start with 08d9b2d. Since the other PRs have been merged, let me know if it'd be easier to review rebased

cmc333333:
I'm going to go ahead and merge this myself due to time constraints and folks being out. Rebasing first.

cmc333333:
Thanks for looking it over @rtwell !

PR #769, opened 2017-12-13
Use smoothscroll within document navigation.

cmc333333:
Smooth scrolling not only looks nicer, but it also hacks around a
react-stickynode bug which was preventing the nav from displaying in some
cases.

Looks like a smoother version of
smoothscroll

### DIFF
--- a/ui/components/document/navigation/nav-link.js
+++ b/ui/components/document/navigation/nav-link.js
@@ -1,14 +1,22 @@
+import jump from 'jump.js';
 import PropTypes from 'prop-types';
 import React from 'react';
 import { connect } from 'react-redux';
 
 import Link from '../../link';
 
+function scrollTo(anchor) {
+  return (e) => {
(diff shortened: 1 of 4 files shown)


ui/components/document/navigation/nav-link.js
```diff
@@ -1,14 +1,22 @@
...
+    jump(`#${anchor}`);
+  };
+}
+
 export function NavLink({ active, children, identifier, title }) {
   const className = `document-nav-heading${active ? ' active' : ''}`;
   return (
     <li>
-      <Link className={className} href={`#${identifier}`}>
+      <Link className={className} href={`#${identifier}`} onClick={scrollTo(identifier)}>
```
tadhg-ohiggins: Everything old is new again

cmc333333:
Thanks, @tadhg-ohiggins, I'll rebase and merge.

PR #775, opened 2017-12-18
Standardize "Policy" object across three pages

cmc333333:
We had defined a `Policy` object for use on the policy search results page. This defines that object's fields a bit more closely and uses it on the homepage and in document metadata.

This came up while adding page titles.


toolness:
Ok I am going to merge this so I can more easily comment on #776.

PR #776, opened 2017-12-18
Add a page title for every page

cmc333333:
Building off #775 ([diff](https://github.com/18F/omb-eregs/compare/699-policy-object...699-page-titles)), this ensures every page sets a title. The only interesting bit here is that the "document" page sets its title to the policy's title-with-number.

### DIFF
--- a/ui/components/header-footer.js
+++ b/ui/components/header-footer.js
@@ -83,7 +83,11 @@ export default function HeaderFooter({ children, showSearch, wrapperClassName })
         { faviconTags() }
         <meta name="viewport" content="width=device-width, initial-scale=1" />
       </Head>
-      { pageTitle() }
+      {
+        /* Set a default page title. This can be overridden if pageTitle()
+          is also called in child components */
+        pageTitle()
+      }
       <div className="page-header-wrapper">
         <Disclaimer key="disclaimer" />
         <Navbar key="navbar" showSearch={showSearch} />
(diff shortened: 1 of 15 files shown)


ui/components/header-footer.js
```diff
@@ -67,30 +69,30 @@ function faviconTags() {
...
+        <script
+          async
+          type="text/javascript"
+          id="_fed_an_ua_tag"
+          src="https://dap.digitalgov.gov/Universal­Federated­Analytics­Min.js?agency=EOP&subagency=OMB"
+        />
+        { faviconTags() }
+        <meta name="viewport" content="width=device-width, initial-scale=1" />
+      </Head>
+      { pageTitle() }
```
toolness: Interesting, so `<Head><title key="title">` is defined here, but it's also potentially "overridden" by the `{children}` embedded at line 91 as per next's documentation on [populating `<head>`](https://github.com/zeit/next.js/#populating-head), is that right?

I wonder if it's worth mentioning this in a comment above this line? Up to you really; I'm sure it's something that next.js veterans find idiomatic, while folks unfamiliar with it might mistake as a bug at first.

cmc333333: You got it! I'll add a comment to make it clearer.

cmc333333:
Thanks @toolness! I'll update with a comment then merge. Next.js's version is very similar to (but ever so slightly different than) [react-helmet](https://github.com/nfl/react-helmet), which you may encounter elsewhere.

PR #777, opened 2017-12-18
Resolve 500 error when search 404ed.

cmc333333:
When a user doesn't have JS, the topic search falls back to a text field which
is passed around via POSTs. If the backend API returned a 404 (as is the case
with bad page numbers), we weren't passing that to the user.

To see it in action, visit: 
```
/search-redirect/topics/?insertParam=requirements__topics__id__in&redirectRoute=policies&q=security&redirectQuery__requirements__req_text__search=&page=44
```

### DIFF
--- a/ui/__tests__/util/api/queries.test.js
+++ b/ui/__tests__/util/api/queries.test.js
@@ -6,6 +6,7 @@ import {
   policiesData,
   redirectQuery,
   requirementsData,
+  searchRedirectData,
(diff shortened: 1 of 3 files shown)


cmc333333:
Ah, looks like it'll get fixed more permanently in [0ff2b6781ec8817b65ef881d22d55024eeb19134]. I'll cherry-pick that and merge.

PR #786, opened 2017-12-19
Visual tweaks to document

cmc333333:
Per #718, this:
* Removes gold from h4 headings
* Uses the same font size/line height for numbered lists
* Adjusts spacing between headings

It also adds some variables/mixins to start the process of standardizing our font sizes and spacing, per discussion last sprint with @rememberlenny and @rtwell 

Looks like

screen shot 2017-12-19 at 10 42 43 am
screen shot 2017-12-19 at 10 42 29 am
screen shot 2017-12-19 at 10 43 03 am

### DIFF
--- a/ui/css/layout/_typography.scss
+++ b/ui/css/layout/_typography.scss
@@ -114,8 +114,7 @@ h6,
 
 p {
   @include font-serif();
-  font-size: 1rem;
-  line-height: 1.69rem;
+  @include size-regular();
 }
 
 @media #{$on-desktop} {
--- a/ui/css/layout/_typography.scss
+++ b/ui/css/layout/_typography.scss
@@ -94,20 +94,11 @@ h3,
   line-height: 1.5rem;
 }
 
-.document-container {
-  h4,
-  .h4 {
-    color: $color-ochre;
-  }
-}
-
 h4,
 .h4 {
   font-size: 1rem;
   line-height: 1.5rem;
 }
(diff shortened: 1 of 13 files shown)


ui/css/layout/_typography.scss
```diff
@@ -123,8 +114,7 @@ h6,
 
 p {
   @include font-serif();
-  font-size: 1rem;
-  line-height: 1.69rem;
+  @include size-regular();
```
cmc333333: Note that this is a slightly different line-height, but the new value matches the Sketch file.

ui/css/layout/_typography.scss
```diff
@@ -94,21 +94,12 @@ h3,
   line-height: 1.5rem;
 }
 
-.document-container {
-  h4,
-  .h4 {
-    color: $color-ochre;
```
toolness: According to my `grep`, this might be the only place `$color-ochre` is used outside of its definition in `_colors.css`--should we remove its definition?

cmc333333: Good question; we have in the past, but let's verify with @rtwell that the color won't be used somewhere else soon.

cmc333333: @rtwell requests we keep it

cmc333333:
Thanks @toolness and @rtwell. I'll rebase and merge.

PR #794, opened 2017-12-19
Support (eventual) inline rich text markup in footnotes

### LINKED ISSUE #782: ombpdf footnote annotations shouldn't contain full text of footnote
(This issue has been migrated from https://github.com/18F/omb-pdf/issues/34.)

Since footnote annotations can contain rich text, e.g. strong/emphasized text, we shouldn't make the footnote text part of the annotation. Instead, just rely on the client to use the actual text of the lines, like it does with list items.

toolness:
~~This fixes the core problem presented by #782, though it doesn't actually resolve that issue itself, because it still keeps the full text of a footnote as part of the `OMBFootnote` annotation. Actually fixing the latter will take some effort because the footnote tests rely on it so much, but it's not that high a priority.~~

This fixes #782.

Note that we don't actually _have_ any inline rich text markup in `semtree.py`, so this doesn't actually do a whole lot, but eventually we will support inline rich text markup, so it's nice to know it'll automatically work okay when it's in footnotes.

The changes in the snapshot files are due to us delegating the content of footnotes to the same logic we use for any block content, like paragraph text.  Because this logic is slightly different from the logic for calculating the text of footnotes--largely involving the handling of whitespace--the snapshots are different.

### DIFF
--- a/api/ombpdf/footnotes.py
+++ b/api/ombpdf/footnotes.py
@@ -3,7 +3,7 @@ from decimal import Decimal
 from textwrap import TextWrapper
 
 from . import util
-from .document import OMBFootnote, OMBFootnoteCitation
+from .document import OMBFootnote, OMBFootnoteCitation, OMBFootnoteMarker
 from .horizlines import get_horizontal_lines
 
 NUMBER_RE = re.compile(r'[0-9]')
@@ -107,6 +107,8 @@ def annotate_footnotes(doc):
                 if match:
                     finish_footnote()
                     footnote, desc = match.groups()
+                    for char in line[:len(footnote) + 1]:
+                        char.set_annotation(OMBFootnoteMarker(int(footnote)))
                     curr_footnote = [int(footnote), desc, [line]]
                 elif curr_footnote:
                     curr_footnote[1] += chars
--- a/api/ombpdf/footnotes.py
+++ b/api/ombpdf/footnotes.py
@@ -8,7 +8,7 @@ from .horizlines import get_horizontal_lines
 
 NUMBER_RE = re.compile(r'[0-9]')
 
-FOOTNOTE_RE = re.compile(r'([0-9]+) (.+)')
+FOOTNOTE_RE = re.compile(r'([0-9]+ )(.+)')
 
 # Maximum distance the horizontal line separating a page's footnote section
 # can be from the top of the first footnote.
@@ -106,10 +106,11 @@ def annotate_footnotes(doc):
                 match = FOOTNOTE_RE.match(chars)
                 if match:
                     finish_footnote()
-                    footnote, desc = match.groups()
-                    for char in line[:len(footnote) + 1]:
-                        char.set_annotation(OMBFootnoteMarker(int(footnote)))
-                    curr_footnote = [int(footnote), desc, [line]]
+                    marker, rest = match.groups()
+                    number = int(marker.strip())
+                    for char in line[:len(marker)]:
+                        char.set_annotation(OMBFootnoteMarker(number))
+                    curr_footnote = [number, rest, [line]]
                 elif curr_footnote:
                     curr_footnote[1] += chars
                     curr_footnote[2].append(line)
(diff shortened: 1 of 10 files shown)


api/ombpdf/footnotes.py
```diff
@@ -107,6 +107,8 @@ def finish_footnote():
                 if match:
                     finish_footnote()
                     footnote, desc = match.groups()
+                    for char in line[:len(footnote) + 1]:
```
cmc333333: This line threw me for a bit:

I initially thought interpreted "footnote" to mean the whole line, which would make the length redundant. Perhaps we should swap the variable name to something like "number" or "marker" to be explicit ("marker" may be more consistent with the rest of the script)? Unfortunately "footnote" can mean different concepts at different times.

Following that, I didn't understand where "+ 1" came from. I *think* it's ensuring the space is grabbed. But why would we want the space to be tagged as a footnote marker?

api/ombpdf/semtree.py
```diff
@@ -212,8 +210,10 @@ def process_footnotes(self):
                 anno = line.annotation
                 if anno != curr_footnote:
                     curr_footnote = anno
-                    self.writer.create_element(Footnote(anno.number,
-                                                        anno.text))
+                    self.close_all_blocks()
+                    self.open_new_block(Footnote(anno.number), anno)
+                self.process_line(line)
+            self.close_all_blocks()
```
cmc333333: Just flagging this line as a departure from what happened before (when there was no need to close any block elements as none were introduced). Is there any situation where closing all of the blocks could cause a problem?

api/ombpdf/footnotes.py
```diff
@@ -116,19 +118,14 @@ def finish_footnote():
...
                         # Here we pull out the previous footnote and continue
                         # adding to it.
-                        curr_ft = footnotes.pop()
-                        curr_obj = curr_ft[0]
-                        curr_num = curr_obj[0]
-                        curr_txt = curr_obj[1]
-                        curr_lines = curr_ft[1]
+                        (curr_num,), curr_txt, curr_lines = footnotes.pop()
                         curr_footnote = [curr_num, curr_txt + chars, curr_lines
                                          + [line]]
```
cmc333333: Only tangential to this PR, but should we use the same data type (now a triple, see `footnote_plus`) here rather than making a new type here (a three-element list)? Looks like it's mutated ~115, but that could be an namedtuple `._replace()`

api/ombpdf/footnotes.py
```diff
@@ -116,19 +118,14 @@ def finish_footnote():
...
             else:
                 finish_footnote()
     finish_footnote()
-    # Return the OMBFootnote list after applying strip() to the text:
-    footnote_objs = [OMBFootnote(f[0][0], f[0][1].strip()) for f in footnotes]
-    return footnote_objs
+
+    return [(footnote, text.strip()) for footnote, text, _ in footnotes]
```
cmc333333: This is much cleaner than the index-based access, nice job!

api/ombpdf/footnotes.py
toolness: Yeah, this code is not very nice, blurg.  I would like to use a named tuple here but I feel weird about making named tuples for intermediate data structures that are used internally by functions... Then again, this whole function is a bit over-long and could be broken up into multiple ones, and improving readability is always a good thing.

I did not know that `_replace()` existed, thanks for suggesting it!

api/ombpdf/semtree.py
toolness: Not currently, as footnotes are essentially a top-level block element, but there are other places we're using this method that I am concerned about--I suspect we may need to visit the `close_all_blocks()` method in general if/when the document model becomes more complex/nested.

api/ombpdf/footnotes.py
toolness: Good points, I just refactored the code in 4d2ad60d5a42f9bbbf7530387d802a1c55eba273 to make this more readable.

The `+ 1` was indeed the space following the footnote; it was a bit of a toss-up as to whether to classify it as part of the marker, and I decided to do it so that it wasn't mistaken to be part of the actual footnote text by downstream clients.  That said, I've moved it into the captured group of the marker in the regexp so that it's not a confusing `+ 1` anymore, at least.

PR #795, opened 2017-12-19
First pass at Akoma Ntoso renderer

cmc333333:
This adds the ability to view an XML version of the document tree, accessible by adding `?format=akn` to the document endpoint. It's close(-ish) to Akoma Ntoso and should satisfy some of the needs @toolness described in #767


toolness:
Hmm, you are on vacation now so I'm just gonna merge this and we can talk about my noobish questions when you get back.

cmc333333:
This does indeed look a lot like the ad-hoc XML format -- that format was a sort of simplified AKN. This is missing some of the document meta data, but should be pretty close on the document content (we've tried to use AKN tags for our node types).

`p` and `para` are a little different; the latter is structural (and can be nested inside anything/nest anything in AKN) while the former's closer to the HTML concept (only containing inlines). AKN actually uses `paragraph`, but they [shorten it](http://docs.oasis-open.org/legaldocml/akn-nc/v1.0/csprd01/akn-nc-v1.0-csprd01.html#_Toc417907268) when constructing element ids. Rather than doing that conversion, we just use `para` everywhere. I figure we'll realign a bit once AKN is a closer goal.

PR #796, opened 2017-12-20
Remove CDN'd version of KaTeX css.

cmc333333:
We can embed the CSS directly from our local node package, which ensures the
CSS and JS stay in sync.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ node_modules
 reqs/static/js/bundle.js
 ui-dist
 ui/static/font/fontawesome-webfont*
+ui/static/font/KaTeX*
 ui/static/styles.css*
 jest*
 ui/lcov-report/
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -9,6 +9,7 @@ node_modules
 reqs/static/js/bundle.js
 ui-dist
 ui/static/font/fontawesome-webfont*
+ui/static/font/KaTeX*
```
toolness: Er, where are these files coming from?  I'm not seeing any logic in this code that creates these files, so I'm uncertain, but I'm probably just missing something obvious.

cmc333333: Sure thing -- assuming I understand it correctly, this stems from webpack's `file-loader`. When building the styles, if webpack encounters a font referenced by the css, it'll pull it out and place it in the `static/font` directory. We don't want to commit these to the repo, though, as they're coming from a library.

cmc333333:
Thanks @toolness

PR #797, opened 2017-12-20
Add a BrowsableAkomaNtosoRenderer

toolness:
This minorly builds on #795 by making the AKN renderer easy to browse, which I think could be helpful for debugging. Currently, clicking the "akn" format in the browsable API (under the "GET" button/popup) triggers a download, which one then has to find and open in their text editor, whereas this PR adds an "akn-api" format that [trivially customizes `BrowsableAPIRenderer`](http://www.django-rest-framework.org/api-guide/renderers/#customizing-browsableapirenderer) to present the XML response in a syntax-highlighted web view, similar to how the normal JSON responses are presented:

> screenshot-2017-12-20 tree omb policy library api beta

It also overrides `rest_framework/api.html`'s [`{% bootstrap_theme %}` block](http://www.django-rest-framework.org/topics/browsable-api/#overriding-the-default-theme) with a bit of well-targeted CSS to ensure that readers don't have to horizontally scroll to see all the content. While I was at it, I also changed the page's title and branding to reflect the OMB Policy API rather than Django Rest Framework.

### DIFF
--- a/api/document/renderers.py
+++ b/api/document/renderers.py
@@ -60,3 +60,10 @@ class AkomaNtosoRenderer(renderers.BaseRenderer):
     def render(self, data, media_type=None, renderer_context=None):
         as_xml = node_to_xml(data)
         return etree.tostring(as_xml, encoding=self.charset, pretty_print=True)
+
+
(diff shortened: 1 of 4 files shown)


toolness:
GET BACK TO YOUR VACATIONING

PR #798, opened 2017-12-21
[WIP EXPERIMENT] Prosemirror fun

toolness:
I'm just trying to figure out how ProseMirror works here, making my experimentation visible as a WIP PR. This probably won't ever get merged--once I understand how things work, I'll probably just start a fresh PR that doesn't cut any corners like I'm doing here.

## Quick start

1. Checkout the PR.
2. Run `manage.py import_xml_doc example_docs/m_16_19_1.xml M-16-19`.
3. Run `docker-compose up dev-api`.
4. Visit `http://localhost:8001/pdf/prosemirror-fun`.

You can then play with the editor and open the browser console to see if the [janky ad-hoc test suite](https://github.com/18F/omb-eregs/blob/prosemirror-fun/api/ombpdf/js/prosemirror-fun/test.js#L1-L3) passed.

You can also press <kbd>Ctrl</kbd>+<kbd>S</kbd> while in the editor to save your changes back to the database. They'll then be reflected in the API and viewer front-end.

## Open questions

I am confused by the structure of the JSON schema coming out of the database and have a bunch of questions about it, such as:

* What is the motivation for the distinction between a node's `content` and `children`? ProseMirror doesn't make a distinction like this, which makes translating between the two a bit ambiguous.
* ~~Why is it possible for paragraphs to have non-[phrasing-content](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Phrasing_content) elements, like lists and tables, as children? This makes it very difficult to represent paragraphs as standard `<p>` elements; for now I'm just creating `<p>` elements that contain the `para`'s `content`, and creating the `para`'s `children` as adjacent sibling nodes.~~ Paragraphs in OMB policy terms aren't like HTML paragraphs; for example, note that in ATF eRegs, there is terminology like "Paragraph A, sub-paragraph B".
* Why is lots of content duplicated? For instance, a `para` often has its text stored _twice_, once as a `text` property and again as a child under `content`. And `footnote_citation` seems to store the full text of its related footnote under its `footnote_node` property, but that node is repeated as a child of the parent element that the citation is in, too.

## To do

- [x] When converting the ProseMirror doc to a db doc, deal with marks somehow.
- [x] Implement an API endpoint that takes the JSON structure of a db doc and writes it to the database, so that policy edits can be saved.


toolness:
Here's an in-progress screenshot of what this is like so far:

> screenshot-2017-12-22 prosemirror fun

The pink boxes represent unimplemented content. I'm thinking that until we've got everything implemented and fully editable, users will be able to edit everything except what's in the pink boxes; perhaps clicking on the pink boxes might open an XML or JSON editor allowing developers or other technical staff to manually edit the content.

toolness:
Ok, at this point everything except tables and math can be edited in a rudimentary way:

> screenshot-2017-12-28 prosemirror fun

toolness:
Some things I've learned:

* ProseMirror comes with default actions for very common kinds of content. I originally made lists mimic the structure that they're stored in the database / rendered in our document view on the front-end, but this proved challenging to edit because ProseMirror didn't have any out-of-the-box actions that could be used to easily manipulate the lists. So in 33398f7 I changed the schema to be a modified version of the one from [prosemirror-schema-list](https://github.com/ProseMirror/prosemirror-schema-list)--that is, I modified it to account for different types of ordered lists, e.g. numbered vs. lettered--which then allowed me to use all the pre-existing editing commands that came with that package.

* I think the document schema we present to the user has UX implications. For instance, using a hierarchical section structure in which there's only really one kind of "heading" element is a very convenient semantic representation for our documents, but it might not be a very convenient representation for users to _edit_.  This is because it potentially implies a suite of editing actions to _manipulate_ the section hierarchy, which may involve more cognitive load than simply presenting users with a "flattened" representation of the document in which they have the freedom to choose different heading levels (a convention they are likely already familiar with from Google Docs and Microsoft Word), and in which we do the work of "unflattening" the document before storing it in the database. (This is actually how the pdf-to-db pipeline works, coincidentally.)

toolness:
Ok, I think this is at the point where the experiment can end. It's possible to visit `/pdf/prosemirror-fun`, make edits to M-16-19, press <kbd>Ctrl</kbd>+<kbd>S</kbd> and then have those edits saved back to the db and see them reflected via API endpoints and the regular front-end viewer.

Some big limitations of this experiment:

* Currently it only supports content with at most one ProseMirror mark on it. Right now that's easy because there are only really two kinds of marks, footnote citations and hyperlinks, which can never overlap. However, once we start having more marks, like highlighting something as a requirement, this will likely change.

* The JSON endpoint right now is manually coded and does very little validation (it's actually largely just a JSON version of the existing XML importer).  In production, we should use Django REST Framework to do this, so that we can take advantage of its fancy validation and stuff.

* I'm not really sure what to do about the `identifier` property of `DocNode`s.  Right now we just throw them away when saving edits to the database, which means they get regenerated using the default algorithm, which means they end up with values like `policy_1__sec_1__heading_1`.  Since *all* identifiers already seem to have such auto-generated values, this doesn't seem like a big deal, but I want to make sure this approach is OK.

* Because we're using ProseMirror's out-of-the-box list schema, which I described [earlier](https://github.com/18F/omb-eregs/pull/798#issuecomment-354304843), list items can't currently be renamed.  Making them rename-able without rewriting lots of ProseMirror's list code might be possible if we use CSS cleverly.

* Because we're using ProseMirror's out-of-the-box paragraph schema, paragraphs can't actually have children, or markers, or sub-paragraphs.  When saving to the database, we basically take adjacent ProseMirror siblings following a paragraph, and make them children of the database `DocNode` representation of the paragraph rather than siblings. We *could* reflect the same hierarchy in ProseMirror, but because this adds extra hierarchy, it will likely make the UI more complicated.

* As discussed during the stand last week, and as I hinted at [earlier](https://github.com/18F/omb-eregs/pull/798#issuecomment-354304843), we should probably "flatten" the heading/section structure when presenting the document to the user in ProseMirror so they can easily edit it, while also showing them a preview of the document outline so they know how their heading choices affect it.

I'm also not entirely sure what the best strategy is for building this in production, but more on that later.

toolness:
Ok, I think that our production functionality mirrors this PR closely enough that we probably won't need to refer to it much anymore, so I am going to close it.  (Obviously we can rez it again if needed!)

PR #799, opened 2017-12-29
Skip intermediary XML layer from pdf

cmc333333:
Rather than writing to XML then importing from that, this has the pdf import write directly to the database. Along the way, it also enables ombpdf tests from the py.test and ptw docker commands.

### DIFF
--- a/api/document/tests/tree_test.py
+++ b/api/document/tests/tree_test.py
@@ -128,6 +128,37 @@ def test_parent():
         == 'root_0__sect_2__par_a'
 
 
+def test_ancestors():
+    root = tree.DocCursor.new_tree('root', '0')
+    root.add_child('sect')
+    sect2 = root.add_child('sect')
+    pa = sect2.add_child('par', 'a')
+    pa1 = pa.add_child('par', '1')
+    pa1.add_child('par', 'i')
+    sect2.add_child('par', 'b')
+
+    assert [n.identifier for n in root.ancestors()] == []
+    result = [n.identifier
+              for n in root['sect_2']['par_a']['par_1'].ancestors()]
+    assert result == [
+        'root_0__sect_2__par_a',
+        'root_0__sect_2',
+        'root_0',
+    ]
+    result = [
+        n.identifier
+        for n in root['sect_2']['par_a']['par_1']['par_i'].ancestors(
+            lambda n: n.node_type == 'par')
+    ]
+    assert result == ['root_0__sect_2__par_a__par_1', 'root_0__sect_2__par_a']
+    result = [
+        n.identifier
+        for n in root['sect_2']['par_a']['par_1']['par_i'].ancestors(
+            lambda n: n.node_type == 'sect')
+    ]
+    assert result == ['root_0__sect_2']
+
+
 @pytest.mark.django_db
 def test_create_save_load():
     """Integration test that shows we get the same results when converting
@@ -195,3 +226,15 @@ def test_default_policy():
     assert root.policy == policy1
     assert root['sec_1'].policy == policy1
     assert root['sec_2'].policy == policy2
+
+
+def test_jump_to():
+    root = tree.DocCursor.new_tree('root', '1')
+    root.add_child('sec')
+    root['sec_1'].add_child('para', 'a')
+    root.add_child('sec')
+    para_a = root['sec_2'].add_child('para', 'a')
+
+    assert root.jump_to('root_1').identifier == 'root_1'
+    assert para_a.jump_to('root_1').identifier == 'root_1'
+    assert para_a.jump_to('root_1__sec_1').identifier == 'root_1__sec_1'
--- a/api/document/tests/tree_test.py
+++ b/api/document/tests/tree_test.py
@@ -17,28 +17,28 @@ def test_new_tree():
 
 def test_item_access():
     root = tree.DocCursor.new_tree('root', '0')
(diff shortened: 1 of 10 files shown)


api/ombpdf/semdb.py
```diff
@@ -1,12 +1,153 @@
...
+    def begin_footnote_list(self, fl):
+        self.cursor_stack = self.cursor_stack[:1]
+
+    def end_footnote_list(self, fl):
+        no_nodes = [key for key, value in self.footnote_citations.items()
+                    if not value.footnote_node]
+        if no_nodes:
+            logger.warning('Unresolved footnote citations exist: %s', no_nodes)
+            for no_node_cite in no_nodes:
+                del self.footnote_citations[no_node_cite]
```
toolness: Hmm, do you think we should modify our DB schema to support unresolved footnote citations if the policy is still in some kind of draft/unpublished state?  my concern is that this removal of information could result in user confusion--e.g., a document with an unresolved footnote citation for "4" will likely be easier to fix than a document that just skips from footnote 3 to 5 without any mention of a "4" at all.

Anyways, I guess this question might be outside the scope of this PR, since leaving it unaddressed doesn't actually lose us any functionality (I assume the current XML intermediary solution has the same problem).

cmc333333: I think I agree with you in general, that adding more meta data around parse errors would be useful. We can probably grep around for warning log messages and turn them into some sort of annotation. I also agree that that's a bit outside of scope for this PR.

In this particular case, I want to emphasize we're dropping the citation annotation rather than the footnote itself. The text for the citation will still exist, and if its corresponding footnote (perhaps misparsed) is somewhere in the document, it'd still be placed at the bottom of the document. I *think* that's low risk at this point; the information we're losing is either a false positive or it'll be relatively easy to reconstruct in the editor.

api/ombpdf/semdb.py
```diff
@@ -1,12 +1,153 @@
...
-    xml = etree.fromstring(dom.toxml())
-    root = import_xml_doc(policy, xml, ignore_preamble=True)
-    return root
+    writer = DatabaseWriter(policy)
+    builder = semtree.SemanticTreeBuilder(doc, writer)
+    builder.build()
+
+    writer.cursor.nested_set_renumber()
+    # Account for a Django bug which prevents us from saving these
+    # directly due to saving the DocNodes in a bulk_insert
```
toolness: Is this Django bug recorded in their issue tracker or anything? I thought it might be useful to link to that, both for additional context, but also so that someday we might follow it to find it fixed, and thus be able to remove this workaround.

api/document/tests/tree_test.py
```diff
@@ -128,6 +128,37 @@ def test_parent():
         == 'root_0__sect_2__par_a'
 
 
+def test_ancestors():
+    root = tree.DocCursor.new_tree('root', '0')
+    root.add_child('sect')
```
toolness: I realize this is only a test, and that `'sect'` might as well be `'foo'`, but insofar as tests function a bit like documentation, do you think this should be `'sec'` just to stay consistent with the names we use in our actual schema?  I know I personally have become confused about whether our real schema's paragraphs are called `'par'` or `'para'` due to inconsistencies like this.

Um, speaking of which, should this test use `'para'` instead of `'par'` too? 😅

api/ombpdf/semdb.py
```diff
@@ -1,12 +1,153 @@
...
+        self.cursor_stack = [PDFAwareCursor.new_tree('policy', policy=policy)]
+        self.footnote_citations = {}
+
+    @property
+    def cursor(self):
+        return self.cursor_stack[-1]
+
+    @property
+    def sec_level(self):
+        return self.cursor.identifier.split('_').count('sec')
```
toolness: I'm a little concerned that this code might be a bit too "clever"--at least, it took some thinking for me to mentally verify that this would actually work, and not have weird edge cases.  I can't really think of any easy way to make it more readable/less clever though.

api/ombpdf/semdb.py
```diff
@@ -1,12 +1,153 @@
...
+        li = self.cursor.add_child('listitem')
+        self.cursor_stack.append(li.add_child('para'))
+        self.cursor.pdf_node = p
+
+        parent_list = next(
+            self.cursor.ancestors(lambda n: n.node_type == 'list'),
+            None,
+        )
+        if not parent_list:
+            logger.warning('Adding list item, but no list parent')
```
toolness: I remember you mentioning this to me on Slack--is it still a problem?  If so, we should file an issue, because it'd definitely be a bug in `semtree.py`!

api/document/tests/tree_test.py
cmc333333: That's a good point; I'll search-and-replace.

api/ombpdf/semdb.py
cmc333333: I'll try to track that down and add a comment.

api/ombpdf/semdb.py
cmc333333: I can try to expand it out a bit more and label the intermediate steps.

api/ombpdf/semdb.py
cmc333333: Indeed it is; created #803

cmc333333:
Thanks for reviewing @toolness !

PR #800, opened 2017-12-29
Display list item titles, if present.

cmc333333:
We've seen multiple policies where individual list items have an associated
title/heading. This modifies our renderers to display that, if present.

Looks like (after mocking out some data):
screen shot 2017-12-29 at 4 09 12 pm


cmc333333:
Thanks @rememberlenny !

PR #801, opened 2018-01-03
Add script for scraping & loading memoranda.

cmc333333:
OMB keeps a list of all their memoranda. Rather than manually downloading and
loading each PDF into the database, we can scrape that list and load them all
at once. This'll allow us to verify them much more quickly and make importing
them into an empty database much easier.

Using current db importer
* 92 imports
* 54 exceptions

Using the #799 importer
* 110 imports
* 36 exceptions


cmc333333:
Thanks @toolness!

Issue #410, opened 2017-06-20
Document type annotation strategy


cmc333333:
Per https://github.com/18F/omb-eregs/pull/403#discussion_r122509801

>  I do think that regardless of whether or not we expect total mypy coverage, it's better to have type annotations on Python 3 projects than not to have them.

We should mention this policy in the README or similar.

PR #802, opened 2018-01-03
Run mypy on whitelisted files.

toolness:
This attempts to productionalize some experimentation I did in #798 by running [mypy](http://mypy.readthedocs.io/) on a whitelist of our source files (stored in `api/mypy-files.txt`).  This allows us to easily opt-in to type checking without slowing us down.

Currently the mypy settings are _very_ lax, which I've found is generally good for legacy codebases and code that uses third-party libraries that lack type annotations (like Django), since it makes it possible to ["sprinkle in" type checking in a targeted way](https://github.com/18F/calc/pull/1505#issuecomment-288855201), e.g. based on where it'd be most useful.

If this seems amenable, I can add documentation for it to `README.md` in which I also explain our approach, which could help with #410.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,6 +60,11 @@ jobs:
             COMPOSE_FILE: docker-compose-circle.yml
           command: docker-compose run --rm flake8
       - run:
+          name: API Mypy
+          environment:
+            COMPOSE_FILE: docker-compose-circle.yml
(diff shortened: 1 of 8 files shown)


toolness:
Yeah, mypy's lack of output on success is weird to me, but I guess flake8 works the same way? Ah well.

PR #805, opened 2018-01-04
[WIP] Make API use a Docker container w/ both Python and Node.

toolness:
In preparation for productionalizing #798, this attempts to forge the API's Docker container into one that supports both Node and Python. It also modifies ombpdf's webpack build process to be a build process for API in general.

## Notes

* I've renamed the `ombpdf-js-bundler` docker container to `api-webpack`, since it's building stuff for the whole API now.

* I've renamed the docker-compose commands `pybash` and `nodebash` to `apibash` and `uibash`, which is probably what they should've originally been named anyways.

* The `Dockerfile` is in `api/.docker`. This isn't just for organizational purposes, it's also so the [Docker build context stays tiny](https://docs.docker.com/engine/reference/builder/#usage):

  > The build is run by the Docker daemon, not by the CLI. The first thing a build process does is send the entire context (recursively) to the daemon. In most cases, it’s best to start with an empty directory as context and keep your Dockerfile in that directory. Add only the files needed for building the Dockerfile.

## To do

- [ ] Figure out how this works in deployment.
- [x] Make sure CircleCI is ok with this new arrangement.
- [ ] Figure out what to do, if anything, about the naming of the individual UI-related docker-compose commands like `npm` and `webpack`, since their names are now ambiguous. We could prefix them with `ui-` for instance, and then make separate `api-` prefixed variants too.
- [ ] I'm concerned about the possibility of race conditions between competing `activate_then`'s when running `docker-compose up dev-api`, as both the `api-webpack` and `dev-api` containers will be running it simultaneously, on their own separate instances of a layered file system, which makes my brain hurt. In CALC we got around this kind of problem by explicitly creating a `docker-update.sh` script that developers are expected to run whenever they `git pull` or switch branches, which ensures that only one process is configuring dependencies at any given time. But obviously this also makes development a bit more manual and cumbersome too.


api/.docker/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM python:3.6.2
+
+RUN curl -sL https://deb.nodesource.com/setup_8.x | bash -
+
+# Note that we want postgresql-client so 'manage.py dbshell' works.
+RUN apt-get update && \
+  apt-get install -y nodejs postgresql-client
```
toolness: I can take out `postgresql-client` if you want, but I thought it'd be useful just so `manage.py dbshell` works.

api/.docker/activate_then
```diff
@@ -22,4 +24,17 @@ else
...
+  echo "Installing deps-ok"
+  npm install --quiet deps-ok
+fi
+
+./node_modules/.bin/deps-ok &> /dev/null
+if [ $? -ne 0 ]; then
+  echo "Installing Node dependencies"
+  npm install --quiet
+fi
+
```
toolness: It was most convenient to just add the `deps_ok_then` script to the end of the `activate_then` script, but we could keep them separate if you want, or we could rename `activate_then` to `activate_and_deps_ok_then` or something else.

api/omb_eregs/settings.py
```diff
@@ -188,6 +188,9 @@
 
 STATIC_URL = '/static/'
 STATIC_ROOT = os.environ.get('TMPDIR', '.') + '/static/'
+STATICFILES_DIRS = [
+    os.path.join(BASE_DIR, 'webpack-static'),
+]
```
toolness: Since the webpack-built stuff can apply to any Django app, I thought it might be best to just have a separate staticfiles directory for all the webpack stuff.  We might be able to configure webpack to output the different bundles into the `static` directories of the respective apps that they belong to, though, if that's preferable.

toolness:
Ok, superseding this with #807!

PR #806, opened 2018-01-04
Resolve a handful of pdf importing exceptions

cmc333333:
These prevent explosions when importing docs that:
* Don't have a clear left margin
* Have footnote citations, but no footnote list
* Have no text (i.e. image-only)


toolness:
The codeclimate warning isn't relevant to the scope of this PR (it's a pre-existing problem with pre-existing code that was moved) so don't worry bout it!

cmc333333:
Thanks @toolness! I took a few stabs at simplifying that function to reduce the CC, but none of them felt clearer.

PR #807, opened 2018-01-04
Convert ombpdf-js-bundler into api-webpack

toolness:
This supersedes #805 by just turning our `ombpdf-js-bundler` container into an `api-webpack` container, as it will be responsible for creating JS bundles for the API in general (not just the ombpdf app).

- [x] Ensure that CircleCI builds the bundle and deploys it properly.
- [ ] Figure out what to do, if anything, about the naming of the individual UI-related docker-compose commands like `npm` and `webpack`, since their names are now ambiguous. We could prefix them with `ui-` for instance, and then make separate `api-` prefixed variants too.

Even though the ombpdf JS bundle is for development-only while the future bundles won't be, I didn't want to make two separate webpack configs because I think it'd add more cognitive, development, and maintenance overhead than it's worth. (Though if it really gets to be a problem in the future, we can always split it into two separate configs later on.)


toolness:
@cmc333333 does the stuff I added in 147b160 seem like it would do the trick for deployment?

cmc333333:
@toolness seems correct on first blush. Mind testing it out by swapping the $CIRCLE_BRANCH of `master` to this branch?

api/webpack.config.js
```diff
@@ -18,8 +17,8 @@ module.exports = {
     ],
   },
   output: {
-    path: path.join(__dirname, 'static/ombpdf/js/'),
-    publicPath: 'static/ombpdf/js/',
+    path: path.join(__dirname, 'webpack-static/bundles/js/'),
+    publicPath: 'static/bundles/js/',
```
cmc333333: Just to make sure I understand this, it shouldn't be `webpack-static` because Django will serve all the static files under the `/static` path?

docker-compose.yml
```diff
@@ -181,6 +181,13 @@ services:
     ports: []
     depends_on: []
 
+  api-webpack:
+    <<: *API_WEBPACK
+    entrypoint: .docker/deps_ok_then ./node_modules/.bin/webpack
+    command: ''
+    ports: []
+    depends_on: []
```
cmc333333: Since we're not inheriting values for `ports` or `depends_on`, do we need to set them to empty?

api/webpack.config.js
toolness: oh, I just called it `webpack-static` so it makes sense to devs what's in the folder, when they're looking at the directory tree on their local systems... does that answer your question? I might be misinterpreting your question.

docker-compose.yml
toolness: Oh good point, I had just copy-pasted from the `webpack` container config, I'll remove those.

toolness:
By "" do you mean changing [this line in `.circleci/config.yml`](https://github.com/18F/omb-eregs/blob/14984ff6e61b7da0f8675df9182d896ed3da3dc2/.circleci/config.yml#L113)?

```diff
-             if [ "${CIRCLE_BRANCH}" == "master" ]; then
+             if [ "${CIRCLE_BRANCH}" == "api-webpack" ]; then
```

... so that this branch is temporarily deploying stuff to the dev site instead of `master`?

cmc333333:
Exactly; we can add that commit to push to demo without merging, then back it out once we know it works. Makes more sense when testing changes to deployment, but it'll work here, too.

toolness:
whoa...

`**ERROR** Could not install pip: signal: segmentation fault`

it's possible that could be caused by this PR, but ... i dunno, that seems weird, since we're just adding a few extra static files... any ideas @cmc333333 ?

toolness:
Oh nevermind, the actual error appears to be:

```
-----> Installing setuptools 38.2.3
       Download [https://buildpacks.cloudfoundry.org/dependencies/setuptools/setuptools-38.2.3-63c39408.zip]
       **ERROR** 
Exit status 223
```

but still, seems weird.  CircleCI also won't let me reset the build, as it says I need write permissions to trigger builds.

toolness:
Hooray, it appears to work!

https://omb-eregs-api-demo.app.cloud.gov/static/bundles/js/ombpdf.main.bundle.js

I will revert that last commit and merge this.

PR #808, opened 2018-01-04
Add command to import PDFs based on policy data.

cmc333333:
Policies were initially loaded with a source URI (they might also have a
replacement pdf). We can use that, along with our knowledge of the Obama
whitehouse archive to import remaining pdfs.

~~This PR isn't complete (it needs tests), but I wanted to send it up before our sprint cutoff.~~


toolness:
Noice! Since you appear to be adding some type annotations, do you mind adding this file to the `mypy-files.txt` so mypy looks at it?

cmc333333:
Re mypy: sure thing, I'll push a commit if it works pretty easily.

api/ombpdf/management/commands/import_pdfs_from_policies.py
```diff
@@ -0,0 +1,77 @@
...
+
+def import_pdfs_from_policies() -> Tuple[Set[str], Set[str]]:
+    """Attempt to import document text from all pdf policies which haven't
+    already been imported."""
+    successes, failures = set(), set()
+    has_docnodes = Exists(DocNode.objects.filter(policy=OuterRef('pk')))
+    query = Policy.objects \
+        .annotate(has_docnodes=has_docnodes) \
+        .filter(uri__endswith='.pdf') \
+        .filter(has_docnodes=False)     # not replacing data
```
toolness: I noticed that there's some similar code in `scrape_memoranda()` from #801 that is almost identical, save for the middle `filter()`.  Do you think we'd use this query elsewhere? I actually recall pair programming #661 with you, in which we used the same `Exists(DocNode.objects.filter(policy=OuterRef('pk')))` annotation, except we called it `has_docnode` (singular) there instead of plural... I wonder if it'd be useful to somehow pull this out into a bit of reusable code, like a method on `Policy`'s `QuerySet` or manager class or something?

Eh, it's up to you really.

api/ombpdf/management/commands/import_pdfs_from_policies.py
```diff
@@ -0,0 +1,77 @@
...
+    that works, then attempt to parse."""
+    uri = policy.uri.strip()
+    urls_to_try = [Url(uri)]
+    if OLD_DOMAIN in uri:
+        urls_to_try.append(Url(uri.replace(OLD_DOMAIN, NEW_DOMAIN)))
+    if policy.document_source:
+        urls_to_try.insert(0, Url(policy.document_source.url))
+
+    url = best_url(urls_to_try)
+    return bool(url and parse_pdf(policy, url))
```
toolness: I should have mentioned this when reviewing #801, but you're welcome to use [`tqdm`](https://pypi.python.org/pypi/tqdm) in `parse_pdf` to show download progress for the PDFs. I added it to our dev requirements file when migrating ombpdf to this repo, since it's used by that package's PDF retriever thingy.

api/ombpdf/management/commands/import_pdfs_from_policies.py
```diff
@@ -0,0 +1,77 @@
...
+
+class Command(BaseCommand):
+    help = import_pdfs_from_policies.__doc__    # noqa
+
+    def handle(self, *args, **options):
+        successes, failures = import_pdfs_from_policies()
+        num_success, num_failure = len(successes), len(failures)
+        logger.info('Imported %s new documents. Attempted %s, with %s '
+                    'failures: %s', num_success, num_success + num_failure,
+                    num_failure, list(sorted(failures)))
```
toolness: huh, is there a reason you're not using f-strings here? not a big deal either way, just curious.

api/ombpdf/management/commands/import_pdfs_from_policies.py
```diff
@@ -0,0 +1,77 @@
...
+            if response.status_code == 200:
+                return url
+        except RequestException:
+            continue
+    return None
+
+
+def import_from_policy(policy: Policy) -> bool:
+    """Convert a policy to a list of potential urls to try, pick the first
+    that works, then attempt to parse."""
```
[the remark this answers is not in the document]
cmc333333: Sure, I'll update. The return value is whether or not the attempt was successful at all.

api/ombpdf/management/commands/import_pdfs_from_policies.py
cmc333333: Cool! I'll give it a quick go; if it's simple, I'll update this PR.

api/ombpdf/management/commands/import_pdfs_from_policies.py
cmc333333: I'll make a little helper function; I agree that it makes sense to be placed on `Policy`, but that's create a circular import.

api/ombpdf/management/commands/import_pdfs_from_policies.py
cmc333333: It doesn't make a big difference here, but I've generally understood logging best practice to be to delay string interpolation until it can be performed by the log formatter. This allows the logging system to do interesting things with the parameters (dumb example: bold them) and can make a difference if the string interpolation is costly (and the logs aren't emitted). Neither case applies here, but still seems like a good idea.

api/ombpdf/management/commands/import_pdfs_from_policies.py
cmc333333: I'll add in an addendum PR as it'll require moving tqdm to the prod requirements.

api/ombpdf/management/commands/import_pdfs_from_policies.py
cmc333333: I'll add in an addendum PR as it'd probably be accurate to use the plural everywhere, but I don't want to modify the frontend here.

cmc333333:
Thanks @toolness!

PR #809, opened 2018-01-05
Add very barebones editor at /document/editor

toolness:
This just adds a very barebones ProseMirror editor at `/document/editor`:

> screenshot-2018-1-5 document editor

RIght now the document can only contain paragraphs, and it can't even load or save actual policies. It's just there to provide a basic foundation (mostly pulled from #798) to build the real thing upon. It *does* support undo/redo, at least, and a few basic keybindings like pressing <kbd>Ctrl</kbd>+<kbd>A</kbd> to select all, but that's about it.

Even the URL is temporary (it also conflicts with policies called `editor`, though I doubt any would  actually exist).  Once we actually support loading policies, we could maybe move it to `/document/:id/edit` or something.


api/document/js/main.ts
```diff
@@ -0,0 +1,83 @@
...
+import {EditorState} from "prosemirror-state";
+import {menuBar, undoItem, redoItem, MenuItem} from "prosemirror-menu";
+import {history, undo, redo} from "prosemirror-history";
+import {keymap} from "prosemirror-keymap";
+import {baseKeymap} from "prosemirror-commands";
+
+declare function require(path: string): null;
+
+require("prosemirror-view/style/prosemirror.css");
+require("prosemirror-menu/style/menu.css");
```
cmc333333: You probably tried this, but would
```js
import "prosemirror-view/style/prosemirror.css";
```
work? That's valid ES6, but maybe not TypeScript.

api/document/js/main.ts
```diff
@@ -0,0 +1,83 @@
...
+      group: 'inline',
+    },
+  },
+  marks: {},
+});
+
+function getDocument(): Node {
+  const doc = Node.fromJSON(schema, EXAMPLE_DOC);
+
+  doc.check();
```
cmc333333: Is this validating the document against that schema? If so, what happens when it fails? Is this a placeholder for something else?

api/document/js/main.ts
```diff
@@ -0,0 +1,83 @@
...
+}
+
+function getEl(id: string): Element {
+  const el = document.getElementById(id);
+  if (!el)
+    throw new Error(`element with id '${id}' not found`);
+  return el;
+}
+
+const view = new EditorView(getEl(EDITOR_ID), {
```
cmc333333: I see that the script isn't included until the bottom of the page, but are there any concerns about initialization order? Can we be positive that the DOM'll be fully parsed by the time this JS loads, or should we wrap this in an onload?

api/document/js/main.ts
```diff
@@ -0,0 +1,83 @@
...
+    plugins: [
+      menuBar({
+        floating: true,
+        content: [
+          [
+            // This forcible type assertion is due to what appears to
+            // be a bug in the documentation/type annotations for
+            // prosemirror-menu. For more details, see:
+            //
+            // https://github.com/ProseMirror/prosemirror-menu/issues/12
```
cmc333333: Excellent! Both documented and triggered an upstream change. Nice work.

api/document/js/main.ts
```diff
@@ -0,0 +1,83 @@
...
+            undoItem as any as MenuItem,
+            redoItem as any as MenuItem,
+          ]
+        ],
+      }),
+      keymap(Object.assign({}, baseKeymap, {
+        'Mod-z': undo,
+        'Shift-Mod-z': redo,
+      })),
```
cmc333333: What do you think about using the spread operator here? It's all over the UI code:
```js
keymap({
  ...baseKeymap,
  'Mod-z': undo,
  'Shift-Mod-z': redo,
}),
```

api/document/templates/document/editor.html
```diff
@@ -0,0 +1,12 @@
...
+<style>
+#editor {
+  border: 1px solid black;
+}
+</style>
+<title>Document editor</title>
+<h1>Document editor</h1>
+<div id="editor"></div>
+{% load static %}
+<script src="{% static 'bundles/js/document.main.bundle.js' %}"></script>
```
cmc333333: I know this is a placeholder, but would you mind adding `head`/`body` tags so we get started on the right foot?

api/document/js/main.ts
toolness: Yeah, that's what I originally tried, but for some reason typescript/webpack dies on it, claiming that it can't find that module.

I suspect it is because `ts-loader` (the typescript plugin for webpack) probably loads `import` statements on its own, without going through webpack, and it doesn't know what to do with non-standard kinds of imports like CSS, so using `require` likely bypasses typescript and goes straight to webpack, which deals with it. That's just my guess though.

That said, I should at least add a comment explaining this, so I'll do that!

api/document/js/main.ts
toolness: Yep, it's validating the document against the schema and raising an exception if it fails--it's really just there as a sanity check, though I suppose in production we might want to make it catch the exception and report an error or something, rather than crashing the whole front-end.

I'll add a comment to this effect.

api/document/js/main.ts
toolness: o sure i can wrap it up

api/document/js/main.ts
toolness: Oh yeah, I totes forgot about that, will bust it up!

api/document/templates/document/editor.html
toolness: Oh sure thing. I've actually been starting with this ever since I found out it was actually valid HTML5, but yeah, I always end up having to add the other tags as the thing evolves, so might as well start with it!

api/document/urls.py
```diff
@@ -1,8 +1,9 @@
 from django.conf.urls import url
 
-from document.views import TreeView
+from document.views import TreeView, editor
 
 urlpatterns = [
+    url(r'^editor$', editor),
```
cmc333333: Not for now, but something to think about: We may want to put this at `[domain]/admin/editor` to keep everything restricted in the `admin` path. In particular, I'm thinking about our CORS configuration.

We might also consider moving the admin to a separate domain.

toolness:
Ok, it looks like [1.7.0 was released](https://github.com/18F/omb-eregs/releases/tag/1.7.0), so I'm going to go ahead and merge this!

PR #810, opened 2018-01-05
Enable ombpdf in production.

cmc333333:
This'll allow us to run the PDF scraping scripts. The associated views are
still hidden behind a DEBUG flag.


cmc333333:
Thanks @toolness!

The decorator's an interesting idea, but I'd wonder about its reuse. Maybe we wait until we have a few more debug-only views?

PR #813, opened 2018-01-05
Addendum to #808

cmc333333:
@toolness pointed out some good improvements to #808, which I've split out here:
* Use tqdm to display pdf download progress when scraping PDFs
* Use a single function to annotate whether or not policies have related docnodes.


api/document/models.py
```diff
@@ -50,6 +50,17 @@ def annotations(self) -> Iterator['Annotation']:
...
 
+def annotate_with_has_docnodes(queryset=None):
+    """We frequently want to filter a Policy queryset by whether or not the
+    policies have associated DocNodes. This annotates a queryset with that
+    data."""
+    if queryset is None:
+        queryset = Policy.objects.all()
+
+    subquery = DocNode.objects.filter(policy=models.OuterRef('pk'))
+    return queryset.annotate(has_docnodes=models.Exists(subquery))
```
toolness: Have you considered making this a [custom `QuerySet` method](https://docs.djangoproject.com/en/2.0/topics/db/managers/#calling-custom-queryset-methods-from-the-manager) instead of a function?  You are definitely way more experienced with Django than me so I'll leave it up to you... I assume you have already considered this option and decided it was horrible for some reason, though I am curious to know why.

api/ombpdf/download_pdfs.py
```diff
@@ -29,16 +34,34 @@ def download(relpath, base_url=BASE_URL, domain=DOMAIN):
...
+        length = safe_content_length(response)
         with tqdm(total=length) as pbar:
-            for chunk in req.iter_content(chunk_size=1024):
-                fd.write(chunk)
+            for chunk in response.iter_content(chunk_size=1024):
+                write_to.write(chunk)
                 pbar.update(len(chunk))
 
-    return path
+    return write_to
```
toolness: Noice! Um, want to add this file to `mypy-files.txt`?

api/document/models.py
cmc333333: It's a great idea, but I was avoiding it due to circular imports. Documents reference Policies; adding a method to Policies (indirectly through custom manager) referencing Documents would make the loop. There might be different way to write the subquery, though. I'll poke around a bit more.

toolness:
Nice, adding the method to `PolicyQueryset` makes everything super slick!

cmc333333:
Thanks @toolness

PR #814, opened 2018-01-05
Add Jest support to API JS

toolness:
~~**Note that this is currently a PR against #809, not `master`.**~~

This builds upon #809 by adding support for Jest, along with one simple test.

Unfortunately, in doing so, it refactors a bunch of #809 by splitting up `main.ts` into a bunch of files, which muddles things up a bit.  Sorry about that...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,7 @@ ui-dist
 ui/static/font/fontawesome-webfont*
 ui/static/font/KaTeX*
 ui/static/styles.css*
-jest*
+jest_*
 ui/lcov-report/
 ui/lcov.info
 .next
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,7 @@ ui-dist
 ui/static/font/fontawesome-webfont*
 ui/static/font/KaTeX*
 ui/static/styles.css*
-jest_*
+jest_*/
 ui/lcov-report/
 ui/lcov.info
 .next
--- a/api/document/js/__tests__/example-doc.test.ts
+++ b/api/document/js/__tests__/example-doc.test.ts
@@ -1,0 +1,5 @@
+import exampleDoc from "../example-doc";
+
+test('example document conforms to schema', () => {
+  exampleDoc.check();
+});
--- a/api/document/js/example-doc.ts
+++ b/api/document/js/example-doc.ts
@@ -1,0 +1,18 @@
+import {Node} from "prosemirror-model";
+
+import schema from "./schema";
+
+const EXAMPLE_DOC = {
+  type: 'doc',
+  content: [{
+    type: 'paragraph',
+    content: [{
+      type: 'text',
+      text: 'Hello, I am a paragraph.',
+    }],
+  }],
+};
+
+const exampleDoc = Node.fromJSON(schema, EXAMPLE_DOC);
+
+export default exampleDoc;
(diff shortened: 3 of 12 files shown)


api/document/js/__tests__/example-doc.test.ts
```diff
@@ -0,0 +1,5 @@
+import exampleDoc from "../example-doc";
+
+test('example document conforms to schema', () => {
+  exampleDoc.check();
```
toolness: This is admittedly a silly test, since we'll be getting rid of the example document soon, but we can always modify this later on to contain example documents or somesuch (and probably rename the suite to `schema.test.js`, as a way to sanity check the schema itself).

api/document/js/example-doc.ts
```diff
@@ -0,0 +1,18 @@
...
+    content: [{
+      type: 'text',
+      text: 'Hello, I am a paragraph.',
+    }],
+  }],
+};
+
+const exampleDoc = Node.fromJSON(schema, EXAMPLE_DOC);
+
+export default exampleDoc;
```
toolness: Not that this file will be around for very long, but I think it's OK to export the actual document because [nodes in ProseMirror are immutable](https://prosemirror.net/docs/ref/#model.Node), yay!

.gitignore
```diff
@@ -11,7 +11,7 @@ ui-dist
 ui/static/font/fontawesome-webfont*
 ui/static/font/KaTeX*
 ui/static/styles.css*
-jest*
+jest_*/
```
toolness: I'm not sure if this is the right pattern, but on my system there were directories called `jest_0`, so I assumed that was the pattern. Had to change it because otherwise `jest-ts-preprocessor.js` was being ignored, and I wanted that filename to have the word "jest" in it.

api/package.json
```diff
@@ -38,5 +39,22 @@
...
+    "moduleFileExtensions": [
+      "ts",
+      "tsx",
+      "js"
+    ],
+    "transform": {
+      "^.+\\.(ts|tsx)$": "<rootDir>/jest-ts-preprocessor.js"
+    },
+    "testMatch": [
+      "**/__tests__/*.(ts|tsx|js)"
```
cmc333333: We specified the `test.js` extension on the ui side because we ended up needed helpers for our tests. I'm not sure if it's worth prepping for that now.

Unrelated: do we want to allow JS or should everything be type script?

api/jest-ts-preprocessor.js
```diff
@@ -0,0 +1,11 @@
+const tsc = require('typescript');
+const tsConfig = require('./tsconfig.json');
+
+module.exports = {
+  process(src, path) {
+    if (path.endsWith('.ts') || path.endsWith('.tsx')) {
```
cmc333333: Is this check required? I'd think the regex in `transform` would cover it.

toolness: Uhhhh good question. To be honest I just copy-pasted this from an [official jest example](https://github.com/facebook/jest/blob/master/examples/typescript/preprocessor.js), so I'm not sure. It does make sense that the regex would cover it...

api/package.json
toolness: Oh I like the idea of making it easy to add helpers to our tests if we need them, without having to modify the test configuration and rename tests.  Also, I like the idea of keeping the test naming conventions as similar as possible between the ui and api sides.

I think we should allow JS for now since we don't have a ton of institutional knowledge of TS at present, and should allow for the possibility that this TypeScript thing might explode in our face, or that it might just end up being more convenient in some cases to not use it. I think our code review process should be enough to ensure that we nudge each other towards TS when possible, without explicitly prohibiting JS in the build pipeline... let me know if you disagree though!

PR #815, opened 2018-01-08
Support PUT on the API's /document/:id endpoint

toolness:
This is an attempt to productionalize the API work done in #798--specifically, the content in [`import_json_doc.py`](https://github.com/18F/omb-eregs/pull/798/files#diff-81dc0e93a7089b74df66fb7f97403fe7)--by modifying the `DocCursorField`'s read-only fields to be writable.
  
To do (some/all can be pushed to separate issues):
- [ ] Support changes to a document's `meta` information (for the root node, I think this will mean `Policy` information).
- [x] Actually validate the content (e.g, annotations) of document nodes on `PUT` requests.
- [ ] Consider adding documentation for this endpoint. Right now it's very unclear, without digging into implementation and/or tests, what fields of the JSON returned by the `GET` endpoint can be altered (e.g. `title`) vs. which ones are automatically generated/re-constructed on the server (e.g. `depth`).
- [x] ~~Figure out what to do~~ Return HTTP 400 if a client [issues a `PUT` to a document's sub-node](https://github.com/18F/omb-eregs/pull/815#issuecomment-356313093).  
- [x] Add all new, annotated files to `mypy-files.txt`.
- [x] Modify `json_importer` to support nested content annotations, and look at `text` only on leaf `__text__` nodes.

### DIFF
--- a/api/document/json_importer/annotations.py
+++ b/api/document/json_importer/annotations.py
@@ -36,10 +36,10 @@ def footnote_citation(cursor: JSONAwareCursor, content: PrimitiveDict,
 @annotator
 def external_link(cursor: JSONAwareCursor, content: PrimitiveDict,
                   start: int) -> ExternalLink:
-    text = get_content_text(content['inlines'])
     return ExternalLink(
         doc_node=cursor.model, start=start,
-        end=start + len(text), href=content['href']
+        end=start + get_content_length(content['inlines']),
+        href=content['href']
     )
 
 
@@ -53,10 +53,8 @@ def find_annotations(items: List[PrimitiveDict],
             start += len(content['text'])
         else:
             yield (content, start)
-            for icontent, istart in find_annotations(content['inlines'],
-                                                     start):
-                yield (icontent, istart)
-                start = istart
+            yield from find_annotations(content['inlines'], start)
+            start += get_content_length(content['inlines'])
 
 
 def derive_annotations(cursor: JSONAwareCursor) -> AnnotationDict:
@@ -74,6 +72,16 @@ def derive_annotations(cursor: JSONAwareCursor) -> AnnotationDict:
             annotations[cls].extend(annos)
 
     return annotations
+
+
+def get_content_length(content: List[PrimitiveDict]) -> int:
+    length = 0
+    for c in content:
+        if c['content_type'] == '__text__':
+            length += len(c['text'])
+        else:
+            length += get_content_length(c['inlines'])
+    return length
 
 
 def get_content_text(content: List[PrimitiveDict]):
--- a/api/document/json_importer/annotations.py
+++ b/api/document/json_importer/annotations.py
@@ -1,5 +1,5 @@
 from collections import defaultdict
-from typing import Callable, DefaultDict, Dict, List, Type
+from typing import Callable, DefaultDict, Dict, Iterator, List, Tuple, Type
 
 from document.models import Annotation, ExternalLink, FootnoteCitation
 from document.tree import JSONAwareCursor, JsonDict
@@ -20,7 +20,7 @@ def annotator(fn: Annotator):
 @annotator
 def footnote_citation(cursor: JSONAwareCursor, content: JsonDict,
                       start: int) -> FootnoteCitation:
-    text = content['text']
+    text = get_content_text(content['inlines'])
     referencing = list(cursor.filter(
         lambda m: m.node_type == 'footnote'
         and m.type_emblem == text.strip()
@@ -36,29 +36,51 @@ def footnote_citation(cursor: JSONAwareCursor, content: JsonDict,
 @annotator
 def external_link(cursor: JSONAwareCursor, content: JsonDict,
                   start: int) -> ExternalLink:
+    text = get_content_text(content['inlines'])
     return ExternalLink(
         doc_node=cursor.model, start=start,
-        end=start + len(content['text']), href=content['href']
+        end=start + len(text), href=content['href']
     )
 
 
 AnnotationDict = DefaultDict[Type[Annotation], List[Annotation]]
 
 
+def find_annotations(items: List[JsonDict],
+                     start: int=0) -> Iterator[Tuple[JsonDict, int]]:
+    for content in items:
+        if content['content_type'] == '__text__':
+            start += len(content['text'])
+        else:
+            yield (content, start)
+            for icontent, istart in find_annotations(content['inlines'],
+                                                     start):
+                yield (icontent, istart)
+                start = istart
+
+
 def derive_annotations(cursor: JSONAwareCursor) -> AnnotationDict:
     annotations: AnnotationDict = defaultdict(list)
-    start = 0
-    for content in cursor.json_content:
+
+    for content, start in find_annotations(cursor.json_content):
         content_type = content['content_type']
-        if content_type != '__text__':
-            if content_type not in annotators:
-                raise ValueError(f"no annotator found for {content_type}")
-            anno = annotators[content_type](cursor, content, start)
-            annotations[anno.__class__].append(anno)
-        start += len(content['text'])
+        if content_type not in annotators:
+            raise ValueError(f"no annotator found for {content_type}")
+        anno = annotators[content_type](cursor, content, start)
+        annotations[anno.__class__].append(anno)
 
     for child_cursor in cursor.children():
         for cls, annos in derive_annotations(child_cursor).items():
             annotations[cls].extend(annos)
 
     return annotations
+
+
+def get_content_text(content: List[JsonDict]):
+    chunks = []
+    for c in content:
+        if c['content_type'] == '__text__':
+            chunks.append(c['text'])
+        else:
+            chunks.append(get_content_text(c['inlines']))
+    return ''.join(chunks)
--- a/api/document/json_importer/annotations.py
+++ b/api/document/json_importer/annotations.py
@@ -2,10 +2,10 @@ from collections import defaultdict
 from typing import Callable, DefaultDict, Dict, Iterator, List, Tuple, Type
 
 from document.models import Annotation, ExternalLink, FootnoteCitation
-from document.tree import JSONAwareCursor, JsonDict
+from document.tree import JSONAwareCursor, PrimitiveDict
 
 Annotator = Callable[
-    [JSONAwareCursor, JsonDict, int],
+    [JSONAwareCursor, PrimitiveDict, int],
     Annotation
 ]
 
@@ -18,7 +18,7 @@ def annotator(fn: Annotator):
 
 
 @annotator
-def footnote_citation(cursor: JSONAwareCursor, content: JsonDict,
+def footnote_citation(cursor: JSONAwareCursor, content: PrimitiveDict,
                       start: int) -> FootnoteCitation:
     text = get_content_text(content['inlines'])
     referencing = list(cursor.filter(
@@ -34,7 +34,7 @@ def footnote_citation(cursor: JSONAwareCursor, content: JsonDict,
 
 
 @annotator
-def external_link(cursor: JSONAwareCursor, content: JsonDict,
+def external_link(cursor: JSONAwareCursor, content: PrimitiveDict,
                   start: int) -> ExternalLink:
     text = get_content_text(content['inlines'])
     return ExternalLink(
@@ -46,8 +46,8 @@ def external_link(cursor: JSONAwareCursor, content: JsonDict,
 AnnotationDict = DefaultDict[Type[Annotation], List[Annotation]]
 
 
-def find_annotations(items: List[JsonDict],
-                     start: int=0) -> Iterator[Tuple[JsonDict, int]]:
+def find_annotations(items: List[PrimitiveDict],
+                     start: int=0) -> Iterator[Tuple[PrimitiveDict, int]]:
     for content in items:
         if content['content_type'] == '__text__':
             start += len(content['text'])
@@ -76,7 +76,7 @@ def derive_annotations(cursor: JSONAwareCursor) -> AnnotationDict:
     return annotations
 
 
-def get_content_text(content: List[JsonDict]):
+def get_content_text(content: List[PrimitiveDict]):
     chunks = []
     for c in content:
         if c['content_type'] == '__text__':
--- a/api/document/json_importer/annotations.py
+++ b/api/document/json_importer/annotations.py
@@ -1,0 +1,64 @@
+from collections import defaultdict
+from typing import Callable, DefaultDict, Dict, List, Type
+
+from document.models import Annotation, ExternalLink, FootnoteCitation
+from document.tree import JSONAwareCursor, JsonDict
+
+Annotator = Callable[
+    [JSONAwareCursor, JsonDict, int],
+    Annotation
+]
+
+annotators: Dict[str, Annotator] = {}
+
+
+def annotator(fn: Annotator):
+    annotators[fn.__name__] = fn
+    return fn
+
+
+@annotator
+def footnote_citation(cursor: JSONAwareCursor, content: JsonDict,
+                      start: int) -> FootnoteCitation:
+    text = content['text']
+    referencing = list(cursor.filter(
+        lambda m: m.node_type == 'footnote'
+        and m.type_emblem == text.strip()
+    ))
+    if not referencing:
+        raise ValueError(f'unable to find footnote for citation {text}')
+    return FootnoteCitation(
+        doc_node=cursor.model, start=start, end=start + len(text),
+        footnote_node=referencing[0].model,
+    )
+
+
+@annotator
+def external_link(cursor: JSONAwareCursor, content: JsonDict,
+                  start: int) -> ExternalLink:
+    return ExternalLink(
+        doc_node=cursor.model, start=start,
+        end=start + len(content['text']), href=content['href']
+    )
+
+
+AnnotationDict = DefaultDict[Type[Annotation], List[Annotation]]
+
+
+def derive_annotations(cursor: JSONAwareCursor) -> AnnotationDict:
+    annotations: AnnotationDict = defaultdict(list)
+    start = 0
+    for content in cursor.json_content:
+        content_type = content['content_type']
+        if content_type != '__text__':
+            if content_type not in annotators:
+                raise ValueError(f"no annotator found for {content_type}")
+            anno = annotators[content_type](cursor, content, start)
+            annotations[anno.__class__].append(anno)
+        start += len(content['text'])
+
+    for child_cursor in cursor.children():
+        for cls, annos in derive_annotations(child_cursor).items():
+            annotations[cls].extend(annos)
+
+    return annotations
--- a/api/document/json_importer/importer.py
+++ b/api/document/json_importer/importer.py
@@ -1,12 +1,12 @@
-from collections import defaultdict
-from typing import (Callable, DefaultDict, Dict, List, NamedTuple, Optional,
-                    Type)
+from typing import List, NamedTuple, Optional
 
 from django.db import transaction
 
-from document.models import Annotation, DocNode, ExternalLink, FootnoteCitation
+from document.models import DocNode
 from document.tree import JSONAwareCursor, JsonDict
 from reqs.models import Policy
+
+from .annotations import derive_annotations
 
 
 class PreparedNode(NamedTuple):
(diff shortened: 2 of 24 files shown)


api/document/serializers/doc_cursor.py
```diff
@@ -1,13 +1,34 @@
...
+    def get_attribute(self, instance: DocCursor) -> DocCursor:
+        return instance
+
+
+class ChildrenField(DocCursorField):
+    def to_representation(self, instance: DocCursor) -> List[JsonDict]:
+        return DocCursorSerializer(
+            instance.children(), many=True,
+            context={**self.context, 'is_root': False},
+        ).data
```
toolness: I've verified the type annotations here by adding `assert` statements into the code and then re-running the tests (since I'm unfamiliar with the code, I wanted to verify that my assumptions were correct).

toolness:
My first commit is really just a refactoring which changes `DocCursorSerializer.children` from a read-only `SerializerMethodField` into a [custom field](http://www.django-rest-framework.org/api-guide/fields/#custom-fields) whose `to_internal_value()` method (used for writing back to the db) is still unimplemented. The next step will be to actually implement it, and then write tests to verify that the serializer can be used for writing to the `children` field.

@cmc333333 does this seem like the right kind of approach to take?

api/document/serializers/doc_cursor.py
cmc333333: Yeah, it's a bit unexpected, but the `.data` field will return either serialized version of an object (an ordered dict of string fields to other ordered dicts or value types) _or_ a list of serialized objects if `many=True`.

api/document/serializers/doc_cursor.py
```diff
@@ -1,13 +1,34 @@
...
+
+    def to_internal_value(self, data: List[JsonDict]) -> DocCursor:
```
cmc333333: I suspect the signature will be `-> List[DocCursor]` at the end of the day, maybe `-> List[DocNode]`? Perhaps we'll do the parent-child relation in `DocCursorSerializer`'s `update()`?

toolness: Ah cool, I will keep that in mind! Thanks for the feedback.

toolness:
Heh, so it looks like the current permission class for DRF, [`DjangoModelPermissionsOrAnonReadOnly`](http://www.django-rest-framework.org/api-guide/permissions/#djangomodelpermissionsoranonreadonly), is actually potentially all we need. unless we want to get more granular and have permissions on a per-instance level.

I basically just swapped out our `RetrieveAPIView` for a [`RetrieveUpdateAPIView`](http://www.django-rest-framework.org/api-guide/generic-views/#retrieveupdateapiview) and it looks like we're good to go as far as permissions and endpoints are concerned. Now we just need to actually implement the serializers for writing.

api/document/tree.py
```diff
@@ -21,6 +23,18 @@ def __getattr__(self, attr):
         """Delegate fields/methods to wrapped model."""
         return getattr(self.model, attr)
 
+    @property
+    def _prefetched_objects_cache(self):
+        """Django REST Framework thinks we're a bona fide Django model,
+        and its UpdateModelMixin sometimes writes to the model's
+        undocumented _prefetched_objects_cache property, so we'll
+        need to make it writable too."""
```
toolness: Um, needless to say, this feels _super_ sketchy and I'm not sure if there's a cleaner way to get around it.

toolness:
Ok, I think I've actually got a bunch of this implemented--it seems like much of my code from the prototype was able to be reused. Stuff I learned:

* Ug, DRF thinks `DocCursor` is an actual Django `Model` subclass and this causes [complications](https://github.com/18F/omb-eregs/pull/815#discussion_r160429898).  It'd be nice if we can find a cleaner fix than the one I've provided.

* It appears the `to_internal_value()` method of `serializers.Field` can really return whatever we want it to return; it's ultimately passed back as validated data to the `update()` method on our model serializer, where we can do whatever we want with it.  So for now, I'm just doing validation in `to_internal_value()` and postponing the creation of actual model instances until `update()`, where I basically delegate out to the JSON importer from the prototype (which itself is largely based on the existing XML importer).

toolness:
Ack, I just realized, from reading the tests, that the document endpoint actually allows sub-nodes of the document to be specified in the URL, e.g. `/{policy.pk}/root_0__sec_1`. However, right now the `update()` method I implemented assumes that the _root_ of the document is being updated, not a sub-node, so it's blowing away all the policy's document nodes and then re-constructing them from the payload. Either my strategy needs to change, or we need to raise a 400 or something if the client tries to `PUT` to a sub-node.

api/document/json_importer/importer.py
```diff
@@ -0,0 +1,131 @@
...
+                    Type)
+
+from django.db import transaction
+
+from document.models import Annotation, DocNode, ExternalLink, FootnoteCitation
+from document.tree import JSONAwareCursor, JsonDict
+from reqs.models import Policy
+
+
+class PreparedNode(NamedTuple):
```
toolness: In the original prototype code this was called `CleanedNode`, but I think `PreparedNode` is a better name--"cleaned" suggests some sort of validation, which isn't being done here (validation is explicitly done by the DRF serializers).

api/document/serializers/doc_cursor.py
```diff
@@ -25,22 +65,17 @@ class Meta:
...
-        annotations = nest_annotations(
-            instance.annotations(), len(instance.text))
-        return NestedAnnotationSerializer(
-            annotations,
-            context={'cursor': instance, 'parent_serializer': self},
-            many=True,
-        ).data
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.context['parent_serializer'] = self
```
toolness: This feels weird, but I couldn't find any easier/cleaner way to pass the parent serializer to the new `ContentField` class.

toolness:
Oh funky, it looks like the serializer fields' `to_internal_value()` methods really are intended to return values that are as "primitive" as possible. I've been watching the beginning of [Tom Christie's 2014 Under The Hood Talk](https://www.youtube.com/watch?v=3cSsbe-tA0E) and he explains how DRF 2.0 had a policy of returning a model instance, but in 3.0 they moved to returning a primitive validated data dictionary because of complications in returning inter-related models that needed to be saved to the database before much could be done with them.

api/document/tree.py
toolness: Hmm, one potential way we could work around this may be by converting `DocCursorSerializer` to inherit from the standard `Serializer` rather than `ModelSerializer`, but somehow using a separate `ModelSerializer` subclass to save just the parts of the `DocNode` model that we need it for. I could try experimenting with this.

api/document/serializers/doc_cursor.py
cmc333333: I bet there's a static field we can override (DRF seems to love this pattern) for "child context" or something similar. No need to find it now, but worth investigating.

api/document/tree.py
cmc333333: I think switching to the `Serializer` base class makes sense; we've only been using `ModelSerializer` for simplicity when generating fields. I think this nastiness out-weights that.

api/document/json_importer/importer.py
cmc333333: Without digging in _too_ deep, it feels like this tuple is duplicating `DocCursor`'s interface. I wonder if there's a more one-to-one way to do this conversion.

api/document/tree.py
```diff
@@ -1,11 +1,13 @@
 from collections import Counter
-from typing import Callable, Optional
+from typing import Any, Callable, Dict, List, Optional
 
 from networkx import DiGraph
 from networkx.algorithms.dag import descendants
 
 from document.models import DocNode
 
+JsonDict = Dict[str, Any]
```
cmc333333: We should probably swap `Json` for something else everywhere, lest we mislead. The deserialization process takes _in_ json, xml, etc. and spits out Python primitives. Would it make sense to use [`ReturnDict`](https://github.com/encode/django-rest-framework/blob/91fa8b923a51e51bafebcfa15e6e8ebab928642e/rest_framework/utils/serializer_helpers.py#L12) as our type?

api/document/json_importer/importer.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+@annotator
+def footnote_citation(cursor: JSONAwareCursor, content: JsonDict,
+                      start: int) -> FootnoteCitation:
+    text = content['text']
+    referencing = list(cursor.filter(
+        lambda m: m.node_type == 'footnote'
+        and m.type_emblem == text.strip()
+    ))
```
cmc333333: This might be copy-pasted from somewhere else, but if we have the referenced node's identifier, we should be able to `cursor.jump_to` it directly.

api/document/json_importer/importer.py
```diff
@@ -0,0 +1,131 @@
...
+
+    for child_cursor in cursor.children():
+        for cls, annos in derive_annotations(child_cursor).items():
+            annotations[cls].extend(annos)
+
+    return annotations
+
+
+def get_content_text(content: List[JsonDict]):
+    return ''.join([c['text'] for c in content])
```
cmc333333: The linter may complain here: there's no need for the list; `join`'ll work on the generator.

cmc333333:
Re `to_internal_values()`, the gymnastics around model relationships is very annoying. I think #823 might solve our use case, but I can see why Christie moved away from models as a general solution.

Regarding PUTing to a sub-document, raising a 400 seems like a good temporary solution. We aren't using the sub-document endpoint anywhere in the app, but it's awfully convenient for inspecting the data.

api/document/tree.py
toolness: Oh funky.  Umm.  The main concept I wanted to convey was that the keys are `str` and the values can be ... well, I guess they can be any kind of data type that is serializable into a format that's like JSON, but might not necessarily be JSON?  Humm.

toolness:
Thanks for the feedback @cmc333333 !

What is a "more one-to-one way"?

api/document/json_importer/importer.py
toolness: Yep, it's copy-pasted from the XML importer annotation code.

Hmm, I'm not sure if we do actually have the referenced node's identifier, since the identifiers change based on where they happen to be in the document structure? Like, won't the identifier be something like `policy_1__sec_1__para_1__footnote_1`?

cmc333333: That's a good point. Let's ponder it a bit more -- I don't think it's safe to assume a footnote marker will be unique, but we aren't generating the identifiers in the editor, so there's no way to refer to another node. This's likely to come up if we add in-document definitions or internal links, but we can punt for now.

api/document/tree.py
cmc333333: Spit balling:
* Attrs
* StringMap
* Primitives
* POPOs
* ObjectData
* Parsed

cmc333333:
Re one-to-one: is it possible we could create `DocCursor`s directly from the `JsonDict`? I'm wondering if we can cut out the `PreparedNode` layer.

toolness:
Actually, what do you think of changing `DocCursorSerializer` to `DocNodeSerializer` and keeping it as a `ModelSerializer`?  We could potentially keep an actual `DocCursor` in the serializer's `context` so we could still use it whenever we wanted to traverse the document... but that way, having a bona fide `DocNode` as the serializer's instance would eliminate the need for hacky workarounds on the `DocCursor` class.

cmc333333:
Re additional steps, I'm in favor of merging as soon as we're happy with the current functionality. We can split up next steps while this first step is merged.

toolness:
Oh ok, in that case I will just add the annotation-validating logic and I think it will be OK to merge after that.

cmc333333:
Re DocCursorSerializer -> DocNodeSerializer: good idea -- let's try it out.

Maybe we should split this into two PRs: one to refactor the current functionality and a second to build the "write" support into that?

api/document/tree.py
toolness: How about `PrimitiveDict`?  It's a bit wordy, but it seems kinda accurate...

toolness:
Ok, since a lot of `PreparedNode`'s logic was actually doing validation/normalization that's now being done by the serializer, I was able to get rid of it entirely! Woot.

I also realized I made a mistake when implementing the original JSON document importer, in that I never looked at a content node's `inlines`, only at its `text`.  This worked because none of the annotations were ever nested in my prototype, but now that I've implemented the serializer to be inlines-aware, I should take the time to modify the JSON document importer to use them too (the document importer will actually crash if I don't do that, so I kinda need to).

Regarding splitting this into two PRs: do you feel strongly about that? Right now this PR doesn't include any refactoring of current functionality, and while the `_prefetched_objects_cache property` is horribly hacky, I think it's ok to submit a PR immediately after this one that fixes it... If you feel strongly about it though I can bust it up.

toolness:
Ok, I think this is ready for review/merge now! Woo!

One thing I could potentially add for now, if we're super concerned about security, is adding a feature flag (e.g. `settings.ENABLE_API_HTTP_PUT`) that is only enabled when `DEBUG` is set.

toolness:
I'm curious how DRF converts XML to its funky JSON blobby format that it passes to serializers. If it's not that different from our current ad-hoc XML format, we could potentially rejigger the XML importer to just use our serializer to do its thing... Eh, just spitballing, and definitely out-of-scope of this PR.

api/document/tests/json_importer/annotations_test.py
```diff
@@ -0,0 +1,43 @@
+from document.json_importer.annotations import derive_annotations
+from document.json_importer.importer import convert_node
+from document.models import ExternalLink, FootnoteCitation
+
+from .importer_test import PARA_WITH_LINK
```
cmc333333: What do you think about duplicating (or simplifying and duplicating) this fixture into this test file? I think ideal unit tests are self-contained; needing to look in a second file to understand these tests is a layer of indirection we can probably avoid.

api/document/serializers/content.py
```diff
@@ -85,10 +86,21 @@ class NestedAnnotationSerializer(serializers.Serializer):
...
     def to_representation(self, data: NestableAnnotation):
         serializer = self.serializer_mapping[data.annotation_class]
         return serializer(data, context=self.context).data
 
+    def to_internal_value(self, data: JsonDict) -> JsonDict:
+        content_type = data.get('content_type')
+        if content_type is None:
+            raise ValidationError("missing content_type")
+        if content_type not in self.content_type_mapping:
+            raise ValidationError(f"unknown content_type: {content_type}")
```
cmc333333: Should these validations be on the `content_type` field instead? I'd expect cross-field validations to occur here.

api/document/views.py
```diff
@@ -52,6 +54,13 @@ def get_serializer_context(self):
             'policy': getattr(self, 'policy', None),
         }
 
+    def put(self, request, *args, **kwargs):
+        if self.kwargs.get('identifier'):
```
cmc333333: Nice, clean solution.

api/document/serializers/content.py
```diff
@@ -123,19 +127,34 @@ def get_content_type(self, instance: NestableAnnotation):
...
+        if self.IS_LEAF:
+            return []
         return NestedAnnotationSerializer(
             instance.children, context=self.context, many=True).data
 
     def get_text(self, instance: Annotation):
         return self.doc_node_text[instance.start:instance.end]
 
+    def to_internal_value(self, data: JsonDict) -> JsonDict:
+        inlines = [] if self.IS_LEAF else data.get('inlines', [])
```
cmc333333: Note that this'll silently drop bad user data (a __text__ with non-empty `inlines`), which seems risky. Should it be raising a validation exception instead?

api/document/serializers/content.py
```diff
@@ -111,19 +127,34 @@ def get_content_type(self, instance: NestableAnnotation):
...
+        result['inlines'] = []
+        if inlines:
+            serializer = NestedAnnotationSerializer()
+            result['inlines'].extend([
+                serializer.to_internal_value(item) for item in inlines
+            ])
+        if self.IS_LEAF:
+            serializer = serializers.CharField(trim_whitespace=False)
+            result['text'] = serializer.to_internal_value(data.get('text'))
+        return result
```
toolness: Overriding the serializer's `to_internal_value()` was more convenient than turning `CONTENT_TYPE`, `inlines`, and `text` into specialized fields, but we could do that instead if we want.

api/document/serializers/content.py
```diff
@@ -123,19 +127,34 @@ def get_content_type(self, instance: NestableAnnotation):
...
+    def to_internal_value(self, data: JsonDict) -> JsonDict:
+        inlines = [] if self.IS_LEAF else data.get('inlines', [])
+        result = super().to_internal_value(data)
+        result['content_type'] = self.CONTENT_TYPE
+        result['inlines'] = []
+        if inlines:
+            serializer = NestedAnnotationSerializer()
+            result['inlines'].extend([
+                serializer.to_internal_value(item) for item in inlines
+            ])
```
cmc333333: I wonder if this makes more sense as a custom field like the `children` field of DocNodes? It feels like we're fighting DRF here.

api/document/serializers/content.py
```diff
@@ -123,19 +127,34 @@ def get_content_type(self, instance: NestableAnnotation):
...
+        if self.IS_LEAF:
+            serializer = serializers.CharField(trim_whitespace=False)
+            result['text'] = serializer.to_internal_value(data.get('text'))
```
cmc333333: This is deserializing into a different shape than what was originally serialized, which seems fishy. I wonder if we should be tracking offsets _here_.

api/document/json_importer/importer.py
```diff
@@ -9,57 +9,35 @@
...
     return ''.join(c['text'] for c in content)
 
 
 def convert_node(node: JsonDict, policy: Optional[Policy]=None,
                  parent: Optional[JSONAwareCursor]=None) -> JSONAwareCursor:
-    prepped = PreparedNode.from_json(node)
+    kwargs = node.copy()
+    children = kwargs.pop('children')
+    content = kwargs.pop('content')
+    kwargs['text'] = get_content_text(content)
```
cmc333333: Nice, much cleaner

api/document/json_importer/annotations.py
```diff
@@ -36,29 +36,51 @@ def footnote_citation(cursor: JSONAwareCursor, content: JsonDict,
...
+                     start: int=0) -> Iterator[Tuple[JsonDict, int]]:
+    for content in items:
+        if content['content_type'] == '__text__':
+            start += len(content['text'])
+        else:
+            yield (content, start)
+            for icontent, istart in find_annotations(content['inlines'],
+                                                     start):
+                yield (icontent, istart)
+                start = istart
```
cmc333333: Do you think we could make this and `derive_annotations` simpler by combining them somehow? This function seems very abstract; I speculate that the logic would be easier to follow if we were yielding annotations instead of `Tuple(JsonDict, int)` pairs.

api/document/tree.py
```diff
@@ -6,7 +6,7 @@
 
 from document.models import DocNode
 
-JsonDict = Dict[str, Any]
+PrimitiveDict = Dict[str, Any]
```
cmc333333: Thanks for doing this; ideally we'd remove `Json` everywhere else (e.g. JSONAwareCursor), too.

api/document/json_importer/annotations.py
```diff
@@ -76,6 +74,16 @@ def derive_annotations(cursor: JSONAwareCursor) -> AnnotationDict:
...
 
 
+def get_content_length(content: List[PrimitiveDict]) -> int:
+    length = 0
+    for c in content:
+        if c['content_type'] == '__text__':
+            length += len(c['text'])
+        else:
+            length += get_content_length(c['inlines'])
+    return length
```
cmc333333: This is _almost_ identical to `get_content_text`. Should we just run `len(get_content_text(content))`?

api/document/json_importer/annotations.py
```diff
@@ -0,0 +1,94 @@
...
+annotators: Dict[str, Annotator] = {}
+
+
+def annotator(fn: Annotator):
+    annotators[fn.__name__] = fn
+    return fn
+
+
+@annotator
+def footnote_citation(cursor: JSONAwareCursor, content: PrimitiveDict,
```
cmc333333: Should this and the other annotation deserializing logic be in the corresponding serializer (e.g. `FootnoteCitationSerializer`)? That seems like the DRF way.

api/document/json_importer/annotations.py
toolness: Yeah, I was probably over-optimizing--that was originally what I was going to do, but I didn't want to require python to allocate a bunch of strings only to not use them. But it's also a bunch of added code that needs to be separately tested and whatnot. Blech.

api/document/json_importer/annotations.py
toolness: Maybe? The thing I am constantly finding confusing about all this is that DRF really seems to want the validation process to return a plain old python dictionary with primitive values, not having anything to do with models, which is why this PR has the deserialization deal _only_ with validation/normalization of input and the JSON importer deal _only_ with actually turning that validated structure into a thing the database can understand.

That said, I think my fear of `DocCursor` is impacting the way I'm working on things in a more general way.  It's also part of why I found it easier to go through the XML intermediary during the PDF import, in #767.  Even without DRF's nudging, I'm finding that I tend to veer towards intermediary abstractions that process data to _prepare_ it for manipulation via a `DocCursor`, and then do the `DocCursor` stuff in a separate step.

Anyways, I guess maybe I should focus a bit on actually learning `DocCursor` better, perhaps by writing some documentation for it, as I had brought up in #767. If I feel more confident using that, I'll probably be more likely to submit PRs that have less unnecessary abstractions in them, not to mention making me better at reviewing code that uses the class!

api/document/serializers/content.py
toolness: Wait, it is?  I was just using `serializers.CharField` to raise any validation errors if e.g. the payload provided a non-string value for the `text` key. My intent was definitely not to deserialize into a different shape than what was originally serialized, at least.

api/document/serializers/content.py
toolness: Yeah, as I think I probably [mentioned while you were reviewing](https://github.com/18F/omb-eregs/pull/815#discussion_r160726146), this part of the code was done rather hastily, partly because I was getting really tired of making custom field types and just wanted to finish up the PR. As I mentioned to you in Slack, I probably should have reduced the scope of this PR but hid the `PUT` endpoint behind a feature flag so that I wouldn't be worried about the incomplete functionality being abused.

api/document/serializers/content.py
toolness: Er, that said, I _am_ deserializing into a _subset_ of the shape that was originally serialized, e.g. non-leaf nodes don't have `text` on them, since that duplicates the content of their inlines. (That said, this goes back to the notion of documenting this endpoint, because it's unclear whether doing a `GET` and then modifying a non-leaf content node's `text` field, or even a child node's `text` field, should do anything. In this case I decided it shouldn't, since the "real" content is made up of all the text nodes, and the text of non-leaf/child nodes is really a lossy representation of what's inside. But maybe we should discuss this more if we both have different ideas of how the `PUT` endpoint should work.)

api/document/serializers/content.py
toolness: Yep, see [my other comment](https://github.com/18F/omb-eregs/pull/815#discussion_r160744971) for more details.

PR #824, opened 2018-01-09
Foothold for custom styles in Django admin

cmc333333:
We know we'll want to reskin a significant chunk of the Django admin. Although [django-admin-interface](https://github.com/fabiocaccamo/django-admin-interface) will get us close, we'll likely want to override other styles as well.

This adds a config for building a SCSS bundle to the api webpack and includes that in the Django admin. The SCSS file includes a commented style, which quickly tests that everything works.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -75,7 +75,7 @@ jobs:
             COMPOSE_FILE: docker-compose-circle.yml
           command: docker-compose run --rm api-npm test
       - run:
-          name: API Build JS
+          name: API Build JS & CSS
           environment:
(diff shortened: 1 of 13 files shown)


cmc333333:
Thanks @rememberlenny!
No new commands here -- `dev-api` depends on the `api-webpack-watcher` script, which should be rebuilding the admin/editor's JS and style sheets. That said, it's very easy for api-webpack-watcher to get bogged down; I'll add some notes to the README.

cmc333333:
Thanks @rememberlenny @tadhg-ohiggins and @toolness ! I'll rebase and merge

PR #837, opened 2018-01-10
Factor out tests/json_importer/factories.py

toolness:
This gets rid of a janky code smell introduced in #815. whereby one of our test suites was importing factory functions and fixtures introduced in an unrelated test suite. It also hopefully lays some groundwork for less verbose JSON and/or serializer-related tests going forward.


toolness:
Oh cool! I didn't realize factory boy let one create the ol' dicts. That is neat.

PR #838, opened 2018-01-10
Factor out InlinesField and TextField in serializers/content.py

toolness:
This un-janks [another yucky hack](https://github.com/18F/omb-eregs/pull/815#discussion_r160726424) introduced in #815 by un-bloating the `BaseAnnotationSerializer` a bit:

* The `text` field is now its own serializer field. Note the comment on its `to_internal_value` method, it's a bit funky.

* The `inlines` field is now its own serializer field. If `inlines` is non-empty on a leaf node (i.e., `__text__`), we raise a validation error, as suggested by @cmc333333 at some point, I think.

### DIFF
--- a/api/document/serializers/content.py
+++ b/api/document/serializers/content.py
@@ -123,10 +123,8 @@ class InlinesField(NestableAnnotationField):
     def to_internal_value(self,
                           data: List[PrimitiveDict]) -> List[PrimitiveDict]:
         if not self.is_leaf_node:
-            serializer = NestedAnnotationSerializer()
-            return [
-                serializer.to_internal_value(item) for item in data
-            ]
+            serializer = NestedAnnotationSerializer(many=True)
+            return serializer.to_internal_value(data)
         elif data:
             raise ValidationError('leaf nodes cannot contain nested content')
         return []
--- a/api/document/serializers/content.py
+++ b/api/document/serializers/content.py
@@ -102,22 +102,73 @@ class NestedAnnotationSerializer(serializers.Serializer):
         return serializer.to_internal_value(data)
 
 
-class BaseAnnotationSerializer(serializers.Serializer):
-    # This determines whether the annotation is a leaf of our
-    # content tree.
-    IS_LEAF = False
+class NestableAnnotationField(serializers.Field):
+    def get_attribute(self,
+                      instance: NestableAnnotation) -> NestableAnnotation:
+        return instance
 
-    content_type = serializers.SerializerMethodField()
-    inlines = serializers.SerializerMethodField()
-    text = serializers.SerializerMethodField()
 
-    @property
-    def CONTENT_TYPE(self):     # noqa; this is a constant
-        raise NotImplementedError('Unknown annotation type')
+class InlinesField(NestableAnnotationField):
+    def __init__(self, is_leaf_node: bool):
+        super().__init__()
+        self.is_leaf_node = is_leaf_node
+
+    def to_representation(self,
+                          instance: NestableAnnotation) -> List[PrimitiveDict]:
+        if self.is_leaf_node:
+            return []
+        return NestedAnnotationSerializer(
+            instance.children, context=self.context, many=True).data
+
+    def to_internal_value(self,
+                          data: List[PrimitiveDict]) -> List[PrimitiveDict]:
+        if not self.is_leaf_node:
+            serializer = NestedAnnotationSerializer()
+            return [
+                serializer.to_internal_value(item) for item in data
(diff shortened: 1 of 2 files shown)


toolness:
Oh, I didn't fiddle with the `content_type` field at all because I was starting to get confused and figured we could save that for a separate PR.

api/document/serializers/content.py
```diff
@@ -102,59 +102,91 @@ def to_internal_value(self, data: PrimitiveDict) -> PrimitiveDict:
...
+        return NestedAnnotationSerializer(
+            instance.children, context=self.context, many=True).data
+
+    def to_internal_value(self,
+                          data: List[PrimitiveDict]) -> List[PrimitiveDict]:
+        if not self.is_leaf_node:
+            serializer = NestedAnnotationSerializer()
+            return [
+                serializer.to_internal_value(item) for item in data
+            ]
```
toolness: Hmm, I'm now realizing that it might be possible to just pass this list to the serializers constructor, with that weird `many=True` option.

api/document/serializers/content.py
```diff
@@ -102,59 +102,91 @@ def to_internal_value(self, data: PrimitiveDict) -> PrimitiveDict:
...
     def doc_node_text(self):
         return self.context['cursor'].text
 
+    def to_representation(self, instance: NestableAnnotation) -> str:
+        return self.doc_node_text[instance.start:instance.end]
+
+    def to_internal_value(self, data: str) -> str:
+        if self.is_leaf_node:
+            serializer = serializers.CharField(trim_whitespace=False)
+            return serializer.to_internal_value(data)
```
toolness: I think that we could potentially just have this field inherit from `CharField`, but there's so much idiosyncratic logic in here that it was easier for me to just instantiate one during this method instead... I could switch things up, though.

api/document/serializers/content.py
cmc333333: I suspect that the constructor's `data` param is only used for serializing. Would this work instead?
```python
serializer = NestedAnnotationSerializer(many=True)
return serializer.to_internal_value(data)
```

api/document/serializers/content.py
```diff
@@ -102,59 +102,91 @@ def to_internal_value(self, data: PrimitiveDict) -> PrimitiveDict:
...
+
+    def to_internal_value(self,
+                          data: List[PrimitiveDict]) -> List[PrimitiveDict]:
+        if not self.is_leaf_node:
+            serializer = NestedAnnotationSerializer()
+            return [
+                serializer.to_internal_value(item) for item in data
+            ]
+        elif data:
+            raise ValidationError('leaf nodes cannot contain nested content')
```
cmc333333: "Leaf" seems like an internal term. What do you think about
```python
content_type = data.get('content_type', 'leaf')
raise ValidationError(f'{content_type} content can't contain nested content')
```

api/document/serializers/content.py
cmc333333: Good point! I'm also torn; maybe we keep it like this until/if we find ourselves wanting to set more validation properties (maybe some annotations can only be x chars long)?

api/document/serializers/content.py
```diff
@@ -102,59 +102,91 @@ def to_internal_value(self, data: PrimitiveDict) -> PrimitiveDict:
...
+        # If this isn't a leaf node, we're not going to complain,
+        # because we do fill out this value on non-leaf nodes during
+        # serialization, and we want clients to be able to modify
+        # serialized responses and send them back without having to
+        # do too much work.
+        #
+        # However, that said, we *will* throw away the text value on
+        # non-leaf nodes during deserialization, to make sure it's not
+        # accidentally used by anything further down the pipeline.
+        return ''
```
cmc333333: Another option would be to remove the text field on non-leaf nodes. That'd complicate our logic (in the UI and during serialization) a bit more, but if it makes a better API, it might be worth it. If we do something like that, it probably'll make sense to split the serializer implementations into different classes. Something to ponder.

api/document/serializers/content.py
toolness: O yea, it looks like that does work! Sweet.

api/document/serializers/content.py
toolness: Uhh, one annoying thing is that because `InlinesField`'s `to_internal_value()` is only being passed the value of `inlines` and not the parent object, um, I can't get ahold of that `content_type` unless I do some weird context gymnastics in the parent serializer.  Blaargh.  REST API is weird.

PR #841, opened 2018-01-11
Share styles between ui and api

cmc333333:
This creates a `shared-styles` directory in the root of the repo which is sym-linked in both the ui and api. The strategy works locally because Docker mounts the whole repository into each container. To make it work in production (tested [here](https://circleci.com/gh/18F/omb-eregs/1230)), this runs `collectstatic` in CI (rather than in cloud.gov), then sends up the static files.

It's kinda round-about but works; happy to hear other ideas.


tadhg-ohiggins:
Out of curiosity, did your test check for handling duplicate filenames? I think Django handles this by only using the first file with a specific name that it finds, and I'm not quite sure whether or not this could end up being a problem in this case.

cmc333333:
@tadhg-ohiggins indeed, Django'll figure out a precedence order based on which app is listed first. In this case, the new styles are in a `shared-styles` subdir, so as long as that doesn't conflict with another app, we should be claiming that namespace.

cmc333333:
This was included in #854

Issue #819, opened 2018-01-08
Status flagging system 


yowill:
Set up status flagging system and filtering for “Imported” “Editing” “Ready for review” "Published"

tadhg-ohiggins:
My thinking is that there are two potential workflows we need to cover:

## Import Workflow

0.  Policy exists in DB, but no document is associated with it. (blank)
1.  Text has been imported, but not checked/cleaned up. ('Cleanup')
2.  Text has been checked/cleaned up and is ready for review. ('Review')
3.  Text has been published. ('Published')

## Edit Workflow

0.  Policy exists in DB, but no document is associated with it. (blank)
1.  Text is being created/edited in the tool, but is not yet ready for
    review. ('Edit')
2.  Text is ready for review. ('Review')
3.  Text has been published. ('Published')

This gives us four potential workflow phases that a policy could be in:
(blank)
`Cleanup`
`Edit`
`Published`
`Review`

toolness:
This looks good to me, I like it!  Maybe we should go over it w/ @austinhernandez at our MVP features list later today?

toolness:
@tadhg-ohiggins one thing I just realized, which might have to do with this and/or #830, is the ability for us to be able to detect whether a  document has been edited at all since being bulk-imported. If it hasn't been, and if we improve our PDF importing logic in the interim, we'll likely want to re-import all the untouched documents, so that once someone _does_ get around to editing them, they'll be using the best-possible import.

tadhg-ohiggins:
@toolness I just saw this comment; I'm not sure it makes sense for `workflow_phase` to track this (although it might), but in any case it won't fit in the current set of phases. Do you want to create a new issue for it?

yowill:
Demoed and closed!

PR #842, opened 2018-01-11
Add workflow phase field

tadhg-ohiggins:
Even workflow phases
Need names, just like the seasons.
Grey falls winter rain.

Adds a `workflow_phase` field to Policy objects, addressing https://github.com/18F/omb-eregs/issues/819.


api/ombpdf/tests/test_management_commands_import_pdfs_from_policies.py
```diff
@@ -96,3 +96,11 @@ def test_import_pdfs_from_policies(monkeypatch):
...
     assert failures == {f'{p4.pk}: exception-but-a-pdf'}
     # p2 doesn't show up as it's not a pdf
+
+    # make sure the policy objects have the correct workflow_phases:
+    p1_db = Policy.objects.get(id=p1.pk)
+    assert p1_db.workflow_phase == WorkflowPhases.cleanup.name
+    p3_db = Policy.objects.get(id=p3.pk)
+    assert p3_db.workflow_phase == WorkflowPhases.cleanup.name
+    p4_db = Policy.objects.get(id=p4.pk)
+    assert p4_db.workflow_phase == WorkflowPhases.failed.name
```
cmc333333: Rather than `Policy.objects.get`, do you think this is clearer:
```python
p1.refresh_from_db()
assert p1.workflow_phase == WorkflowPhases.cleanup.name
...
```

api/reqs/models.py
```diff
@@ -69,6 +69,34 @@ class PolicyTypes(Enum):
...
 
 
+@unique
+class WorkflowPhases(Enum):
+    """
+    We're assuming two potential workflows.
+
+    Import Workflow
+
+    0.  Policy exists in DB, but no document is associated with it. (blank)
```
cmc333333: Did you consider have a value for this state rather than keeping it blank? It seems odd to have 6 workflow states but only 5 can be accessed through the enum. Easy enough to add a default value on our field which would backfill data.

toolness:
This looks nice!

> I think we'll need to apply something similar in the `pdf_to_db` and `scrape_memoranda` commands as well.

Would it be better to move the related logic into the model, or some utility, and just have all relevant code call into that?

Relatedly, I recently ran across a [blog post of Tom Christie's](https://www.dabapps.com/blog/django-models-and-encapsulation/) that advises the following:

> Never write to a model field or call `save()` directly. Always use model methods and manager methods for state changing operations.

I thought that was interesting but don't know if it's a policy we should try to follow or not.

api/ombpdf/tests/test_management_commands_import_pdfs_from_policies.py
tadhg-ohiggins: Yes, I think that's a lot clearer, and will make that change.

tadhg-ohiggins:
@toolness The `model.Manager` approach is interesting, in this case it looks like we'd add setters to a PolicyManager class and make all changes use the `reversion.create_revision()` context?

I'm not sure we want to go that way, though it's worth considering, but for now I'm going to punt on that and add the same approach to `pdf_to_db` and `scrape_memoranda`.

api/ombpdf/management/commands/pdf_to_db.py
```diff
@@ -23,8 +25,13 @@ def handle(self, *args, **options):
         policy_name = options['policy'] or Path(fp.name).stem.upper()
         policy = fetch_policy(policy_name)
 
-        if policy is None:
-            raise CommandError(f"Policy '{policy_name}' not found.")
+        with reversion.create_revision():
+            if policy is None:
+                raise CommandError(f"Policy '{policy_name}' not found.")
+                policy.workflow_phase = WorkflowPhases.failed.name
```
cmc333333: `policy` is `None` here -- won't this fail? Would it make sense to set the workflow phase to `failed` around original line 30 and `cleanup` at the end of the file? If an exception's thrown, we'd stay in the `failed` state.

api/ombpdf/management/commands/scrape_memoranda.py
```diff
@@ -42,13 +43,19 @@ def parse_pdf(policy: Policy, url: Url) -> bool:
         pdf = download_with_progress(url)
         pdf.name = basename(urlparse(url).path)     # this used by from_file
         doc = OMBDocument.from_file(pdf)
+        with reversion.create_revision():
+            policy.workflow_phase = WorkflowPhases.cleanup.name
+            policy.save()
```
cmc333333: Should this probably happen after the `to_db` line, in case something goes wrong there?

tadhg-ohiggins: Yup, it definitely should.

PR #851, opened 2018-01-11
Add documentation for DocNode, Annotation, and DocCursor (w/ doctests)

toolness:
This documents our `DocCursor` model a bit more.  I ended up also adding a `__str__()` implementation to it (though we could rename it to `describe()` or something if we don't like the idea of `__str__()` returning multiple lines of text) to make it easier to see what a cursor is "pointing" at.

Note that the documentation is tested via a weird `doctest_test.py` that's largely copy-pasted from #758.  As mentioned in that issue, it's a workaround because enabling doctest finding in pytest across the whole codebase results in a mysterious segfault. What's also unfortunate is that if the test fails, it produces unhelpful output; the most helpful output can be found by running `pytest --doctest-modules document/tree.py` (which doesn't segfault because it's only looking at one file).

**Update:** This PR also adds some documentation to `document.models`.

### DIFF
--- a/api/document/tree.py
+++ b/api/document/tree.py
@@ -22,25 +22,18 @@ class DocCursor():
 
     For example, here we'll create a document structure:
 
-        >>> root = DocCursor.new_tree('root')
+        >>> root = DocCursor.new_tree('root', title='my cool doc')
         >>> sec1 = root.add_child('sec', '1', text='section 1')
         >>> para = sec1.add_child('para', 'a', text='paragraph a')
 
-    Here's its structure:
+    Each of the above variables is a DocCursor pointing to a
+    part of the document. We can see where they're pointing:
 
-        >>> print(root)
-        <root identifier="root_1" type_emblem="1">
-          <sec identifier="root_1__sec_1" type_emblem="1">
-            section 1
-            <para identifier="root_1__sec_1__para_a" type_emblem="a">
-              paragraph a
-            </para>
-          </sec>
-        </root>
-
-    Note that the above XML doesn't conform to any particular schema; it's
-    just a familiar notation we're using to visualize the structure of
-    the document.
+        >>> print(sec1)
+        parent: root_1 title="my cool doc"
+        |- sec_1
+           'section 1'
+           |- para_a
 
     The underlying DocNode model can be revealed through the `.model`
     property:
@@ -166,31 +159,43 @@ class DocCursor():
                 yield cursor
 
     @property
-    def _opening_tag(self):
+    def _parent_desc(self):
+        parent = self.parent()
+        if parent is None:
+            return "None"
+        return f"{parent.identifier}{parent._attr_str}"
+
+    @property
+    def _attr_str(self):
         attr_strs = []
-        for attr in ['identifier', 'marker', 'title', 'type_emblem']:
+        for attr in ['title', 'marker']:
             val = getattr(self, attr, None)
             if val:
                 attr_strs.append(f'{attr}="{val}"')
-        attr_str = ' ' + ' '.join(attr_strs) if attr_strs else ''
-        return f'<{self.node_type}{attr_str}>'
+        return (' ' + ' '.join(attr_strs)) if attr_strs else ''
 
     @property
-    def _closing_tag(self):
-        return f'</{self.node_type}>'
+    def _short_desc(self):
+        return f"{self.node_type}_{self.type_emblem}{self._attr_str}"
 
-    def _describe(self, indent=''):
-        lines = [f'{indent}{self._opening_tag}']
-        next_indent = indent + '  '
-        if self.text:
-            lines.append(f'{next_indent}{self.text}')
-        for child in self.children():
-            lines.extend(child._describe(next_indent))
-        lines.append(f'{indent}{self._closing_tag}')
-        return lines
+    def _short_text(self, max_len=40):
+        text = self.text
+        if len(text) > max_len:
+            text = text[:max_len] + '...'
+        return repr(text)
 
     def __str__(self):
-        return '\n'.join(self._describe())
+        indent = ''
+        lines = [
+            f"{indent}parent: {self._parent_desc}",
+            f"{indent}|- {self._short_desc}",
+        ]
+        next_indent = indent + '   '
+        if self.text:
+            lines.append(f'{next_indent}{self._short_text()}')
+        for child in self.children():
+            lines.append(f'{next_indent}|- {child._short_desc}')
+        return '\n'.join(lines)
 
     def nested_set_renumber(self, left=1, bulk_create=True):
         """The nested set model tracks parent/child relationships by requiring
--- a/api/document/tree.py
+++ b/api/document/tree.py
@@ -10,9 +10,64 @@ PrimitiveDict = Dict[str, Any]
 
 
 class DocCursor():
-    """DocNodes don't keep track of their children/relationships within a
+    """
+    DocNodes don't keep track of their children/relationships within a
     tree, so we will generally access them indirectly through a wrapping tree
-    (a DiGraph). This DocCursor points to a specific node within that tree."""
+    (a DiGraph). This DocCursor points to a specific node within that tree.
+
+    Note, though, that the DocCursor doesn't actually need to be
+    backed by a database at all, and can be constructed independently
+    of one. In fact, this a convenient way to create DocNodes before
+    committing them to a database.
+
+    For example, here we'll create a document structure:
+
+        >>> root = DocCursor.new_tree('root')
+        >>> sec1 = root.add_child('sec', '1', text='section 1')
+        >>> para = sec1.add_child('para', 'a', text='paragraph a')
+
+    Here's its structure:
+
+        >>> print(root)
+        <root identifier="root_1" type_emblem="1">
+          <sec identifier="root_1__sec_1" type_emblem="1">
+            section 1
+            <para identifier="root_1__sec_1__para_a" type_emblem="a">
+              paragraph a
+            </para>
+          </sec>
+        </root>
+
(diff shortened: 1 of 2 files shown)


toolness:
Um... there doesn't seem to be a way to write the document to the DB if one didn't give it a policy when they initially called `new_tree()`.  For the sake of clear documentation, do you mind if I add an optional `policy=None` arg to the end of `nested_set_renumber()` that assigns the policy to all child nodes before saving?

toolness:
To clarify on my last comment... I'm used to APIs and documentation that allow one to "play around" with a model (or model-esque thing, which is what `DocCursor` appears to be) with minimal setup, and then commit it to a database later if they want. I basically want to add a section at the end of my docs that's like, "if you decide you're happy with your document and want to save it to the database, just do this!"

However, in order to be saved to the database, it needs to have a policy associated with it, right? As far as I can tell, the only way to associate a policy with all the `DocNode`s in a `DiGraph` is to manually walk through all the nodes in the document and assign the policy to them, which feels cumbersome.  I mean, I could also modify the documentation to simply create a policy at the very beginning and pass it into `new_tree` so that all the `DocNode`s have a policy from the very beginning, but this feels counter to the philosophy of being able to play around with an API with minimal setup (it also raises the question of why omitting `policy` from a call to `new_tree()` is ever something one would ever want to do)...  Does that make sense?

toolness:
Alternatively, instead of adding `policy` as an optional argument to `nested_set_renumber`, what if we just added another method to `DocCursor` called `set_policy()`?  It would almost exclusively be used by the documentation, I guess, but it'd enable the kind of approach/philosophy I described in my last comment.

toolness:
As expected, writing this documentation is also forcing me to confront things I don't currently understand about `DocCursor` and `DocNode`:

* In `DocNode`, is `left` and `right` essentially the count of the node's previous and next siblings?  If so, do you mind if I add that as either comments or `help_text` to those fields?  Because I've been super confused by what those mean...

* In `DocCursor`, the name `nested_set_renumber` feels confusing.  I guess this is probably because of my confusion with `left` and `right`, which is what it's tinkering with... But from observing the contexts in which the method is used, I've come to think of it as "that one method I call to prepare the document for saving and optionally save it."  This makes me wonder if there's a better name for it, though I don't really know what that would be.  Maybe just surfacing what exactly these `left` and `right` properties are in the documentation will be enough to make the purpose of this method clearer, though.

toolness:
I think part of what's confusing me about these classes is that I have pre-existing assumptions about the way documents are structured, largely influenced by the Document Object Model (DOM).  A lot of the properties and methods across the different classes, like `depth` and `add_child`, fit in with that mental model, but others don't, like the way text/annotations are done, and this left/right thing I still don't understand.  So I'm hoping these docs will help clear up any misconceptions newcomers have about how these APIs work.

cmc333333:
Regarding the policy problem, I think we want the documentation to match how we intend the class to be used.
* If we are thinking we will always have a policy when creating the document, we should make the policy a required argument on `new_tree` and adjust all the tests to send a `Mock`
* If we think there's a situation where we don't have the policy until _after_ creating the document tree (e.g. maybe during the import, we attempt to pull policy info out of the doc?), let's create that method/extend `nested_set_renumber` now.

Regarding `left`, `right`, and `nested_set_renumber`, we're following the [Nested Set Model](https://en.wikipedia.org/wiki/Nested_set_model) for storing the hierarchy in a flat database. For the most part, you shouldn't want to worry about those fields; they're for accounting. We originally had `nested_set_renumber` separate from the `save`, but we called them together so frequently it made sense to combine them.

cmc333333:
Copying from DMs:
Do you think the position in a tree is important to describing a doc cursor? the xml notation suggested is only displaying children, which answers one type of question, but doesn’t feel particularly encompassing. What about something like:
```python
>>> print(sec)
parent: full__path__root_1 title=M-16-19: Something something
|- sec_1 title=Some section
    "Text cut at xxx charact..."
    |- para_a marker=(a)
    |- para_b marker=(b) title=Paragraph title
```
something that’s give the context of this cursor’s position in the document tree without actually serializing that doc. In particular, it only looks at the node’s children and parent, and only prints (some) of the current node’s text

toolness:
Noice sounds good!  Yeah, I like the `__str__` suggestion, will implement that.

api/document/tree.py
```diff
@@ -125,45 +217,51 @@ def nested_set_renumber(self, left=1, bulk_create=True):
         if bulk_create:
             self._bulk_create()
 
-    def _bulk_create(self):
+    def _bulk_create(self) -> None:
         DocNode.objects.bulk_create(node.model for node in self.walk())
 
-    def next_sort_order(self):
+    def next_sort_order(self) -> int:
         return self.tree.out_degree(self.identifier)
```
toolness: I _think_ this returns an `int`. According to the [`out_degree` documentation](https://networkx.github.io/documentation/latest/reference/classes/generated/networkx.DiGraph.out_degree.html#networkx.DiGraph.out_degree), it could also return an iterator if multiple nodes are requested, but it seems like we're only requesting one node here.

api/document/tree.py
```diff
@@ -125,45 +217,51 @@ def nested_set_renumber(self, left=1, bulk_create=True):
...
         """Convert a (linear) list of DocNodes into a tree-aware version.
         We assume that the models are already sorted."""
         parent = self
         for child in models:
             # not a child of this parent; move cursor up
             while child.left > parent.right:
-                parent = parent.parent()
+                parents_parent = parent.parent()
+                if parents_parent is None:
+                    raise AssertionError("Parent's parent must exist!")
```
toolness: Looks like mypy found an edge case here! I'm type narrowing and raising an error if `parent.parent()` is `None`, but I'm not sure what the best assertion message to raise is, since I'm still not very familiar with the nested set algorithm--any suggestions?

api/document/tree.py
```diff
@@ -8,11 +8,66 @@
 
 PrimitiveDict = Dict[str, Any]
 
+T = TypeVar('T', bound='DocCursor')
```
toolness: Note that we're using a [type variable with an upper bound](http://mypy.readthedocs.io/en/latest/generics.html#type-variable-upper-bound) here to ensure that our `DocCursor` methods that return instances of `DocCursor` are actually typed to the sub-class of the `DocCursor` they come from.  Otherwise e.g. `JSONAwareCursor.new_tree()` will actually return a `DocCursor` rather than a `JSONAwareCursor`.

toolness:
Ok, I think I'm going to stop here for now--the combined doctests and type annotations comprise a decent improvement on their own, and I'd like to deal with the "saving to the db" part in a separate PR, since it is going to involve changing the type signature of `new_tree()`, which could result in lots of changes elsewhere in the codebase.

api/document/tree.py
cmc333333: You got it! We use this number to tag the edges between nodes. This allows us to fetch child nodes in order (as trees don't have an inherent order but documents do).

api/document/tree.py
```diff
@@ -212,45 +212,51 @@ def nested_set_renumber(self, left=1, bulk_create=True):
...
 
-    def parent(self):
+    def parent(self) -> Optional['DocCursor']:
         predecessors = list(self.tree.predecessors(self.identifier))
         if predecessors:
             return type(self)(self.tree, predecessors[0])
+        return None
 
-    def ancestors(self, filter_fn=None):
+    def ancestors(self, filter_fn: Callable[['DocCursor'], bool]=None) \
```
cmc333333: Should this be `Optional[Callable[['DocCursor'], bool]]`?

api/document/tree.py
cmc333333: Awesome. I think this case could arise if either:
1. The DocNodes weren't given in the right sort order
2. The DocNodes don't have sensible left/right values (they didn't go through our nested set accounting)

So maybe something along the lines of:
"Couldn't convert DocNodes into a tree. Perhaps they weren't sorted or there was data corruption?"

api/document/tree.py
cmc333333: I know type vars are usually represented by single letters, but it makes the method type annotations hard to read, in my opinion. Would it be a bad idea to do something like
```python
DocCursorSubType = TypeVar('DocCursorSubType', bound='DocCursor')
```
?

api/document/tree.py
```diff
@@ -125,45 +219,50 @@ def nested_set_renumber(self, left=1, bulk_create=True):
...
 
-    def parent(self):
+    def parent(self: T) -> Optional[T]:
         predecessors = list(self.tree.predecessors(self.identifier))
         if predecessors:
             return type(self)(self.tree, predecessors[0])
+        return None
 
-    def ancestors(self, filter_fn=None):
+    def ancestors(self: T, filter_fn: Callable[[T], bool]=None) \
```
cmc333333: (Moving from a comment on a changeset)
Should this be `Optional[Callable[['DocCursor'], bool]]`?

api/document/tree.py
toolness: Yeah, I like that better than `T`.  Although I cry at how this will affect our 80-character line limit, since all methods using this type will need to repeat `DocCursorSubType` twice 😣.

api/document/tree.py
toolness: mypy actually infers the `Optional` from the fact that we're defaulting it to `None`, hooray for less verbosity!

api/document/tree.py
toolness: That said, though, I'd also argue that since the `TypeVar` *does* appear twice on method signatures, I think it actually compensates for the opaque-ness of the type variable name, as readers can actually just infer from the `self: T` or `cls: Type[T]` that it's referring to the parent class; they don't even really need to find the definition of `T`, in some sense. How about we leave it as-is and change it if folks find it confusing?

PR #854, opened 2018-01-12
Initial markup for admin menu

cmc333333:
Building on #841 (new commits begin at debad57), this adds a quick pass at a sidebar navigation in the Django admin. It's not pretty, but it should be a place for @rememberlenny to expand from.

We _may_ want to hide this behind a feature flag, I'm not positive.

Looks like 
screen shot 2018-01-12 at 2 43 37 pm


cmc333333:
I'm going to refactor to make the functions simpler, then squash the diff, one moment.

cmc333333:
Thanks @tadhg-ohiggins ! I'm going to rebase and merge this (which will also grab #841).

PR #856, opened 2018-01-12
Moar document tests

toolness:
This improves `document.tests.serialzers`' coverage over `document.serializers`, and `document.tests.json_importer` coverage over `document.json_importer`.

It also moves `document.tests.json_importer.factories` up a level, so that other test suites can use it.

Coverage can be examined for the packages via:

```
pytest --cov=document.serializers --cov-report term-missing document/tests/serializers/
pytest --cov=document.json_importer --cov-report term-missing document/tests/json_importer/
```


toolness:
Arrg, so coverage is at 100% (although obviously coverage doesn't mean _too_ much) but it's revealed the curmudgeonny nature of the deserializer, e.g. it expects `inlines: []` even on text nodes, which can't actually have any inlines.

api/document/tests/json_importer/importer_test.py
```diff
@@ -13,3 +18,25 @@ def test_convert_paragraph_works():
...
+    assert para.text == 'Hello there'
+    annos = [anno for anno in para.model.annotations()]
+    assert len(annos) == 1
+    assert annos[0].href == 'http://example.org/'
+
+    para2 = import_json_doc(policy, para_primitive)
+    assert para.pk != para2.pk
+
+    with pytest.raises(DocNode.DoesNotExist):
+        assert para.model.refresh_from_db()
```
cmc333333: This took me a few re-reads to understand. What do you think about
```python
assert not DocNode.objects.exists(pk=para.pk)
```

api/document/tests/serializers/doc_cursor_test.py
```diff
@@ -275,3 +287,18 @@ def test_content_outside():
     assert [c['content_type'] for c in content] == [
         'external_link', '__text__', 'external_link',
     ]
+
+
+def test_children_field_to_internal_value_works():
+    # TODO: We really shouldn't *have* to supply a type_emblem here,
+    # because the deserializer will ultimately pass the payload through
+    # DocCursor, which will auto-assign one.
```
cmc333333: I think we can replace the auto-generated `type_emblem` field of `DocCursorSerializer` with one that allows blank.

api/document/tests/json_importer/importer_test.py
toolness: Ah good idea! That didn't work (`exists` didn't like the keyword arg) but this did:

```python
assert not DocNode.objects.filter(pk=para.pk).exists()
```

api/document/tests/serializers/doc_cursor_test.py
toolness: Cool... this is making me start to wonder if it's better to just make `DocCursorSerializer` a standard `Serializer`, rather than turn it into a `DocNode` serializer as I'd suggested in an earlier PR.  It seems a number of the validations performed by the `DocNode` model aren't actually going to be needed by the deserializer due to all the stuff it's auto-populating.

PR #858, opened 2018-01-16
Add package-lock file.

cmc333333:
This ensures our dependency versions are identical; see
https://before-you-ship.18f.gov/infrastructure/pinning-dependencies/

Heads up that this includes tslint (see upcoming PR).


cmc333333:
Thanks @toolness

PR #860, opened 2018-01-16
Add tslint for editor JS

cmc333333:
This makes our typescript look a bit more similar to our ES6 and ensures it is linted for common logic errors. We can add [prettier](https://prettier.io/) on top of this when desired.


cmc333333:
Thanks @toolness . I'll rebase and merge

PR #861, opened 2018-01-16
Decouple the DocCursorSerializer from ModelSerializer

toolness:
This attempts to decouple the `DocCursorSerializer` from DRF's `ModelSerializer`. For rationale, see the included docstring for `docnode_field_attrs`.

I was really hoping this would get rid of our need for the ugly `_prefetched_objects_cache` in `DocCursor`, but that is actually accessed by DRF's `UpdateModelMixin` (inherited by our `TreeView` view), _not_ `ModelSerializer`.  I still think the decoupling in this PR may generally be a good idea though... and I can issue a future PR (or add onto this one) that decouples us from the `UpdateModelMixin`.

### DIFF
--- a/api/document/serializers/doc_cursor.py
+++ b/api/document/serializers/doc_cursor.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import Any, Dict, List
 
 from rest_framework import serializers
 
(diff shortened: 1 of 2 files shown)


api/document/serializers/doc_cursor.py
```diff
@@ -49,31 +49,65 @@ def to_internal_value(self,
...
+        'blank': 'allow_blank',
+    }
+    metadata = DocNode._meta.get_field(field)
+    result = {
+        attr_map.get(attr, attr): getattr(metadata, attr)
+        for attr in attrs
+    }
+    if result.get('allow_blank'):
+        result['required'] = False
+    return result
```
toolness: Blerg, I guess this helper is confusing/bloaty enough that we can just get rid of it and do things more manually/explicitly.  I just wanted to reduce DRY by not repeating stuff from the model too much...

cmc333333: Any chance we can leverage utilities from DRF? I agree that this is probably more confusing that being redundant.

toolness: Unfortunately, not as far as I can tell. It's ok though, I'm just gonna get rid of this funky function and make things more explicit while still preserving DRY as much as possible.

PR #865, opened 2018-01-16
Lazily create menu archetype.

### LINKED ISSUE #863: Running individual tests that modify Django's URLconf explode
I noticed when working on just the `document` package today that running `pytest document/` fails on `document.tests.views_test.test_editor_works` with an _extremely_ long error log culminating in the following:

```
django.urls.exceptions.NoReverseMatch: 'admin' is not a registered namespace
```

However, running `pytest` by itself, without any arguments, works fine.

After some sleuthing I realized this is due to #854, which added the following top-level code in a template tag module:

```python
menu_archetype = MenuItem.new('', children=(
    MenuItem.new('Policies', children=(
        MenuItem.new('View all policies',
                     url=reverse('admin:reqs_policy_changelist')),
        MenuItem.new('Add new', url=reverse('admin:reqs_policy_add')),
    )),
    MenuItem.new('Users', url=reverse('admin:auth_user_changelist')),
    # Settings depends on the user id, so is added later
))
```

The culprit is the use of `reverse()` on an admin URL.  Because `test_editor_works` changes Django's URLconf to `document.urls`, which doesn't include the admin package, `admin` is no longer a valid URL namespace at the time that Django initializes itself, which is when it loads all the templatetags.  Commenting out the above code fixes the problem.

I think one solution to this might be to make `menu_archetype` a lazily-instantiated module-level variable, so it starts out at `None` and only acquires its actual value the first time its template tag is used.  What do you think @cmc333333 ?

toolness:
Fixes #863.

### DIFF
--- a/api/ereqs_admin/templatetags/nav_menu.py
+++ b/api/ereqs_admin/templatetags/nav_menu.py
@@ -1,3 +1,4 @@
+from functools import lru_cache
 from typing import NamedTuple, Optional, Tuple
 
 from django import template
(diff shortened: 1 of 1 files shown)


api/ereqs_admin/templatetags/nav_menu.py
```diff
@@ -32,23 +33,30 @@ def append_child(self, child):
...
-    MenuItem.new('Policies', children=(
-        MenuItem.new('View all policies',
-                     url=reverse('admin:reqs_policy_changelist')),
-        MenuItem.new('Add new', url=reverse('admin:reqs_policy_add')),
-    )),
-    MenuItem.new('Users', url=reverse('admin:auth_user_changelist')),
-    # Settings depends on the user id, so is added later
-))
+@lru_cache()
+def get_menu_archetype():
```
toolness: It does feel a bit silly to create an LRU cache for a function with no arguments, but um, it seems to get the job done? 🤷‍♂️

PR #866, opened 2018-01-16
Add migration to change theme styles for form

rememberlenny:
# What

Add migration against the `ereqs_admin` app to update default admin theme settings. Also add CSS for form to match font and style of @austinhernandez and @rtwell designs.

screen shot 2018-01-16 at 1 21 36 pm
screen shot 2018-01-16 at 2 23 58 pm
screen shot 2018-01-16 at 4 07 10 pm
screen shot 2018-01-16 at 4 12 11 pm


api/ereqs_admin/migrations/0002_auto_20180116_2105.py
```diff
@@ -0,0 +1,51 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.5 on 2018-01-16 21:05
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def forward(apps, schema_editor):
+    Theme = apps.get_model('admin_interface', 'Theme')
+    Theme.objects.create(
```
cmc333333: This creates a second theme in the database. Shouldn't we be modifying the existing entry instead? That'd allow us to specify only the fields that are changing.

api/ereqs_admin/migrations/0002_auto_20180116_2105.py
```diff
@@ -0,0 +1,51 @@
...
+        css_module_link_color="#C2C2C2",
+        css_module_link_hover_color="#E1F3F8",
+        css_module_rounded_corners=True,
+        css_generic_link_color="#205493",
+        css_generic_link_hover_color="#0071BC",
+        css_save_button_background_color="#3DA4C8",
+        css_save_button_background_hover_color="#3794B5",
+        css_save_button_text_color="#FFFFFF",
+        css_delete_button_background_color="#CD2026",
+        css_delete_button_background_hover_color="#981B1E",
```
cmc333333: Several of these colors aren't part of our palette. Can we use our existing palette in place of any of these variations? If we can't combine them, we should extend the palette to include them by adding new color variables to `_colors.scss`.

* E1F3F8
* 222222
* C2C2C2
* 205493
* 0071BC
* 3DA4C8
* 3794B5
* CD2026
* 981B1E

api/ereqs_admin/migrations/0002_auto_20180116_2105.py
```diff
@@ -0,0 +1,51 @@
...
+        recent_actions_visible=True
+    )
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('ereqs_admin', '0001_initial'),
+    ]
+
+    operations = [migrations.RunPython(forward)]
```
cmc333333: It's best practice to also have a `backward` to revert these changes.

api/ereqs_admin/static/admin/override.scss
```diff
@@ -7,6 +7,10 @@
 $banner-height: 60px;
 $nav-menu-width: 200px;
 
+* {
+  box-sizing: border-box;
```
cmc333333: This subtly changes the admin -- I see it reducing the size of the content pane and slightly shortening the length of many text boxes.  Just to push back a little, are we sure this is necessary? I'm worried it'll break something unexpected.

api/ereqs_admin/static/admin/override.scss
```diff
@@ -46,3 +50,134 @@ $nav-menu-width: 200px;
   position: absolute;
   width: $nav-menu-width;
 }
+
+.app-admin_interface,
+.admin-interface {
+  background: #F0F0F0;
+}
+
+.object-tools a {
```
cmc333333: When overriding styles within the admin, let's comment what they're changing. Otherwise, there's not a good way to figure out why these changes were made.

api/ereqs_admin/static/admin/override.scss
```diff
@@ -46,3 +50,134 @@ $nav-menu-width: 200px;
...
+    padding: 1rem;
+  }
+
+  td,
+  th,
+  thead th,
+  tfoot td {
+    line-height: 1.5rem;
+    padding: 1rem;
+    font-size: 1rem;
```
cmc333333: Let's use our existing color, font-size, and spacing variables instead of writing one-offs.

api/ereqs_admin/static/admin/override.scss
```diff
@@ -46,3 +50,134 @@ $nav-menu-width: 200px;
...
+  }
+
+  .form-row [type="number"],
+  .form-row [type="text"] {
+    // width: 100%;
```
cmc333333: Should this be commented out? If so, let's remove the rule.

api/ereqs_admin/static/admin/override.scss
```diff
@@ -46,3 +50,139 @@ $nav-menu-width: 200px;
...
+}
+
+.module {
+  background: none;
+}
+
+#content  h1 {
+  font-weight: 500;
+  color: $color-black;
+  font-size: $space-lg;
```
cmc333333: Rather than use the `space` variables for font size, let's use (and add) mixins from `type-size`. We'll generally want to set the line height whenever we're setting font size. Setting the line height will also likely resolve some of the need for other padding.

Issue #817, opened 2018-01-08
Create an Akoma Ntoso PUT endpoint in the API


toolness:
An Akoma Ntoso GET endpoint was added in #795, and given that the ProseMirror-based WYSIWYG editor prototyped in #798 won't be fully-functional and able to edit all aspects of a policy document for a while, we should consider creating a GET endpoint that allows staff to edit the XML and submit changes, as a kind of "advanced/raw editing" capability. On the front-end we could use CodeMirror to make the XML editing process a bit friendlier.

Thoughts?

toolness:
Honestly, doing this would be a lighter lift than the full-fledged WSIWYG editor--I'm curious if we should consider building it _before_ the fancy editor.

toolness:
Oh, this was fixed by ~~#868~~ #815!

Issue #839, opened 2018-01-10
Document PUT API doesn't allow changes to meta information


toolness:
We added a PUT endpoint to documents in #815, but we don't currently support writing to the `meta` endpoint.

I'm actually not sure how urgent this is, though.  I think I recall @cmc333333 telling me something about how `meta` on non-root document nodes could be used to store arbitrary stuff, but we're not currently using it for anything. And for the root note, we include policy details, but it's possible that changing that information is more easily done through the Django admin.

PR #868, opened 2018-01-17
Add Akoma Ntoso XML parser

toolness:
This adds a super simple [Django REST Framework parser](http://www.django-rest-framework.org/api-guide/parsers/) for Akoma Ntoso, which consumes the kind of XML output by our [Akoma Ntoso renderer](https://github.com/18F/omb-eregs/pull/795) and outputs a primitive dictionary that can be passed to our `DocCursorSerializer` for deserialization.

The conversion process is fairly liberal and basically converts any XML attributes into properties in the corresponding dictionary for the XML node.  If we want to be more strict, we can be, but my assumption was that the serializer would take care of ensuring that the passed-in data is actually valid.

This is a step towards #817.  Before exposing this parser to the outside world, we should probably also add `defusedxml` to ensure that we're not vulnerable to a bunch of weird XML attacks (see e.g. https://github.com/18F/calc/pull/833).


toolness:
Err also note that, like the current `DocCursor` serializer, this parser doesn't actually parse any of the top-level metadata about the policy itself. I guess I can file an issue about that after this PR is merged so it's recorded like #839 is.

api/document/parsers.py
```diff
@@ -0,0 +1,61 @@
+from typing import List
+
+from lxml import etree
+from rest_framework.exceptions import ParseError
+from rest_framework.parsers import BaseParser
+
+from document.tree import PrimitiveDict
+
+
+def text(text: str) -> PrimitiveDict:
```
cmc333333: Vanishingly small detail, but we use `text` to refer to strings in most of the code; should we make this `text_content` or something similar to distinguish?

api/document/tests/parsers_test.py
```diff
@@ -0,0 +1,106 @@
...
+    assert convert_xml_node('<blah title="foo"/>')['title'] == 'foo'
+
+
+def test_content_attribs_are_converted():
+    assert 'title' not in convert_content_node('<b/>')[0]
+    assert convert_content_node('<b title="foo"/>')[0]['title'] == 'foo'
+
+
+def test_xml_attribs_do_not_overwrite_node_data_keys():
+    assert convert_xml_node('<a node_type="b"/>')['node_type'] == 'a'
```
cmc333333: Excellent thought and check

api/document/tests/parsers_test.py
```diff
@@ -0,0 +1,106 @@
...
+    ]
+
+
+def test_error_raised_on_multiple_content_elements():
+    with pytest.raises(ParseError,
+                       match="<a> contains multiple <content> elements"):
+        convert_xml_node('<a><content/><content/></a>')
+
+
+def test_children_are_converted():
```
cmc333333: Two corner cases that'd be good to verify:
* If _multiple_ children are present, they're created
* If `content` and/or `num` are present, they are _not_ added as `children`

toolness:
Woot!  Good suggestions, I will address those and then merge.

I am actually a bit confused about defusedxml right now; it was a no-brainer for CALC, which used `xlrd` to read Excel spreadsheets and had a lot of vulnerabilities, but it *seems* like lxml is largely resilient to them--it's definitely resilient to billion laughs, which I just added a test for to ensure that we're protected against.

The [lxml FAQ](http://lxml.de/FAQ.html#how-do-i-use-lxml-safely-as-a-web-service-endpoint) mentions `defusedxml` only as a library that provides "an example setup and a wrapper API for lxml that applies certain counter measures internally".

In any case, I think we can address the defusedxml issue in a future PR, since this API isn't even accessible to the outside world yet.

toolness:
Oh, I should note that any 400 responses returned from XML PUTs might not be super easy to understand, since they'll generally be coming from the serializer, which is thinking in terms of JSON PUT requests rather than XML ones.  So for instance, errors might mention something about `node_type` or `content_type`, which is basically a tag name in XML, rather than an explicit field. This could be puzzling if we surface the errors back to users of the XML editor, but I figure we can polish this later if/when we find that folks are confused by it.

PR #870, opened 2018-01-17
Add "unimplemented" to editor, load data from API

cmc333333:
This pulls a lot from @toolness's #798 to add an "unimplemented" set of schema elements to the editor. That was enough to get us started loading document data from the API, and it seemed like a good time to implement authentication on the page load.

Unlike #798, the unimplemented content type is a _mark_ here rather than a full-blown schema node. I think that keeps everything mapped the same.

This also moves the editor from /document/editor to /admin/document-editor/[policy id].


api/tsconfig.json
```diff
@@ -5,5 +5,12 @@
...
     "alwaysStrict": true,
     "sourceMap": true
-  }
+  },
+  "lib": [
+    "DOM",
+    "ES6",
+    "DOM.Iterable",
+    "ScriptHost",
+    "ES5.Promise"
```
toolness: Oh, this was a thing I sort of glossed over when I made the original PR for this editor--I had assumed that staff/admins would be running a pretty recent version of a modern browser, so I didn't include much/anything in the way of polyfills or even transpilation, IIRC. Should we file an issue to address that later?

toolness:
Oh, it seems tslint doesn't like something:

```
ERROR: document/js/schema.ts[21, 30]: Unnecessarily quoted property 'class' found.
ERROR: document/js/schema.ts[29, 31]: Unnecessarily quoted property 'class' found.
```

api/tsconfig.json
cmc333333: Good point! Creates #872

cmc333333:
These "unimplemented" sections can be accidentally deleted (see #871); let me know if you think that's essential to getting this in.

cmc333333:
Thanks @toolness -- I'll rebase and merge

PR #873, opened 2018-01-17
Support PUT requests w/ content type akn+xml

toolness:
This builds upon #868 by adding the Akoma Ntoso parser to our `TreeView` so clients can PUT Akoma Ntoso documents via the REST API.

Note that this PR does _not_ currently add `defusedxml` because I'm confused about whether it's actually needed, as explained in https://github.com/18F/omb-eregs/pull/868#issuecomment-358361139.  @cmc333333 let me know if you think I should add it, otherwise I'll leave it off!

### DIFF
--- a/api/document/serializers/content.py
+++ b/api/document/serializers/content.py
@@ -92,7 +92,7 @@ class NestableAnnotationField(serializers.Field):
 
 class InlinesField(NestableAnnotationField):
     def __init__(self, is_leaf_node: bool) -> None:
-        super().__init__()
+        super().__init__(default=lambda: [])
         self.is_leaf_node = is_leaf_node
 
     def to_representation(self,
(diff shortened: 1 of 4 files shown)


api/document/serializers/content.py
```diff
@@ -92,7 +92,7 @@ def get_attribute(self,
 
 class InlinesField(NestableAnnotationField):
     def __init__(self, is_leaf_node: bool) -> None:
-        super().__init__()
+        super().__init__(default=lambda: [])
```
toolness: I also went ahead and made the serializer default `inlines` to an empty list, so that it didn't expect it in silly cases (such as when `__text__` content, which doesn't even support inlines, is submitted).

toolness:
Hooray!  I will file a separate issue about lxml security and merge this.

PR #874, opened 2018-01-17
Add "para" and "sec" node types to editor

### LINKED ISSUE #871: Don't allow "unimplemented" nodes be modified/deleted
When users go to "edit" unimplemented nodes in the editor, they replace the node with a new paragraph. We shouldn't allow those nodes to be modified (though moving them around is fine).

cmc333333:
This
* Resolves #871 by avoiding text nodes within `unimplemented_node`
* Swaps our conversion step to using ProseMirror's node factories rather than raw objects
* Sets up the "para" and "sec" types. The former breaks from #798 a bit as it moves all of the text content into a single ProseMirror type ("inline") and only includes that at most once.

This doesn't handle the API's `content` (ProseMirror's `mark`) types; that'll be the next step.

### DIFF
--- a/api/document/js/schema.ts
+++ b/api/document/js/schema.ts
@@ -3,22 +3,23 @@ import { Schema } from 'prosemirror-model';
 const schema = new Schema({
   nodes: {
     doc: {
-      content: '(paragraph | unimplemented_node)+',
+      content: 'block+',
     },
-    paragraph: {
-      content: 'inline*',
+    inline: {
+      content: 'text+',
       toDOM: () => ['p', 0],
     },
-    text: {
-      group: 'inline',
-    },
+    text: {},
     unimplemented_node: {
-      content: 'inline*',
+      group: 'block',
       atom: true,
       attrs: {
         data: {}, // will hold unrendered content
       },
-      toDOM: () => ['div', { class: 'unimplemented' }, 0],
+      toDOM(node) {
+        const nodeType = node.attrs.data.node_type || '[no-node-type]';
+        return ['div', { class: 'unimplemented' }, nodeType];
+      },
     },
   },
   marks: {
--- a/api/document/js/schema.ts
+++ b/api/document/js/schema.ts
@@ -9,6 +9,11 @@ const schema = new Schema({
       content: 'text+',
       toDOM: () => ['p', 0],
     },
+    sec: {
+      content: 'block+',
+      group: 'block',
+      toDOM: () => ['section', 0],
+    },
     text: {},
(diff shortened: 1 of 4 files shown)


api/document/js/schema.ts
```diff
@@ -3,22 +3,33 @@ import { Schema } from 'prosemirror-model';
...
     },
-    paragraph: {
-      content: 'inline*',
+    inline: {
+      content: 'text+',
       toDOM: () => ['p', 0],
     },
-    text: {
-      group: 'inline',
+    para: {
```
toolness: One funky thing I've noticed about ProseMirror's ecosystem is that sometimes optional commands "test" for the existence of certain kinds of schema elements by querying whether a certain name exists on the current schema.  It seems like a sort of "duck typing", but for the existence of schema features.  For example, even though `prosemirror-example-setup` is just an example, it's got a function called [`buildMenuItems`](https://github.com/ProseMirror/prosemirror-example-setup/blob/master/src/menu.js#L163-L216) that tests to see if certain names exist on the schema, and uses that to infer whether it should provide menu items that manipulate those types of content.

Because of this, in my original prototype I had decided to keep the name for paragraphs as `paragraph` (what the [basic schema](https://github.com/ProseMirror/prosemirror-schema-basic/blob/master/src/schema-basic.js#L11-L18) calls paragraphs) rather than `para` (what our own  schema calls paragraphs).  My assumption was that some of ProseMirror's special default functionality for e.g. splitting a paragraph into two if the user pressed <kbd>Enter</kbd> in the middle of one was based on paragraphs being called `paragraph` in the schema.

However, digging into ProseMirror's [`createParagraphNear`](https://github.com/ProseMirror/prosemirror-commands/blob/1e7f0e005b3d9097ede59dd9bc8d24c282e42521/src/commands.js#L247-L262)  command, it seems like it's calling some weird private method on `Node` called [`defaultContentType`](https://github.com/ProseMirror/prosemirror-model/blob/06af1aa1c814176c56f4c82cef7ee39c87fe27dd/src/node.js#L320-L322) to figure out what the type of a paragraph is, so that's weird, and now I'm just confused. 😕

cmc333333: Ah, good point, and pressing enter inside the text isn't creating a new paragraph for me. Unfortunately, switching it back to `paragraph` doesn't seem to resolve. I suspect this is related to the rather different `content` configuration; will follow up with more investigation.

cmc333333:
Thanks @toolness! I'm going to merge now but look into the enter-to-break-paragraph issue.

PR #880, opened 2018-01-18
Disable admin nav menu

cmc333333:
We didn't finish the styling for this menu, so this disables it. The changeset does make activation/deactivation as simple as flipping a variable, though.

## $nav-menu-active: true;
screen shot 2018-01-18 at 10 12 50 am

## $nav-menu-active: false;
screen shot 2018-01-18 at 10 12 16 am


cmc333333:
Thanks @toolness! I'll rebase and merge

PR #881, opened 2018-01-18
Ensure logins redirect to the admin screens.

### LINKED ISSUE #877: Visiting document editor URL without login redirects to nonexistent URL
I just noticed that visiting `/admin/document-editor/:id` on my install redirects to `/accounts/login/` instead of `/admin/login/`, which results in a 404. Not sure if this is something messed up with my system, but visiting `/admin` _does_ redirect me to the right place.

cmc333333:
By default, Django uses `/accounts/` urls for login redirects. We're only
using logins with the admin, however, so it makes sense to redirect to the
admin-specific sections.

This resolves #877.


cmc333333:
Thanks for catching it, @toolness !

PR #883, opened 2018-01-18
Share paragraph, heading styles between ui and editor

cmc333333:
This adds import/schema support for "heading"s and puts the paragraph and heading styles in the shared-styles dir. These are largely spacing settings -- we'll want to change all of the typography at once rather than by element.

@toolness this maintains the h2, h3, etc. but I really liked the use-h1-everywhere idea. Do you know if support is good enough that we could do that on the public-facing `ui`? It'd make the logic much simpler, but I want to be consistent.

Looks like
screen shot 2018-01-18 at 11 42 20 am


api/document/js/fetch-doc.ts
```diff
@@ -9,6 +9,15 @@ export function convertNode(node) {
...
 
 const NODE_TYPE_CONVERTERS = {
+  heading(node) {
+    // Duplicates logic in `ui`
+    const depth = node.identifier
+      .split('_')
+      .filter(e => e === 'sec')
+      .length + 1;
+    const text = (node.text || '').replace(/\s+/g, ' ');
+    return schema.nodes.heading.create({ depth }, schema.text(text));
```
toolness: Do you think we should deal with the potential for the depth to be greater than 6?  We could, for instance, use [`aria-level`](https://www.w3.org/WAI/GL/wiki/Using_role%3D%22heading%22_to_identify_headings#Example_2) in that case... ah well, probably something you've already thought about w/ the viewer, and probably pretty unlikely anyways.

cmc333333: That's a good question; we'd need to check in with design, too, as I'm sure they'd have an opinion. I _really_ like the use-h1s-for-everything option as a no-thinking solution; do you think that's supported enough for us to use in the public-facing ui?

cmc333333:
Thanks @toolness !

api/document/js/fetch-doc.ts
toolness: Good question, I just addressed this in #896!

Issue #867, opened 2018-01-17
Akoma Ntoso renderer may not render href of external_link


toolness:
I'm not very familiar with Akoma Ntoso, but this is how it renders external links:

```xml
<content>OMB Memorandum M-15-14, “<external_link>Management and Oversight of Federal Information Technology</external_link>,” June 10, 2015.</content>
```

As far as I can tell, there is no `href` attribute (or semantic equivalent) that indicates where the highlighted text should actually go.

cmc333333:
[Looks like](https://github.com/18F/omb-eregs/blob/b3076801d2c7fc4ce3440c27af667781ddcfcea4/api/document/renderers.py#L39) I forgot to add these attributes. I bet we can start by adding all string attrs without deeper inspection; footnote citations will be a bit more challenging.

PR #884, opened 2018-01-18
Add Akoma Ntoso editor

toolness:
This adds an Akoma Ntoso XML editor endpoint at `/admin/document-editor/:id/akn` via a simple [CodeMirror](https://codemirror.net/) embed.

> screen shot 2018-01-18 at 12 44 16 pm

The editor visual theme is easily changeable, either to one of [CodeMirror's built-in themes](https://codemirror.net/demo/theme.html#eclipse) or our own.

Right now the document can be saved via <kbd>Ctrl</kbd>+<kbd>S</kbd>, _but_ the actual PUT endpoint isn't accepting M-16-19 properly yet, so I need to fix up those issues first.


toolness:
@cmc333333 so I am a bit torn w/ this PR because I'd really like to get it in by 2pm so we can demo it tomorrow, but I also realize that the TypeScript file added here copy-pastes some code from the ProseMirror editor, with some simple modifications, so it's not exactly the best code.  Also, obviously the UI sucks.  So I'm not sure if this is something that should stay on a branch but that we demo tomorrow anyways, or that we merge soon-ish but file issues to fix up ASAP in the next sprint, or what. Any suggestions?

toolness:
Ok, this is working for the time being, but I continue to be _super_ confused about the difference between `marker` and `type_emblem`.  I noticed that `DocCursor` auto-set `type_emblem`, so in #868 I made it only set `marker`, but then saving M-16-19 failed because the footnote-finder was searching for footnotes with a `type_emblem` matching the footnote number, _not_ the `marker`... Aagghhh.

toolness:
For some reason saving the document also takes a _really_ long time, around 20 seconds. I'm not sure why this is, will need to look into it.

toolness:
Weird, it seems the actual time spent deserializing/validating/re-serializing the PUT request isn't much, somehow the response is just taking a really long time to get back to my browser.  It could be something wonky with my Docker setup, or it could be that my computer's JAMF'd... Odd.

toolness:
That's odd, earlier today the PUT worked fine, but when I just tried it recently, it justifiably rejected the document because our renderer wasn't adding `href` to our external links.  I fixed that (#867) in a very conservative way (via white-list) for now.

toolness:
> For some reason saving the document also takes a really long time, around 20 seconds. I'm not sure why this is, will need to look into it.

Uh, it looks like the vast majority of time is spent transmitting the document that was just PUT back to the client?  Changing `document.views.TreeView.put()` to return the following makes the PUT super fast:

```diff
-        return Response(serializer.data)
+        return Response('cool thanks bud')
```

cmc333333:
@toolness that's probably an n+1 query problem; if we don't explicitly [prefetch annotations](https://github.com/18F/omb-eregs/blob/0b19ea69b71a2dd69ad53e93a967b1e1ffa74570/api/document/views.py#L18), Django will _helpfully_ make a db query for us. Unfortunately, if we have a few hundred nodes, we'll have multiple-hundreds of queries.

toolness:
Oh funkadelic. Ok I think I just fixed that--_however_, it still feels like returning the whole JSON document on PUT is overkill if the client is just going to throw it away, right?  For example, the JSON serialization of M-16-19 is about 162K, so it feels weird to return that massive payload for no reason. Ah well, I guess I will just keep it in there for now.

cmc333333:
Agreed. For eRegs we just returned a 204 (Success, no content) for all of the PUT/POSTs.

toolness:
Oh heh.  Okay, I just fixed that part.

So the only part of this PR that I don't like is the one that works around the whole `type_emblem` vs. `marker` confusion by having the JSON importer fall back to `marker` if it can't find a footnote with the right `type_emblem`.  I'm not sure if that should be kept for this PR and fixed in a future one, or fixed in this one. What do you think?

toolness:
The one other thing I don't like, but that we can address in a separate PR, is that `ValueError`s raised by the JSON importer, which really correspond to validation errors in the PUT payload (e.g. having a `<footnote_citation>` for a nonexistent footnote), get converted into 500s rather than 400s with helpful error messages.  I think we could deal with that in a separate PR though.

cmc333333:
Regarding footnote citations, let me sketch out my understanding of the current process.
* After we've marshaled from JSON/XML into primitives, when we hit a citation, we search the subtree below our position for any footnote node that could match.
* The citation models themselves only hold an fk to the referenced node.
* When we serialize from database models, we're converting that reference into a whole, nested footnote node.

Maybe we're too imprecise on the import side and asking for something that's hard to encode in a validator. Does it make sense to require something closer to a foreign key? "Give the footnote an explicit identifier and reference that in your citation" I'd like to see how this sort of thing is handled in other systems.

api/document/js/akn/main.ts
```diff
@@ -0,0 +1,92 @@
...
+  const pathParts = (path || window.location.href).split('/');
+  const policyId = pathParts[pathParts.length - 2];
+  return axios({
+    method: 'put',
+    url: `/document/${policyId}`,
+    data,
+    headers: {
+      'Content-Type': 'application/akn+xml',
+      'X-CSRFToken': getEl('[name=csrfmiddlewaretoken]')
+        .getAttribute('value'),
```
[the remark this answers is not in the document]
toolness: Oh, I guess I had assumed that we used [`HttpOnly` cookies for security](https://www.owasp.org/index.php/HttpOnly).  If we don't though, I could do that.

api/document/renderers.py
```diff
@@ -5,6 +5,10 @@
 
 from document.serializers.content import PlainTextSerializer
 
+# Properties of content nodes that should be converted to XML
+# attributes.
+CONTENT_ATTRIBS = ['href']
```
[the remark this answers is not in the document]
toolness: Yeah, my original implementation (which I didn't commit) actually skipped over `content_type` and `text` but additionally tested to see if the value of the key was a string, as it originally choked on `footnote_node` (a dict).  But I felt like that was prone to weird behavior, so I decided to make the converter thingy a bit more explicit... humm.

api/document/js/akn/main.ts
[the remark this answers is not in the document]
cmc333333: Good point, and we are! Ignore me.

toolness:
I decided to pass the entire document API URL to the editor front-end via a `data-document-url` attribute. This way, our front-end code won't need to duplicate our URL mapping logic, and thanks to our use of `django.urls.reverse`, we're also guaranteed (or at least more likely) to be pinging a real API endpoint.

api/document/js/akn/main.ts
```diff
@@ -0,0 +1,80 @@
...
+import { getEl, getElAttr } from '../util';
+
+
+// We need to load our CSS via require() rather than import;
+// using the latter raises errors about not being able to find
+// the module.
+
+require('codemirror/lib/codemirror.css');
+require('codemirror/theme/eclipse.css');
+require('codemirror/addon/dialog/dialog.css');
```
cmc333333: For a follow up PR: let's add these to our compiled style sheets; better to be consistent than use CSS-in-JS some of the time and straight SASS compilation others.

api/document/js/main.ts
```diff
@@ -13,23 +14,15 @@ import fetchDoc from './fetch-doc';
 // do with non-standard kinds of imports like CSS, so using require()
 // likely bypasses TypeScript and goes straight to webpack, which
 // deals with it correctly. I could be wrong, though. -AV
-declare function require(path: string): null;
```
cmc333333: If we don't need to declare this anymore, how much of the comment above it should go, too?

api/document/templates/document/editor_akn.html
```diff
@@ -0,0 +1,67 @@
...
+      padding:.1em .6em;
+      text-shadow:0 1px 0 #fff;
+    }
+    </style>
+    <title>Akoma Ntoso Document editor</title>
+  </head>
+  <body>
+    <h1>Akoma Ntoso Document editor</h1>
+    <pre id="status"></pre>
+    <div id="editor" data-document-url="{{ document_url }}"></div>
```
cmc333333: I like passing in the whole url much more than just the doc number. Do we need to do anything around escaping here? It might be a good idea to use the `{% url %}` directive to handle that for us.

Not for now, but I wonder if there's a meta-data or other RDFa tag that'd make this semantic.

api/document/js/__tests__/schema.test.ts
```diff
@@ -23,7 +23,7 @@ describe('unimplemented_node', () => {
 });
 
 describe('heading', () => {
-  [2, 3, 4, 5].forEach(depth => {
+  [2, 3, 4, 5].forEach((depth) => {
```
toolness: This lint error actually comes from a recently-merged PR, but it wasn't checked in that PR because of a problem w/ the `lint` script in the `package.json`, which caused the linter to skip this file. I've fixed that problem in this PR though, which made the linter catch this.

api/document/js/main.ts
toolness: Oh, I figured the comment above still applied, because the fact that we're still using `require()` over `import` is still potentially confusing IMO... But as you mentioned above, since we're going to switch to just putting the CSS in our funky SCSS build pipeline thing, we can get rid of it very soon regardless!

toolness: Uhh wait, now CircleCI is complaining about `require` being undefined.  I have no idea why.  Guess I'll re-introduce it for now...

api/document/templates/document/editor_akn.html
toolness: Hmm, I don't think that we need to do any escaping there--I had decided to just pass it in from the view function so that the view functions for our prosemirror + codemirror editors could reuse the same logic for coming up w/ that URL, rather than dupicating it in the templates, but I guess either way works!

toolness:
Cool, I will merge this and file a bunch of follow-up issues!

PR #885, opened 2018-01-18
Add initial "mark" support to editor import.

cmc333333:
This handles nested content by keeping a stack of derived markers and
linearizing them into an array of `text`s (each with the stack of marks).

Similar to the unimplemented_node, we want to track all of the meta
information around an unknown content/annotation/inline/mark type (yay
different names for similar concepts!).

Looks like
screen shot 2018-01-18 at 2 00 34 pm

### DIFF
--- a/api/document/js/fetch-doc.ts
+++ b/api/document/js/fetch-doc.ts
@@ -1,4 +1,5 @@
 import axios from 'axios';
+import { flatten } from 'lodash/array';
 import { Mark, Node } from 'prosemirror-model';
 
 import schema from './schema';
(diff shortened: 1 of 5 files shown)


api/document/js/fetch-doc.ts
```diff
@@ -8,10 +9,25 @@ export function convertNode(node) {
...
+    const text = (content.text || '').replace(/\s+/g, ' ');
+    return [schema.text(text, marks)];
+  }
+  const contentType = content.content_type in CONTENT_TYPE_CONVERTERS ?
+    content.content_type : 'unimplemented_mark';
+  const mark = CONTENT_TYPE_CONVERTERS[contentType](content);
+  const updatedMarks = marks.concat([mark]);
+  const children = (content.inlines || []).map(child =>
+    convertContent(child, updatedMarks));
+  return [].concat(...children);
```
toolness: I wonder if adding a type annotation to `children` here, purely for the sake of documentation, would be useful?  Or maybe rename it to `childNodeLists`?

Also, I find `Array.prototype.concat`'s behavior really confusing because of how it does a different thing depending on whether its arguments are arrays or not.  I wonder if it'd help to assign this value to a well-named variable, like say, `flattenedChildren`, or maybe factor it out into a well-named utility function like `shallowFlatten` or something... eh, whatever, I'm probably just dense.

cmc333333: Can do, good thoughts, thanks @toolness

cmc333333: I'll throw in lodash; it's well known and likely to come in handy later.

cmc333333:
Updated! I'll rebase and merge now.

PR #887, opened 2018-01-18
Update README.md

yowill:
Adding more detail for the requested reporting for Joanne, CPS, SMT teams that Product teams are supposed to have published


README.md
```diff
@@ -16,11 +16,41 @@ YouTube](https://www.youtube.com/playlist?list=PLu2xTIVb2mmQb-QURyZQEoqK4c2XrW43
...
+
+### Risks & Mitigation Strategies
+The primary risks we are facing is updating antiquated documentation processes and tools, and trying to improve them without making the conversion too onerous on agencies. By intensely focusing on user-centered design, we are working with agency stakeholders and users to inform us on how to most effectively relieve the burden they face, while also not adding to their workload.
+
+### Metrics
+Besides financial and time savings by offering a more effective and organized user experience, these are some of our draft current goals for the MVP launch of the product:
+
+**Goal:** Users are less confused about what is required of them by policy. 
+**Metric:** EHelp ‘tickets’ (Calls, emails, etc.) to OMB desk officers go down in volume/length of time.
```
cmc333333: A lot of these metrics have an extra "E" in front; I'm guessing copy-paste conversion issue?

README.md
```diff
@@ -16,11 +16,41 @@ YouTube](https://www.youtube.com/playlist?list=PLu2xTIVb2mmQb-QURyZQEoqK4c2XrW43
...
+Besides financial and time savings by offering a more effective and organized user experience, these are some of our draft current goals for the MVP launch of the product:
+
+**Goal:** Users are less confused about what is required of them by policy. 
+**Metric:** EHelp ‘tickets’ (Calls, emails, etc.) to OMB desk officers go down in volume/length of time. 
+
+**Goal:** Users can quickly/easily find the parts of policy that they need. 
+**Metric:** ETime spend searching for policy reduces.
+
+**Goal:** Convert the policy library to pure digital data:
+**Metric:** 100% of policies will be in real text as opposed to PDF, by the time we’re done.
```
cmc333333: Should these "Metric"s be describing a number to look at or where we want that metric to be? If the former, we may want to wordsmith these a little differently. For example, this line could be
**Metric** Portion of policies available in a text format (as opposed to PDF)

yowill: Ah that's correct.

README.md
yowill: Yessir

PR #889, opened 2018-01-18
Add migration to go through policies setting workflow_phase to 'no_doc'.

tadhg-ohiggins:
Add correct defaults
The winds grow tricksy, and now
It rains in the west

Adds a Django migration to set all the `Policy` objects that don't have values for `workflow_phase` to have a `workflow_phase` value of `no_doc`.


tadhg-ohiggins:
Thanks @cmc333333 !

PR #890, opened 2018-01-19
Use the "System Font Stack" for all sans-serif.

### LINKED ISSUE #708: Discuss/implement font updates
@rtwell shared an [article](https://booking.design/implementing-system-fonts-on-booking-com-a-lesson-learned-bdc984df627f) about using OS fonts rather than webfonts or system fonts. Get on the same page around strategy and then implement that change.

cmc333333:
Each OS has a custom, roughly equivalent default font. Unfortunately, they're
not used with the "sans-serif" font family. This uses the "System Font Stack"
to grab the most popular, per @rtwell 's recommendation.

See https://css-tricks.com/snippets/css/system-font-stack/ for details.

This also removes Merriweather and SourceSansPro, which we carried around from
USWDS but don't need any longer.

Resolves #708 

### Before
screen shot 2018-01-18 at 7 01 00 pm

### After
screen shot 2018-01-18 at 7 01 34 pm

### Before
screen shot 2018-01-18 at 7 09 22 pm

### After
screen shot 2018-01-18 at 7 09 50 pm


rtwell:
😍

cmc333333:
Thanks @rememberlenny !

PR #895, opened 2018-01-19
Add a few "problem" policies.

cmc333333:
We have stricter memory limits in our cloud.gov environment than on our
development machines. This adds exceptions for a handful of pdfs that could be
processed locally, but can't in cloud.gov.


cmc333333:
Thanks @toolness . I'll rebase and merge.

Issue #924, opened 2018-01-22
Consider adding end-to-end tests for XML and JSON serialization/deserialization


toolness:
It might be useful to add tests that take some of our policies (at least M-16-19) and verify the idempotency of serializing them, deserializing them, and re-serializing them. (I am not sure if I actually used the word "idempotency" properly there, but hopefully you know what I mean.)

Obviously this isn't a very granular test, but it could help surface integration issues not identified by our individual unit tests, such as ones I encountered while working on #884.

PR #926, opened 2018-01-23
Add an end-to-end test for AKN

toolness:
This attempts to help with #924 by adding an end-to-end test for AKN using a real-world policy (M-16-19).

It is kind of convoluted, and I'm not sure how useful it is. It did, however, help me realize that the preamble of our XML example documents is actually part of the document that's stored in the database, which I hadn't realized--I thought the preamble was solely stored as metadata on the `Policy` model, but it actually seems to be stored in _both_ the `Policy` _and_ under a `preamble` node type in the document. But because some/all of our DRF pipeline ignores it, it gets lost at some point.

Because of this discrepancy, and a few others that I didn't investigate terribly closely, I ended up doing more round-trips between the AKN parser/renderer than I had originally intended. I'm not really sure how useful the approach is, and it's also possible that this test might be a lot simpler (and easier to parametrize to JSON instead of AKN) if it just operated via the REST API instead of using our renderer/parser/serializer subclasses directly.

Anyhow, @cmc333333 can you take a look at this and let me know what you think?

### DIFF
--- a/api/document/tests/end_to_end_test.py
+++ b/api/document/tests/end_to_end_test.py
@@ -17,6 +17,7 @@ MY_DIR = Path(__file__).parent.resolve()
 EXAMPLE_DOCS_DIR = MY_DIR / '..' / '..' / 'example_docs'
 
 
+@pytest.mark.xfail(raises=AssertionError, reason="Fix this!!")
 @pytest.mark.django_db
 def test_akn_works():
     # Phase 1: Import the document from XML, serialize it, and
@@ -40,16 +41,6 @@ def test_akn_works():
     data = DocCursorSerializer(cursor).data
     akn = AkomaNtosoRenderer().render(data)
 
-    # Phase 4: Parse the AKN, deserialize it, and save it.
-    parsed_akn_data = AkomaNtosoParser().parse(BytesIO(akn))
-    serializer = DocCursorSerializer(cursor, data=parsed_akn_data)
-    serializer.is_valid(raise_exception=True)
-    cursor = serializer.save()
-
-    # Phase 5: Re-serialize the document and render it to AKN.
-    data2 = DocCursorSerializer(cursor).data
-    akn2 = AkomaNtosoRenderer().render(data2)
-
-    # Now ensure the AKN from phase 3 matches the AKN from
-    # phase 5.
-    assert akn.decode('utf-8') == akn2.decode('utf-8')
+    # Now ensure the AKN from phase 1 matches the AKN from
+    # phase 3.
+    assert original_akn.decode('utf-8') == akn.decode('utf-8')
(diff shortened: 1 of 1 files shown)


api/document/tests/end_to_end_test.py
```diff
@@ -0,0 +1,55 @@
...
+
+@pytest.mark.django_db
+def test_akn_works():
+    # Phase 1: Import the document from XML, serialize it, and
+    # render it to AKN.
+    policy = mommy.make(Policy, omb_policy_id='M-16-19')
+    call_command('import_xml_doc', str(EXAMPLE_DOCS_DIR / 'm_16_19_1.xml'),
+                 'M-16-19')
+    docnode = DocNode.objects.filter(policy=policy, depth=1).first()
+    cursor = DocCursor.load_from_model(docnode)
```
cmc333333: Should we use that prefetched version of the queryset to speed this up?

api/document/tests/end_to_end_test.py
```diff
@@ -0,0 +1,55 @@
...
+
+    # Phase 4: Parse the AKN, deserialize it, and save it.
+    parsed_akn_data = AkomaNtosoParser().parse(BytesIO(akn))
+    serializer = DocCursorSerializer(cursor, data=parsed_akn_data)
+    serializer.is_valid(raise_exception=True)
+    cursor = serializer.save()
+
+    # Phase 5: Re-serialize the document and render it to AKN.
+    data2 = DocCursorSerializer(cursor).data
+    akn2 = AkomaNtosoRenderer().render(data2)
```
cmc333333: I'm probably missing something -- what's Phase 4 and Phase 5 getting us above Phase 2 and 3? I'd think we could compare `akn` to `original_akn` and cover the same ground?

api/document/tests/end_to_end_test.py
toolness: Yeah! At the time I wrote this, that PR wasn't merged yet, but now that it is, I will use it.

api/document/tests/end_to_end_test.py
toolness: So that's what I was originally doing, but there were weird differences, like the `<preamble>` existing in `original_akn` but not `akn`... I think there were other differences but I didn't look very closely... um, these could be bugs in our system that we actually want to fix, though, so the better solution might be to compare `akn` to `original_akn` but mark the test as `xfail` until we fix those things...

cmc333333: Hah, yeah, I'm pretty sure that means our system's broken ;) `xfail` seems like the way to go.

toolness: Ok, I will remove the extra phases and `xfail` and merge!

PR #927, opened 2018-01-23
Link to document by slug, look at workflow phase

### LINKED ISSUE #678: Link to policy detail if it has a slug but not an OMB policy id

cmc333333:
This resolves #678 by linking to the document page when a policy doesn't have an M-number.

It also checks the workflow phase field rather than the presence/absence of docnodes when determining what to link to. Note, however, this means we'll be removing all our current links. I figure we can use the admin to adjust where desired, but we can also do a data migration if there's a better option.

### DIFF
--- a/api/ombpdf/tests/test_management_commands_scrape_memoranda.py
+++ b/api/ombpdf/tests/test_management_commands_scrape_memoranda.py
@@ -4,7 +4,6 @@ from unittest.mock import Mock
 import pytest
 from model_mommy import mommy
 
-from document.models import DocNode
 from ombpdf.management.commands import scrape_memoranda
 from reqs.models import Policy, WorkflowPhases
 
@@ -94,9 +93,11 @@ def test_scrape_memoranda(monkeypatch):
         lambda p, _: not p.omb_policy_id.endswith('3')
 
     for i in range(1, 5):
-        policy = mommy.make(Policy, omb_policy_id=f'M-0{i}-0{i}')
-        if i == 1:     # Immitate a policy having "already" been processed
-            mommy.make(DocNode, policy=policy)
+        workflow_phase = WorkflowPhases.no_doc.name
+        if i == 1:  # Immitate a policy having "already" been processed
+            workflow_phase = WorkflowPhases.published.name
+        mommy.make(Policy, omb_policy_id=f'M-0{i}-0{i}',
+                   workflow_phase=workflow_phase)
 
     successes, failures = scrape_memoranda.scrape_memoranda()
     # M-01-01 has a docnode, so is ignored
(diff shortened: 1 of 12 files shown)


api/ombpdf/tests/test_management_commands_scrape_memoranda.py
```diff
@@ -94,9 +93,11 @@ def test_scrape_memoranda(monkeypatch):
         lambda p, _: not p.omb_policy_id.endswith('3')
 
     for i in range(1, 5):
-        policy = mommy.make(Policy, omb_policy_id=f'M-0{i}-0{i}')
-        if i == 1:     # Immitate a policy having "already" been processed
-            mommy.make(DocNode, policy=policy)
+        workflow_phase = WorkflowPhases.no_doc.name
+        if i == 1:  # Immitate a policy having "already" been processed
```
toolness: Not a big deal at all, and I see it's also not new code, but `Immitate` is spelled `Imitate`.

api/reqs/models.py
```diff
@@ -107,23 +105,11 @@ def __str__(self):
         return self.name
 
 
-class PolicyQueryset(models.QuerySet):
```
toolness: Farewell,  humble `QuerySet` subclass, you served us well... 😢

api/reqs/tests/model_tests.py
```diff
@@ -33,6 +33,19 @@ def test_original_url():
...
+    (models.WorkflowPhases.edit, False),
+    (models.WorkflowPhases.cleanup, False),
+    (models.WorkflowPhases.failed, False),
+    (models.WorkflowPhases.no_doc, False),
+    (models.WorkflowPhases.published, True),
+    (models.WorkflowPhases.review, False),
+))
+def test_has_document(phase, has_document):
+    policy = models.Policy(workflow_phase=phase.name)
+    assert policy.has_document is has_document
```
toolness: Hmm, so I am actually a bit confused by the name `has_document`. It definitely makes sense in the context of the viewer UI, but it doesn't necessarily make sense to me in the context of the whole workflow; from a staff member's perspective, for instance, I'd expect `has_document` to be `True` for all `WorkflowPhases` except `no_doc` and `failed`... Should this property be called `has_published_document` instead, maybe? Or am I misinterpreting things?

cmc333333: You've got it right; I _suspect_ we'll be getting rid of this (and the "public" field) relatively soon, but I'm happy to update.

cmc333333:
Thanks @toolness! Updated, rebased, and will merge.

Issue #923, opened 2018-01-22
Remove document.xml_importer


toolness:
With #868 and #884 landing, I *think* we can safely remove `document.xml_importer`, if we want, by:

1. loading the AKN representation for our two current XML docs in our AKN editor,
2. replacing their respective XML files in the repo with the content from step 1,
3. modifying our `import_xml_doc` command to use our AKN XML parser to bring the document into the database, rather than `document.xml_importer`.

Issue #903, opened 2018-01-22
XML parser doesn't properly populate type_emblem


toolness:
When I implemented our AKN XML parser in #868, I misinterpreted the [difference between `marker` and `type_emblem`](https://github.com/18F/omb-eregs/issues/891), which led to an awkward hack in 18b3160 whereby our JSON importer falls back to looking at the `marker` field of a footnote if it can't find what it's looking for in the `type_emblem`.  We should get this straightened out so that the XML parser is parsing correctly, and the JSON importer is only looking at `type_emblem`. And, as #891 mentions, we should document the difference so nobody makes this kind of mistake again.

toolness:
I _think_ this has mostly been fixed by #929; all that's remaining is to revert the awkward hack in 18b3160d5e764b0e2520b410bae99eae2f3c5e9f and make sure everything still works.

PR #929, opened 2018-01-23
Fix end to end AKN test

### LINKED ISSUE #928: Fix end-to-end AKN test
I added an end-to-end AKN test in #926 but had to `xfail` it because it revealed that our underlying pipeline is a bit broken. 😳   So we should fix it and un-`xfail` this test!

(Also, the fix should [prefetch annotations](https://github.com/18F/omb-eregs/pull/926#discussion_r163264772) to speed up the test.)

toolness:
This fixes #928.

So part of the reason the test I added in #926 was failing was because I noticed the `DocNode.depth` field was `PositiveIntegerField` and assumed the root node would have a `depth` of `1`, since that is the lowest positive integer AFAIK.  However, after merging that PR and investigting the failure more closely, I learned that `PositiveIntegerField` actually supports `0`, and that I should have been querying against that instead.

So that was one reason for the test failing. The other was that the legacy XML importer seems to give slightly different `id` attributes to the nodes it creates than our AKN deserializer, so those were differing in minor ways.

To fix this, for now I've just stripped out the `id` attributes from the rendered XML strings before comparing them. I figure we can remove this part once we take care of #923 and have `manage.py import_xml_doc` use our AKN parser instead of the legacy one, which _should_ ensure that the same IDs are always assigned.

However, there is another potential fix, which would involve changing the AKN parser to actually pass in the existing `id` attribute of a node to the deserializer, which could then use it when available.  This would probably require us to add additional validation logic to ensure that all the `id`s are unique, though. This might be better for the long-term, since I assume we _will_ want those `id`s to stay the same in some cases.

@cmc333333 do you have a preference for which fix I implement?  Or is there yet another potential fix I haven't considered yet?


api/document/tests/end_to_end_test.py
```diff
@@ -17,16 +18,33 @@
...
+    s = re.sub(r'id="[a-z0-9_]+', '', s)
+
+    return s
+
+
+def get_cursor_for_policy(policy: Policy) -> DocCursor:
+    docnode = DocNode.objects.filter(policy=policy, depth=0).\
+        prefetch_annotations().first()
+    cursor = DocCursor.load_from_model(docnode, subtree=False)
+    cursor.add_models(cursor.descendants().prefetch_annotations())
```
toolness: I'm not a big fan of duplicating this code from our `document.views.TreeView.get_object()` method. Do you think we should actually have `DocNode.load_from_model` call `prefetch_annotations` on its subtree queryset automatically? Are there cases where auto-prefetching those annotations wouldn't be helpful?

api/document/tests/end_to_end_test.py
```diff
@@ -17,16 +18,33 @@
...
-@pytest.mark.xfail(raises=AssertionError, reason="Fix this!!")
+def clean(content: bytes) -> str:
+    s = content.decode('utf-8')
+
+    # We're going to strip out `id` attributes here because our
+    # legacy XML importer seems to auto-assign ids in a different
+    # way than our serializer.
+    s = re.sub(r'id="[a-z0-9_]+', '', s)
+
+    return s
```
cmc333333: Mmm. I'd think that this indicates another bug; I suspect the deserializer isn't accounting for the type emblem.

api/document/tests/end_to_end_test.py
cmc333333: Good point; the prefetch was originally separate in case we wanted to dynamically select which annotations would be present in the result, but that's all been smashed together now. No harm making it even more direct.

api/document/models.py
```diff
@@ -25,6 +25,12 @@ def prefetch_annotations(self):
                              requirement_prefetch)
 
 
+# Django's `PositiveIntegerField` is named in a very misleading way,
+# because it actually allows values of 0. For developers who aren't
+# aware of this oddity, we'll alias it to something more accurate.
+NonNegativeIntegerField = models.PositiveIntegerField
```
cmc333333: Hah, sounds good.

api/document/tests/end_to_end_test.py
toolness: Oh blaaaarg indeed, I think this goes back to the ol' #903.

cmc333333:
Sorry, responded to the code rather than the comments. I don't think we should pay particular attention to the incoming ids, as users are likely to change their order. As we've been discussing in Slack, we _will_ want to account for data around emblems, however, as we'll want list item `b. ...` to be identified as `...._b`. How we determine which of these emblems is meaningful and which isn't will be tricky; I suspect we'll base it on what's modify-able in the editor.

toolness:
> How we determine which of these emblems is meaningful and which isn't will be tricky; I suspect we'll base it on what's modify-able in the editor.

Hmm, so it seems like the only way to properly make this test pass is by implementing that tricky behavior, right?

cmc333333:
I _think_ we can probably make it pass by including the type emblem in the serialized XML (if it's not already) and reading that in when we deserialize. We'll just want to think about what to do in the editor -- maybe we don't send the type emblem unless it's a list item or something? That can probably wait.

toolness:
Oh ok... I _think_ I just did that in e530bba0d507c6c1a53007c1e4638208811830bc, what do you think?

api/document/tests/end_to_end_test.py
toolness: Oh I had forgotten about the dynamically selection which annotations would be present in the result!  Hmm.

Meh, I will leave it as-is for now. Maybe I will do it in a separate PR.

toolness:
Woot! I also filed #932 to make sure we validate the emblem values too.

PR #934, opened 2018-01-24
Save data from prosemirror

cmc333333:
This moved and combined templates, stylesheets, and a shared Api module so that the ProseMirror editor can re-use a goodly chunk of the AKN logic. With that, it was pretty straight forward to add a save function.

This is a lot of changes in terms of line count, but much of it is composed of renaming and moving functions around.

### DIFF
--- a/api/document/js/Api.ts
+++ b/api/document/js/Api.ts
@@ -22,7 +22,7 @@ export default class Api {
   contentType: string;
   csrfToken: string;
 
-  constructor(url: string, contentType: string, csrfToken: string) {
+  constructor({ contentType, csrfToken, url }) {
     this.url = url;
     this.contentType = contentType;
     this.csrfToken = csrfToken;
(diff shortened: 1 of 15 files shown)


api/document/js/Api.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+  setStatus(errMsg, 'editor-status-error');
+}
+
+export default class Api {
+  url: string;
+  contentType: string;
+  csrfToken: string;
+
+  constructor(url: string, contentType: string, csrfToken: string) {
```
toolness: If it's possible to use destructuring notation with these arguments and have them passed-in as an object, that might make it easier to use and less easy to misuse, e.g. by confusing the ordering of `contentType` and `csrfToken`, especially as we move forward and potentially add more arguments to this constructor, but it's up to you!

cmc333333: Good idea; I'll send that up.

cmc333333:
Re React: I think that's a good thought to keep in the back of our minds. I suspect the status message bar will be completely replaced by the end of this sprint, though, so I'm not too worried about keeping it as is.

I'll do param destructurring for that constructor and try to clean up as many of the sass lint issues as  is feasible before merging.

PR #937, opened 2018-01-24
Validate type_emblem

### LINKED ISSUE #932: Validate "type_emblem" in PUT API requests
Since `type_emblem` is ultimately part of an element's (auto-generated) identifier, we should make sure that it:

* only contains alphanumeric-ish characters (figure out what the actual criteria is).
* doesn't cause an element's identifier to become non-unique. I _think_ this means that we should ensure that the `(node_type, type_emblem)` is unique among all immediate children of any given node.

toolness:
This fixes #932.

### DIFF
--- a/api/document/models.py
+++ b/api/document/models.py
@@ -1,6 +1,7 @@
 import itertools
 from typing import Iterator, List
 
+from django.core.validators import RegexValidator
 from django.db import models
 
 
@@ -46,12 +47,18 @@ class DocNode(models.Model):
     '''
 
     policy = models.ForeignKey('reqs.Policy', on_delete=models.CASCADE)
-    # e.g. part_447__subpart_A__sec_1__para_b
+    # e.g. part_447__subpart_A__sec_1__para_b.
     identifier = models.CharField(max_length=1024)
     # e.g. para
     node_type = models.CharField(max_length=64)
     # e.g. b
-    type_emblem = models.CharField(max_length=16)
+    type_emblem = models.CharField(
+        max_length=16,
+        validators=[RegexValidator(
+            r'[A-Za-z0-9]+',
+            message="Only alphanumeric characters are allowed."
+        )]
+    )
     text = models.TextField(blank=True)
     # e.g. "(a)", "From:", "1.", "•"
     marker = models.CharField(max_length=64, blank=True)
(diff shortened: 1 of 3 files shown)


api/document/models.py
```diff
@@ -46,12 +47,18 @@ class DocNode(models.Model):
...
     # e.g. para
     node_type = models.CharField(max_length=64)
     # e.g. b
-    type_emblem = models.CharField(max_length=16)
+    type_emblem = models.CharField(
+        max_length=16,
+        validators=[RegexValidator(
+            r'[A-Za-z0-9]+',
+            message="Only alphanumeric characters are allowed."
+        )]
```
toolness: It's interesting to note that this validator, when used on `DocNode`, doesn't actually do anything useful, because we only ever bulk create `DocNode`s and never validate their fields.  It's only when we reuse this validator in our `DocCursorSerializer` that the logic is used.  So we could potentially _only_ include this in `DocCursorSerializer`, but I thought it'd be useful to have `DocNode` as its primary home to help developers understand the core model better.

api/document/serializers/doc_cursor.py
```diff
@@ -22,12 +23,27 @@ def to_representation(self, instance: DocCursor) -> List[PrimitiveDict]:
             context={**self.context, 'is_root': False},
         ).data
 
+    def validate_type_emblem_uniqueness(self, data: List[PrimitiveDict]):
+        seen: Set[Tuple[str, str]] = set()
+        for child in data:
+            if 'type_emblem' not in child:
+                continue
```
cmc333333: We'd better check that the auto-incrementing logic will prevent a conflict, too. Otherwise, you could imagine input like
```
<parent>
    <child />
    <child emblem="1" />
</parent>
```
That makes me wonder if this check should be happening somewhere else? Maybe in the deserializer?

toolness: whoa.

just so i fully understand the problem, should the auto-incrementer work _around_ the potential conflict, or should an error be raised to the user?  I'd assume the former, since otherwise users (or at least editors of AKN XML) would need to understand the auto-incrementing algorithm, but this AKN spec has been full of other weird surprises, so who knows 😆

cmc333333: Oh, good point. We should probably trust the user's input here -- if they want to label the first child "2", let's let them.

toolness: Hehe ok. Do you think I should add that kind of work-around logic in this PR, or a separate one?

cmc333333: Separate PR or even kicking the can down the road in a separate issue sounds fine.

PR #946, opened 2018-01-24
Remove document.xml_importer

### LINKED ISSUE #923: Remove document.xml_importer

toolness:
This fixes #923 by:

1. loading the AKN representation for our two current XML docs in our AKN editor,
2. replacing their respective XML files in the repo with the content from step 1,
3. modifying our `import_xml_doc` command to use our AKN XML parser to bring the document into the database, rather than `document.xml_importer`.
4. Removing `document.xml_importer`.


api/document/management/commands/import_xml_doc.py
```diff
@@ -17,6 +18,14 @@ def fetch_policy(identifier: str):
         return Policy.objects.filter(omb_policy_id=identifier).first()
 
 
+def import_xml_doc(policy, xmlstream):
+    parsed_data = AkomaNtosoParser().parse(xmlstream)
+    cursor = DocCursor.new_tree('', policy=policy)
```
toolness: Aggh I _really_ dislike this line.  Our `DocCursorSerializer` was built to serialize existing `DocCursor`s, not to create brand-new ones. So here I'm just "creating" a new `DocCursor` for no other reason than so our importer can eventually pluck out its `policy`. Ooooof.

api/document/tests/import_xml_doc_test.py
```diff
@@ -15,3 +19,30 @@ def test_fetch_policy_pk():
 def test_fetch_policy_number():
     policy = mommy.make(Policy, omb_policy_id='M-12-13')
     assert import_xml_doc.fetch_policy('M-12-13') == policy
+
+
+@pytest.mark.django_db
+def test_import_xml_doc():
```
toolness: This is really just an old test from `xml_importer`, but I made it run twice just to make sure the code is OK with both creating a new `DocNode` tree where none existed before, _and_ with overwriting an existing `DocNode` tree if it already exists.

api/document/management/commands/import_xml_doc.py
cmc333333: Would it make sense to override the serializer's `create()` method? We'd need to pass the policy as context, but I wonder if that's where folks would expect this sort of logic?

api/document/management/commands/import_xml_doc.py
```diff
@@ -17,6 +18,14 @@ def fetch_policy(identifier: str):
...
+    serializer = DocCursorSerializer(cursor, data=parsed_data)
+    serializer.is_valid(raise_exception=True)
```
cmc333333: I'm betting this raises some gnarly, but data-rich exceptions. Not necessary for this round, but we should think about ways to surface them to users.

api/document/management/commands/import_xml_doc.py
toolness: Ah that's a great idea! I just filed #950 to address it.

Issue #823, opened 2018-01-09
Investigate/use modelcuster for annotation creation


cmc333333:
When inserting documents, we're very careful about insertion order and separating annotation from their referenced docnodes. Django doesn't allow models to have related objects until after insertion, which means we have to perform a lot of gymnastics around tracking related objects prior to insertion.

Wagtail's [modelcluster](https://github.com/wagtail/django-modelcluster) claims to have solved that problem. Investigate it and implement if it would clean up our code.

PR #949, opened 2018-01-24
Validate footnotes better

### LINKED ISSUE #900: Submitting PUT content with invalid footnote citations raises a 500
From https://github.com/18F/omb-eregs/pull/884#issuecomment-358967043:

> The one other thing I don't like, but that we can address in a separate PR, is that `ValueError`s raised by the JSON importer, which really correspond to validation errors in the PUT payload (e.g. having a `<footnote_citation>` for a nonexistent footnote), get converted into 500s rather than 400s with helpful error messages.  I think we could deal with that in a separate PR though.

Also note that we could fix this in more than one way; for example, we could change the format for submitting footnote-related data, as suggested in https://github.com/18F/omb-eregs/pull/884#issuecomment-358990289.

toolness:
This fixes #900 by validating that:

* All footnotes have explicit `type_emblem` metadata.
* No duplicate footnotes (that is, footnotes with the same `type_emblem`) exist.
* No footnote citations exist that refer to non-existent footnotes.

If any of these conditions occur, the client will receive an HTTP 400 with an informative help message, instead of a 500.


api/document/serializers/doc_cursor.py
```diff
@@ -118,15 +118,55 @@ def __init__(self, *args, **kwargs):
...
         )
         return MetaSerializer(meta, context={'parent_serializer': self}).data
 
+    def validate_footnote_citations(self, data: PrimitiveDict,
+                                    footnote_emblems: Set[str]):
+        citations = [
+            util.get_content_text(f['inlines']).strip()
+            for f in util.iter_inlines(data['content'])
+            if f['content_type'] == 'footnote_citation'
+        ]
```
cmc333333: Inspecting the `PrimitiveDict` this deeply feels wrong. We have a somewhat sophisticated set of navigation tools for DocCursor already -- should we delay this validation until we have more of the tree constructed?

api/document/serializers/doc_cursor.py
```diff
@@ -118,15 +118,55 @@ def __init__(self, *args, **kwargs):
...
+            if citation not in footnote_emblems:
+                raise ValidationError(
+                    f"Citation for '{citation}' has no matching footnote"
+                )
+
+    def validate_footnote_emblems(self, data: PrimitiveDict) -> Set[str]:
+        footnote_emblems = [
+            f['type_emblem'] for f in util.iter_children(data['children'])
+            if f['node_type'] == 'footnote'
+        ]
```
cmc333333: Similarly, this feels like a validation that should occur _after_ creating the tree. What do you think?

api/document/serializers/doc_cursor.py
```diff
@@ -118,15 +118,55 @@ def __init__(self, *args, **kwargs):
...
+        return set(footnote_emblems)
+
+    def to_internal_value(self, data: PrimitiveDict) -> PrimitiveDict:
+        data = super().to_internal_value(data)
+        if data['node_type'] == 'footnote' and not data.get('type_emblem'):
+            raise ValidationError('Footnotes must have type emblems')
+        if self.is_root:
+            footnote_emblems = self.validate_footnote_emblems(data)
+            self.validate_footnote_citations(data, footnote_emblems)
+        return data
```
cmc333333: Looks like the [`validate()`](http://www.django-rest-framework.org/api-guide/serializers/#object-level-validation) serializer-level method might be preferable to overriding `to_internal_value`.

api/document/tests/serializers/doc_cursor_test.py
```diff
@@ -295,6 +295,53 @@ def test_children_field_to_internal_value_works():
...
+    serializer.to_internal_value(f.para([], children=[
+        f.footnote(1, []),
+        f.footnote(2, [])
+    ]))
+
+    err_msg = "Multiple footnotes exist with type emblem '1'"
+    with pytest.raises(ValidationError, match=err_msg):
+        serializer.to_internal_value(f.para([], children=[
+            f.para([], children=[f.footnote(1, [])]),
+            f.para([], children=[f.footnote(1, [])]),
```
cmc333333: Any reason the second layer of `para`s is needed in this test? It's different from the "success" case, which makes me wonder if there's more going on.

toolness: Yeah, the reason that there's two layers is because if the footnotes were the child of the same node, the "siblings are not unique across (node_type, type_emblem) values" validation would get triggered instead of the footnote one.

api/document/serializers/doc_cursor.py
toolness: Oh good point.  So many methods, it gets confusing which one to put stuff in sometimes...

api/document/serializers/doc_cursor.py
toolness: So I felt that tension while writing this code, but there are a number of different competing issues at play:

1. DRF's serialization framework _really_ wants validation to occur on data structures that are completely independent of the database, whereas `DocCursor` is fairly coupled to the database. So using `DocCursor` during validation feels like fighting against DRF, at least a little bit.

2. `DocCursor` by itself doesn't have any way of representing annotations/inlines as part of a tree;  though the `JSONAwareCursor` does, its accesses them as primitive dicts, so we'd still need to write our own helpers to navigate through those, like `get_content_text()` and `iter_inlines()`.  So it feels like extra cognitive weight to treat _one_ aspect of the content (the nodes) as a nice `DiGraph`-based tree, while treating another aspect of the content (the content/inlines) as not-a-tree, even though it ultimately is, at least the way the client has delivered it to us.

That said, I could see a middle ground where we at least start constructing a `DocCursor`-based tree, independent of the database, in the `to_internal_value()` methods of our serializers, though we'd need to either come up with an alternative `DiGraph` structure for our annotations or keep handling them as JSON blobs accessible via a `JSONAwareCursor`.  But that said, it feels both like a non-trivial refactoring, and like something we could do later on, independently of this PR (though I suppose this PR does slightly increase the amount of code we'd need to migrate over to the `DocCursor`-based way of doing things).

At the very least, could I commit the _tests_ from this PR as `xfail` tests, so that we could fix them later, through whatever means we eventually decide on?

cmc333333: You bring up very good points. I _think_ DRF may be the wrong layer for this sort of validation. I wonder if there's an analogy why Django's models -- fields have validations, models can have validations cross field, but cross-model validations are in the form-handling layer.

Regarding annotation tracking, let me take a crack at #823; I think that might be a cleaner way to keep all the models in memory _and_ keep them related to each other.

All that said, this is definitely an improvement over _no_ validations. I just wanted to pause us a little to think through some alternatives.

api/document/serializers/doc_cursor.py
cmc333333: I'd argue that sort of unknown inheritance problem is one of the biggest deficits of OO.

api/document/serializers/doc_cursor.py
toolness: Noice, that sounds good!

How soon are you thinking of forging #823?  I had an idea for unifying `json_importer.annotations` with `serializers.content`, to reduce the number of places new annotation types had to create new code at, but it can wait...

api/document/serializers/doc_cursor.py
toolness: Yeah! Down with OO!

I like typeclasses and traits, tho...

PR #954, opened 2018-01-25
Initial editor implementation for displaying/editing lists

cmc333333:
This displays lists and list items, including their marker text using the same styles (sans fonts) used in the frontend. There's no way to add a new list item at the moment, but this should get us started.

This allows the marker text (e.g. "1.") to be edited, which will likely be overly confusing for end users. We'll need to figure out what to do there as a team.

Looks like (please ignore misalignment due to font sizes)
screen shot 2018-01-25 at 10 10 18 am

### DIFF
--- a/api/document/js/__tests__/parse-doc.test.ts
+++ b/api/document/js/__tests__/parse-doc.test.ts
@@ -1,6 +1,7 @@
 import axios from 'axios';
 
 import parseDoc, { convertContent } from '../parse-doc';
+import { apiFactory } from '../serialize-doc';
 
 jest.mock('axios');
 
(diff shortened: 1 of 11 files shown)


cmc333333:
I agree that this would be confusing to users and would posit that the _need_ to modify these markers won't come up very often. On the other hand, I think one of the prime directives of the editor is that it shouldn't lose data (the "unimplemented" bits are a wonderful example), so replacing whatever marker is there with a renumbered list isn't viable.

What do you think about us splitting the difference between those two ends of the spectrum and encoding the markers as data on their respective nodes? From the user's perspective, it'd look the same as that screenshot above, but the markers wouldn't be editable. We could add logic around adding a new list item that figures out what the following marker should be.

cmc333333:
I'm going to go ahead and merge this now; we can discuss editable markers at a standup.

toolness:
Oh sounds good, sorry I didn't respond to your last question more promptly!

Issue #917, opened 2018-01-22
Carry over fonts from the UI to the editor


cmc333333:
Depends on #890

cmc333333:
#955 resolves

PR #955, opened 2018-01-25
Override fonts in Django admin, editor

### LINKED ISSUE #916: Carry over fonts from the UI to the admin
Depends on #890

cmc333333:
This should resolve #916 and #917 by overriding the default fonts in Django's admin. Not a radical change, but worth the consistency.

## Before
screen shot 2018-01-25 at 11 03 07 am
screen shot 2018-01-25 at 11 03 29 am
screen shot 2018-01-25 at 11 03 46 am

## After
screen shot 2018-01-25 at 11 04 24 am
screen shot 2018-01-25 at 11 04 33 am
screen shot 2018-01-25 at 11 04 44 am

### DIFF
--- a/api/ereqs_admin/static/admin/layout/_typography.scss
+++ b/api/ereqs_admin/static/admin/layout/_typography.scss
@@ -1,0 +1,20 @@
+// --- Rules to override Django's font selection
+body {
+  @include font-san-serif();
+  @include size-regular();
+}
+
+td,
+th {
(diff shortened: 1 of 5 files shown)


cmc333333:
Thanks @tadhg-ohiggins! cc @rememberlenny, this should integrate pretty well with what you're working on, but there'll probably be a few small file merges

Issue #855, opened 2018-01-12
Design, content, and schema audit of current CRUFT system


austinhernandez:
Design and content audit on the admin side to see what content is back there now that still needs to be accounted for. This issue is semi-related to #852 which audits the preamble content.

## AC
-  [ ] Look at a current CRUFT system to catalog elements, content, and stored data
- [ ] A decision is made about whether we need to keep it or not
- [ ] Consult with Product Owner for final approval

carodew:
Linking to @austinhernandez 's guidance about what is currently in the admin system here – 

https://app.zenhub.com/workspace/o/18f/omb-eregs/issues/855#issuecomment-357335745

carodew:
Nothing an assumption/constraint – per convo between @rtwell  @cmc333333 – the importer can suck in any (or most) of the preamble information that's shown on the first page, but won't be able to identify them as belonging to separate form fields. For example, it will pull in the title text, but dump it at the top of the main text box.

carodew:
Here is the preamble information as it stands right now on CRUFT. I assumed the gray input field labels meant that this information was not necessary, but I may be wrong. @cmc333333 mentioned it might be necessary to sort and categorize. 

* Policy number [mandatory]
* Title  [mandatory]
* Uri (I assume this is URL) [mandatory]
* Omb policy id
* Policy type
  * None
  * Circular
  * Executive
  * Guidance
  * HSPD
  * Law
  * Memoranda
  * National Action Plan
  * Plan
  * PPD
  * Policy Review
  * Standards
  * Strategy 
* Slug [mandatory]
* Issuance [mandatory]
* Sunset
* Policy Status
* Document Source
* Public (this is a checkbox)
* Issuing body [mandatory]
* Managing offices

![cruftpreamble](https://user-images.githubusercontent.com/23194720/34891983-8c02b1b6-f7a5-11e7-990f-18795fda5606.png)

rtwell:
I spoke with @cmc333333 about the origins of the preamble information. Here's a quick overview:

* **Policy number** 🚫  [mandatory] old system, big spreadsheet (internal number, only for the sake of the spreadsheet. 
* **Title** ✅  [mandatory] 
* **Url** 🚫  (I assume this is URL) [mandatory] (cruft pointed to WH.gov / 404s, but we want to link to the PDF that the user uploaded. although what is the “real” PDF?  We should have the one source of truth in the library (when the WH admin switched, all those links broke). 
[...]

PR #956, opened 2018-01-25
"Archive" unnecessary fields

### LINKED ISSUE #908: Resolve public vs workflow field
We currently only allow `public` policies to be accessed in the document endpoint. We recently added a workflow phase field which describes the _document_'s phase of being validation.

Do we replace the `public` field with that document workflow? Which policies are we listing and which load in the document view? How do we ensure that logged in users _do_ have access to edit documents?

cmc333333:
Per @rtwell and @austinhernandez's [analysis](https://github.com/18F/omb-eregs/issues/855), a large number of fields we imported from the original data set are no longer helpful. This keeps imported data, but hides a swath of fields from the admin, api, and ui.

This keeps `public` (until we resolve #908) and `workflow_phase` (until we implement automated transitions). It also does nothing about requirements, though we'll likely want to archive them soon.

## After
screen shot 2018-01-25 at 3 00 48 pm


api/ereqs_admin/tests/admin_tests.py
```diff
@@ -85,6 +85,7 @@ def test_user_edit_max(admin_client):
     assert user.first_name == 'posttest'
 
 
+@pytest.mark.skip(reason="We've temporarily removed pdf upload")
```
toolness: Wait, we _supported_ pdf upload?! 😂

tadhg-ohiggins: Heh, indeed, although just in the file sense, without any import-and-convert aspect.

api/reqs/models.py
```diff
@@ -108,30 +108,32 @@ def __str__(self):
...
-    )
     slug = models.SlugField(max_length=title.max_length)
     issuance = models.DateField()
     sunset = models.DateField(blank=True, null=True)
-    policy_status = models.CharField(max_length=256, blank=True)
-    document_source = models.FileField(blank=True)
     public = models.BooleanField(default=True)
-    issuing_body = models.CharField(max_length=512)
-    managing_offices = models.ManyToManyField(
-        Office, blank=True, related_name='policies')
```
tadhg-ohiggins: I have this feeling that a bunch of the above are going to come back, but I can see the argument for removing them now.

cmc333333: Agreed! I suspect they might also come back on a different form.

cmc333333:
Thanks @toolness and @tadhg-ohiggins!

Issue #904, opened 2018-01-22
Semantic errors when editing AKN XML aren't very helpful


toolness:
Because of the way Django REST Framework's serialization pipeline works, non-syntax errors that make it to our AKN editor (#884) aren't very friendly.  From https://github.com/18F/omb-eregs/pull/868#issuecomment-358477250:

> Oh, I should note that any 400 responses returned from XML PUTs might not be super easy to understand, since they'll generally be coming from the serializer, which is thinking in terms of JSON PUT requests rather than XML ones.  So for instance, errors might mention something about `node_type` or `content_type`, which is basically a tag name in XML, rather than an explicit field.

toolness:
Hmm, so in trying to figure out how to report better errors, I noticed something: when we instantiate a serializer in some of our `to_internal_value()` implementations, such as that of `ChildrenField`, passing `many=True` to the serializer and passing it an array, rather than passing `many=False` and iterating through items ourselves (which makes mypy happier), makes a difference in validation errors that occur.

For example, in our current codebase, if I load M-16-19 and remove the `emblem` attribute from the first footnote, I get this error:

```json
{
  "children": {
    "children": {
      "children": [
        "Footnotes must have type emblems"
      ]
    }
  }
}
```

However, if I make the following tweak to the `to_internal_value()` implementation of `ChildrenField`:

```diff
                           data: List[PrimitiveDict]) -> List[PrimitiveDict]:
         serializer = DocCursorSerializer(
             context={**self.context, 'is_root': False},
+            many=True
         )
-        children = [serializer.to_internal_value(item) for item in data]
+        children = serializer.to_internal_value(data)
         self.validate_type_emblem_uniqueness(children)
         return children
```

I get the following error instead:

```json
{
  "children": [
    {},
    {
      "children": [
        {},
        {
          "children": [
            [
              "Footnotes must have type emblems"
            ]
          ]
        },
        {},
        {},
        {}
      ]
    },
    {}
  ]
}
```

While neither output is particularly elegant, the latter definitely conveys more information: the former merely tells us that some `<footnote>` three levels deep doesn't have an emblem, whereas the latter gives us an actual path to the offending node.  So we might want to just squelch mypy and use `many=True` instead of rolling our own iteration--either that, or write our own functionality that constructs `ValidationError` objects in a similar way.

toolness:
[...]

PR #959, opened 2018-01-26
Make ValidationErrors more detailed/accurate

toolness:
This helps with #904 by making `ValidationError` exceptions raised by our `DocCursorSerializer` less vague, in that they will contain enough data to allow someone to figure out _where_ in their payload a validation error occurred. See https://github.com/18F/omb-eregs/issues/904#issuecomment-360800292 for more details.

### DIFF
--- a/api/document/serializers/util.py
+++ b/api/document/serializers/util.py
@@ -1,4 +1,6 @@
-from typing import Iterator, List, Set, TypeVar  # noqa
+from typing import Iterator, List, Set, Type, TypeVar  # noqa
+
+from rest_framework.serializers import Field
 
 from document.tree import PrimitiveDict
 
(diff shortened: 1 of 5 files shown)


api/document/serializers/util.py
```diff
@@ -43,3 +45,16 @@ def iter_inlines(inlines: List[PrimitiveDict]) -> Iterator[PrimitiveDict]:
...
+
+def list_to_internal_value(data: List[PrimitiveDict],
+                           field_class: Type[Field],
+                           *args, **kwargs) -> List[PrimitiveDict]:
+    # The only real reason this function exists is because
+    # the way DRF's `many=True` changes the type signatures of
+    # a serializer's methods is extremely hard/impossible to
+    # express as a type annotation. So this is really just a
+    # way to "twist mypy's arm" into annotating things the way we
+    # want.
```
toolness: 😭

api/document/serializers/util.py
```diff
@@ -43,3 +45,16 @@ def iter_inlines(inlines: List[PrimitiveDict]) -> Iterator[PrimitiveDict]:
...
+    serializer = field_class(*args, **{'many': True, **kwargs})
+    return serializer.to_internal_value(data)
```
toolness: I just realized that given https://github.com/18F/omb-eregs/pull/961#issuecomment-361252229, it might be smarter to do this the "proper" way for `Serializer` subclasses, e .g. by passing `data` to their constructor, then calling `.is_valid()`, then returning `.validated_data`.  If we do that, we should also change the type signature of this function to ensure that `field_class` is really `serializer_class` and is of `Type[Serializer]` (which should be OK, I think that we're only ever passing such objects to this function anyways).

cmc333333: I _do_ like propriety.

Issue #957, opened 2018-01-26
Pasting a single footnote into the AKN editor and saving causes a 500


toolness:
Pasting the following into the AKN editor:

```xml
<footnote>
  hi
</footnote>
```

raises a 500 with an exception message of `ValueError: too many values to unpack (expected 2)`, which appears to be occurring during the handling of the actual error that was raised. The whole stack traceback comes from files in Django or DRF.

toolness:
Ok, it looks like this is happening because I didn't heed @cmc333333's advice in https://github.com/18F/omb-eregs/pull/949#discussion_r163854955 and move the "footnotes must have type emblems" validation from `to_internal_value()` to `validate()`.

Although I can't find documentation about it in DRF, from reading the source code, it seems like any `ValidationError` we raise in `to_internal_value` needs to have a `detail` that is a dictionary, not a string, where the key is the field the error is related to (or `non_field_errors` if it's not field-specific). Much easier is just raising a string-based `detail` in `validate()`, which automatically gets converted to a dict-style `detail` with a `non_field_errors` key by DRF.

PR #960, opened 2018-01-26
Add source line metadata to validation errors if possible

toolness:
~~**Note: this PR is targeted against #959, not `master`.**~~

This helps with #904 by "annotating" validation errors with source line metadata. Accomplishing this requires two steps:

1. The AKN parser "annotates" each node and annotation object with a `_sourceline` property that indicates the line where the XML element representing the object came from.

2. During validation, our node and annotation serializers catch any raised `ValidationError`s and "annotate" them with the `_sourceline` property passed into them by the upstream parser. If the upstream parser didn't include a `_sourceline` property (e.g. DRF's default JSON parser), nothing additional is done. The `ValidationError` is then re-raised.

Here's example validation error output of an <code>&lt;external_link&gt;</code> tag that lacks an <code>href</code> attribute:

```json
{
  "children": [
    {
      "children": [
        {
          "content": [
            {},
            {
              "href": [
                "This field is required."
              ],
              "_sourceline": "4"
            },
            {}
          ],
          "_sourceline": "3"
        },
        {},
        {},
        {},
        {}
      ],
      "_sourceline": "2"
    },
    {},
    {}
  ],
  "_sourceline": "1"
}
```

## To do

- [x] Add tests
- [x] ~~Consider moving some of our `ValidationError` raising from `to_internal_value()` to `validate()` (which will effectively convert the raised error into a dictionary that can be annotated with a source line). **Update:** Given https://github.com/18F/omb-eregs/issues/957#issuecomment-360925181, we really ought to do this, though it can (and probably should) be done in a separate PR.~~ Done in #961.

@cmc333333 does this kind of strategy seem OK to you?  There doesn't seem to be any particular guidance on how to do this kind of thing in the DRF docs, so I'm basically improvising, but I'm not sure how sound (or just plain understandable) my approach is.

### DIFF
--- a/api/document/serializers/util.py
+++ b/api/document/serializers/util.py
@@ -1,7 +1,8 @@
-from typing import Iterator, List, Set, Type, TypeVar  # noqa
+from contextlib import contextmanager
+from typing import Any, Iterator, List, Set, Type, TypeVar  # noqa
 
 from rest_framework.exceptions import ValidationError
-from rest_framework.serializers import Field, empty
+from rest_framework.serializers import Field
 
 from document.tree import PrimitiveDict
 
@@ -61,23 +62,26 @@ def list_to_internal_value(data: List[PrimitiveDict],
     return serializer.to_internal_value(data)
 
 
-class SourcelineErrorMixin:
+def has_sourceline(data: Any):
+    if not isinstance(data, dict):
+        return False
+    return '_sourceline' in data
+
+
+@contextmanager
+def add_sourceline_to_errors(data: Any):
     '''
-    A mixin for DRF Serializers that adds metadata about the
+    A context manager for DRF Serializers that adds metadata about the
     source line number that validation errors originated
     from, if possible.
 
     In order for this to work, the data passed to the
-    Serializer must have source line number metadata.
+    context manager must have source line number metadata.
     '''
 
-    def run_validation(self, data=empty):
-        try:
-            return super().run_validation(data)  # type: ignore
-        except ValidationError as e:
-            if (isinstance(e.detail, dict) and
-                    '_sourceline' not in e.detail and
-                    isinstance(data, dict) and
-                    isinstance(data.get('_sourceline'), int)):
-                e.detail['_sourceline'] = data['_sourceline']
-            raise e
+    try:
+        yield
+    except ValidationError as e:
+        if not has_sourceline(e.detail) and has_sourceline(data):
+            e.detail['_sourceline'] = data['_sourceline']
+        raise e
--- a/api/document/serializers/util.py
+++ b/api/document/serializers/util.py
@@ -48,10 +48,8 @@ def iter_inlines(inlines: List[PrimitiveDict]) -> Iterator[PrimitiveDict]:
         yield from iter_inlines(inline['inlines'])
 
 
-def has_sourceline(data: Any):
-    if not isinstance(data, dict):
-        return False
-    return '_sourceline' in data
+def has_sourceline(data: Any) -> bool:
+    return isinstance(data, dict) and '_sourceline' in data
 
 
 @contextmanager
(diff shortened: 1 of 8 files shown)


api/document/serializers/util.py
```diff
@@ -58,3 +59,25 @@ def list_to_internal_value(data: List[PrimitiveDict],
...
+    source line number that validation errors originated
+    from, if possible.
+
+    In order for this to work, the data passed to the
+    Serializer must have source line number metadata.
+    '''
+
+    def run_validation(self, data=empty):
+        try:
+            return super().run_validation(data)  # type: ignore
```
toolness: I had to tell mypy to ignore this, as otherwise it would complain that this class' superclass has no `run_validation()` method.

Hmm, maybe mixins actually aren't the right approach here. I could turn this into a context manager instead, and then just add a trivial `run_validation()` to the two classes that need it, which simply wraps `super().run_validation()` in the context manager.  Or for that matter I could make it a decorator...

toolness:
For reference, two other approaches I considered:

* The `ValidationError` that ultimately comes out of the serializer in the end has enough metadata in it, thanks to #959, to pinpoint the tag that any error originated from. So theoretically we could potentially trace errors back to their XML source in a DRF [custom exception handler](http://www.django-rest-framework.org/api-guide/exceptions/#custom-exception-handling) rather than in the serializer itself. However, doing this seemed like more work so I didn't pursue it.

* I was thinking that if the validation error `detail` raised wasn't actually a dictionary, I could find any error messages in the detail and append something like `(line {_sourceline})` to the end of them, so that the actual error message itself is more helpful.  I guess we could still do this, the approach isn't incompatible w/ the one I chose for this PR, but it is additional work. Also, just modifying our current `ValidationError` raises to occur during `validate()` instead of `to_internal_value()`, which would ensure that more of our `ValidationError`s have dictionary `detail`s (and which is currently a TODO in this PR), seems like a more robust approach anyways.

api/document/serializers/util.py
cmc333333: I like that context managers would be a but more explicit about when the "magic" is happening, so I'd lean that direction.

api/document/serializers/util.py
```diff
@@ -58,3 +59,25 @@ def list_to_internal_value(data: List[PrimitiveDict],
...
+        except ValidationError as e:
+            if (isinstance(e.detail, dict) and
+                    '_sourceline' not in e.detail and
+                    isinstance(data, dict) and
+                    isinstance(data.get('_sourceline'), int)):
```
cmc333333: You've pointed to a few potential improvements in your comments, but I just wanted to flag this group of clauses as kinda hairy.

I'm not sure it's possible, but may be worth trying: if my mental model is right, there's a serializer instance associated with each object that gets deserialized. I could imagine the _state_ of those instances including the source line rather than tacking it on to the data. If that actually worked, we could reduce the number of corner cases that we need to check here.

toolness:
Cool, thanks @cmc333333! Good question about the deeply nested error messages, I will start looking into the APIs of web-based linters and such.  I think this first pass is ready for code review!

api/document/serializers/util.py
toolness: Hmm, I will think about if there's a way to simplify that gnarlyness a bit more, but right now I'm not sure if there is... Hmmmmmm.

api/document/serializers/util.py
```diff
@@ -58,3 +60,29 @@ def list_to_internal_value(data: List[PrimitiveDict],
     # want.
     serializer = field_class(*args, **{'many': True, **kwargs})
     return serializer.to_internal_value(data)
+
+
+def has_sourceline(data: Any):
+    if not isinstance(data, dict):
+        return False
+    return '_sourceline' in data
```
cmc333333: Would the one-liner version be too cute?
```python
return isinstance(data, dict) and '_sourceline' in data
```

PR #961, opened 2018-01-29
Ensure ValidationErrors raised by serializers are dictionaries

### LINKED ISSUE #957: Pasting a single footnote into the AKN editor and saving causes a 500

toolness:
This fixes #957.

I find validation for DRF very confusing; in _some_ cases, [such as simple fields](http://www.django-rest-framework.org/api-guide/fields/#raising-validation-errors), it's recommended to raise one with a string `detail` during `to_internal_value()`. The DRF code that calls this method takes care of turning such errors into ones with dict-based `detail`.

However, in _other_ cases, such as [deserializers with multiple fields](http://www.django-rest-framework.org/api-guide/serializers/#overriding-serialization-and-deserialization-behavior), there are multiple options, one of which is also raise a `ValidationError` during `to_internal_value()`, _except passing a dictionary instead of a string for `detail`_.  This is super confusing and DRF doesn't even seem to care whether we follow it, until something unexpected like #957 occurs.

### DIFF
--- a/api/document/serializers/content.py
+++ b/api/document/serializers/content.py
@@ -198,9 +198,11 @@ class NestedAnnotationSerializer(serializers.Serializer):
     def to_internal_value(self, data: PrimitiveDict) -> PrimitiveDict:
         content_type = data.get('content_type')
         if content_type is None:
-            raise ValidationError("missing content_type")
+            raise ValidationError({"content_type": "This field is required."})
         if content_type not in self.content_type_mapping:
-            raise ValidationError(f"unknown content_type: {content_type}")
+            raise ValidationError({
+                "content_type": f"'{content_type}' is an invalid content type."
+            })
         serializer = self.content_type_mapping[content_type]()
         return serializer.to_internal_value(data)
 
--- a/api/document/serializers/doc_cursor.py
+++ b/api/document/serializers/doc_cursor.py
@@ -2,7 +2,6 @@ from typing import List, Set, Tuple, Iterator  # noqa
 
 from rest_framework import serializers
 from rest_framework.exceptions import ValidationError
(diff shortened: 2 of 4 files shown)


toolness:
Note that I originally tried moving our validation logic in our `DocCursorSerializer` to `validate()`, which is the recommended behavior, but this didn't work out well because validating our footnotes requires fully normalized fields, which the data passed to `validate()` is not necessarily... err, at least when I call `validate()` directly in unit tests... Actually there might be a way a way around that, let me check something.

toolness:
Oof, ok, disregard my last comment; I've now moved all `DocCursorSerializer` logic from `.to_internal_value()` to `.validate()`; it just took a bit of test case rejiggering, but the actual implementation is simpler now.

toolness:
I'm now wondering if, when we call out to other `Serializer` subclasses for deserialization in our `ChildrenField` and our `ContentField`, we should actually be using them [as recommended for deserialization](http://www.django-rest-framework.org/api-guide/serializers/#deserializing-objects)--e .g. by passing `data` to their constructor, then calling `.is_valid()`, then returning `.validated_data`--rather than directly calling `to_internal_value()` ourselves.  I am pretty sure I committed the code that does the latter, and it was probably noobish of me to do so. Though we should wait for #959 to land before changing anything in that regard.

toolness:
Yep, you're correct, as far as I can tell--I just wish DRF was a bit more explicit about it somehow. Ah well.

PR #963, opened 2018-01-29
Run mypy on all files, instead of a whitelist

toolness:
We added mypy support in #802, but we only run on whitelisted files specified by `mypy-files.txt`.  However, this is kind of a hassle to keep updated, and it's quite possible we can _forget_ to update it even when we add type annotations to a file. So this PR just runs mypy on everything, but applies module/package/filename-specific [configuration rules](http://mypy.readthedocs.io/en/latest/config_file.html) where needed to squelch certain kinds of errors. (In the future, we can remove these as we make more of our code amenable to mypy.)

## To do

- [x] Update `README.md` and/or any other documentation as needed.
- [x] Remove `mypy-files.txt` and update CircleCI config.


api/omb_eregs/settings.py
```diff
@@ -35,7 +36,7 @@
 
 # Application definition
 
-INSTALLED_APPS = (
+INSTALLED_APPS: Tuple[str, ...] = (
```
toolness: These kinds of annotations are needed because we're using tuples as immutable lists here, not as data types whose dimensions and types need to remain constant.

PR #964, opened 2018-01-29
Adding button for add-paragraph, refactor to factories

cmc333333:
We'll most likely want to do paragraph splitting, collapsing, and other fun logic, but this starts us off by allowing a paragraph to be added through a button press. We'll no doubt be re-working this soon.

Along the way, it refactors all of the node-creation bits to use factories instead. That's particularly helpful for nodes with complex structures, like paragraphs and list items. We can split this into two PRs if that'd make review easier; let me know.

## Looks like
create-paragraph


toolness:
This is kind of hard core, partly because I'm unfamiliar with adding functionality like this to ProseMirror--can you walk me through it over a screen share session?  It's ok if not, it'll just take me a little while to review!

cmc333333:
:+1: Thanks @toolness . I'm going to split this into two PRs so I can work off the refactors in a different branch

PR #969, opened 2018-01-30
Use factories when creating ProseMirror nodes, alert on schema fail

cmc333333:
In a few spots, our ProseMirror schema is a bit complex (having required nodes, needing specific attributes, etc.). Rather than require we remember the schema, we can use factory methods with parameters for each of the relevant bits.

This also alerts the user on schema validation failure (we had been loading an invalid doc structure without knowing).

### DIFF
--- a/api/document/js/Api.ts
+++ b/api/document/js/Api.ts
@@ -9,7 +9,7 @@ function setStatus(msg: string, className: 'editor-status-error'|'' = '') {
   status.className = `editor-status ${className}`;
 }
 
-function setStatusError(e: Error) {
+export function setStatusError(e: Error) {
   let errMsg = 'An error occurred.';
   const data = e['response'] && e['response']['data'];
   if (data) {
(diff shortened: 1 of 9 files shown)


cmc333333:
Thanks @toolness ! I think making PRs to non-master solves for a different problem; they seem great for offering an addendum or a different solution, but they're ultimately modifying the PR they build off. In this case, I wanted to split the PR so that the first chunk could be merged separately (and earlier) than the later ones, which isn't super easy with the PR-to-PR approach. If Github allowed a PR to switch branches, I think we'd be set.

I usually include a diff when building on other PRs that helps with some of that; sorry to get lazy this time!

I'll rebase and merge.

toolness:
Hmm, in most of the cases where I've done a PR-to-PR thing, I've just re-targeted the second PR to `master` once the first lands, which is what I did with #959 and #960.  Personally I find it easier, but if you include a diff when building on other PRs that seems like it'd work well too!

Another thing that might help could be to just put **This PR builds on #XXX, please review that before reviewing this** in bold at the top of a PR, just to telegraph what's going on.

cmc333333:
Oh hey, I didn't think we _could_ retarget a PR. That solves the problem quite well. Thanks @toolness !

toolness:
Oh sure thing--I believe re-targeting is a relatively recent feature that they added in the past year or so, so it's easy to miss!

PR #970, opened 2018-01-30
Add button for add-paragraph

cmc333333:
This builds on #969. We'll most likely want to do paragraph splitting, collapsing, and other fun logic, but this starts us off by allowing a paragraph to be added through a button press. We'll no doubt be re-working this soon.

## Looks like
create-paragraph

### DIFF
--- a/api/document/js/commands.ts
+++ b/api/document/js/commands.ts
@@ -1,0 +1,26 @@
+import { Node } from 'prosemirror-model';
+import { TextSelection } from 'prosemirror-state';
+
+import { factory } from './schema';
+
+export function addParagraph(state, dispatch) {
+  if (!dispatch) {
+    return true;
+  }
+  const pos = state.selection.$head;
+  let depth = pos.depth;
+  while (depth >= 0) {
+    const node = pos.node(depth);
+    if (node.type.spec.group === 'block') {
+      const insertPos = pos.after(depth);
+      let tr = state.tr.insert(insertPos, factory.para(' '));
+      // + 2 to get us inside the inline
+      tr = tr.setSelection(TextSelection.create(tr.doc, insertPos + 2));
+      dispatch(tr.scrollIntoView());
+      return true;
+    }
+    depth -= 1;
+  }
+
+  return true;
+}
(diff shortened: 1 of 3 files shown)


api/document/js/commands.ts
```diff
@@ -0,0 +1,26 @@
...
+  const pos = state.selection.$head;
+  let depth = pos.depth;
+  while (depth >= 0) {
+    const node = pos.node(depth);
+    if (node.type.spec.group === 'block') {
+      const insertPos = pos.after(depth);
+      let tr = state.tr.insert(insertPos, factory.para(' '));
+      // + 2 to get us inside the inline
+      tr = tr.setSelection(TextSelection.create(tr.doc, insertPos + 2));
+      dispatch(tr.scrollIntoView());
```
toolness: Do we want to verify that the document still conforms to the schema after the transaction is executed, and roll it back if not?  (Or alternatively, is there a redux-esque-middleware we could add that does this for all transactions, at least during development?)

api/document/js/menu.ts
```diff
@@ -1,4 +1,18 @@
...
+import { addParagraph } from './commands';
+
+
+const newParagraph = new MenuItem({
+  css: 'cursor: pointer;',
+  title: 'Add paragraph',
+  run: addParagraph,
+  label: 'P',
+  // These defaults are needed due to a doc issue. See
+  // https://github.com/ProseMirror/prosemirror-menu/issues/15
```
toolness: Thanks for linking to the issue here!

api/document/js/commands.ts
```diff
@@ -0,0 +1,26 @@
+import { Node } from 'prosemirror-model';
+import { TextSelection } from 'prosemirror-state';
+
+import { factory } from './schema';
+
+export function addParagraph(state, dispatch) {
```
toolness: I wonder if it'd be useful to add a comment above this function explaining exactly what the function does, e.g. "Appends a paragraph to the closest valid point after the head of the current selection"?  Eh, just a thought I guess.

tadhg-ohiggins: I had that thought too, I think it's a good idea.

api/document/js/commands.ts
cmc333333: I'll throw in a `.check()` after the dispatch. That'll raise an exception, but should keep chugging along.

cmc333333:
Thanks for the comments, both! Updated, now will rebase and merge.

PR #971, opened 2018-01-30
Make errors in AKN XML editor a bit friendlier.

toolness:
This makes error output a bit friendlier. It actually affects the ProseMirror editor too, but most of the enhancements are specific to the XML editor:

* When an error is just a string, such as the raw content of a Django 500 traceback, we just print out the whole string, rather than printing its human-unfriendly JSON representation.

* When an error is just a JS object with a single `detail` key that has a string value, we just print out that value, instead of the whole object's JSON representation.

* When an error has `_sourceline` metadata in it, we assume that the input was XML, and we provide slightly friendlier help than just printing out a big JSON blob. It's still far from ideal, but it's a start, and it's much better than just printing out a JSON blob.

### DIFF
--- a/api/document/js/Api.ts
+++ b/api/document/js/Api.ts
@@ -1,6 +1,7 @@
 import axios from 'axios';
 
 import { getEl } from './util';
+import { makeErrorFriendly } from './friendly-error';
 
 function setStatus(msg: string, className: 'editor-status-error'|'' = '') {
   const status = getEl('#status');
@@ -9,10 +10,10 @@ function setStatus(msg: string, className: 'editor-status-error'|'' = '') {
 }
 
 function setStatusError(e: Error) {
-  let errMsg = `An error occurred: ${e}`;
+  let errMsg = 'An error occurred.';
   const data = e['response'] && e['response']['data'];
   if (data) {
-    errMsg += '\n' + JSON.stringify(data, null, 2);
+    errMsg += '\n' + makeErrorFriendly(data);
   }
   setStatus(errMsg, 'editor-status-error');
 }
--- a/api/document/js/__tests__/friendly-error.test.ts
+++ b/api/document/js/__tests__/friendly-error.test.ts
@@ -1,0 +1,98 @@
+import { makeErrorFriendly } from '../friendly-error';
+
+describe('makeErrorFriendly()', () => {
+  it('converts simple "detail" errors', () => {
+    const detail = ('XML syntax error - expected \'>\', line 4, ' +
+                    'column 3 (<string>, line 4)');
+    expect(makeErrorFriendly({ detail })).toBe(detail);
+  });
+
+  it('returns string errors as-is', () => {
+    const msg = 'NameError at /document/M-16-19\nname \'boom\' is not ...';
+    expect(makeErrorFriendly(msg)).toBe(msg);
+  });
+
+  it('returns other errors as their JSON stringification', () => {
+    expect(makeErrorFriendly(null)).toBe('null');
+  });
+
+  it('provides details on XML errors w/ invalid content types', () => {
+    /* tslint:disable */
+    const err = {
+      "children": [
+        {
+          "content": [
+            {
+              "content_type": "'k' is an invalid content type.",
+              "_sourceline": "3"
+            },
+            {}
+          ],
+          "_sourceline": "2"
+        }
+      ],
+      "_sourceline": "1"
+    };
+    /* tslint:enable */
+    expect(makeErrorFriendly(err)).toBe(
+      `In an element starting at line 3:\n` +
+      `* content_type - 'k' is an invalid content type.`,
+    );
+  });
+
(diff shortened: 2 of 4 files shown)


api/document/js/Api.ts
```diff
@@ -9,10 +10,10 @@ function setStatus(msg: string, className: 'editor-status-error'|'' = '') {
 }
 
 function setStatusError(e: Error) {
-  let errMsg = `An error occurred: ${e}`;
+  let errMsg = 'An error occurred.';
```
toolness: I removed the exception stringification to reduce the verbosity of the error output; note that this _does_ mean that we won't be able to easily distinguish between a 400 and a 500 without looking at our developer tools' "network" tab, but I think that the actual content that follows will be more helpful than the status code in almost all cases.

api/document/js/__tests__/friendly-error.test.ts
```diff
@@ -0,0 +1,98 @@
...
+          ],
+          "_sourceline": "2"
+        }
+      ],
+      "_sourceline": "1"
+    };
+    /* tslint:enable */
+    expect(makeErrorFriendly(err)).toBe(
+      `In an element starting at line 3:\n` +
+      `* content_type - 'k' is an invalid content type.`,
```
toolness: This can be improved in the future, since `content_type` is really the tag name when it comes to XML.

api/document/js/__tests__/friendly-error.test.ts
```diff
@@ -0,0 +1,98 @@
...
+    const err = {
+      "non_field_errors": [
+        "Multiple footnotes exist with type emblem '1'"
+      ],
+      "_sourceline": "1"
+    };
+    /* tslint:enable */
+    expect(makeErrorFriendly(err)).toBe(
+      `In an element starting at line 1:\n` +
+      `* non_field_errors - Multiple footnotes exist with type emblem '1'`,
```
toolness: This can be improved in the future, since `non_field_errors` isn't terribly helpful. This test will give us a foothold through which we can easily tweak our output.

api/document/js/__tests__/friendly-error.test.ts
```diff
@@ -0,0 +1,98 @@
...
+          ],
+          "_sourceline": "3"
+        }
+      ],
+      "_sourceline": "1"
+    };
+    /* tslint:enable */
+    expect(makeErrorFriendly(err)).toBe(
+      `In an element starting at line 4:\n` +
+      `* href - This field is required.`,
```
toolness: This can be improved in the future, since `href` is really referring to an XML attribute rather than a "field".  Also, it suggests that we might want to also provide a `_sourcetag` property in our error output or something, so that we can easily provide the name of the actual element that's lacking an attribute,

Anyhow, this test will give us a foothold through which we can easily change all that.

api/document/js/__tests__/friendly-error.test.ts
```diff
@@ -0,0 +1,98 @@
...
+              ],
+              "_sourceline": "4"
+            }
+          ],
+          "_sourceline": "3"
+        }
+      ],
+      "_sourceline": "1"
+    };
+    /* tslint:enable */
```
toolness: Sorry about all these tslint disablings/enablings, but I was just copy-pasting the JSON stringification from my JS console and fixing all the property names was a hassle. Bleh.

toolness:
Note that, as mentioned in https://github.com/18F/omb-eregs/pull/960#pullrequestreview-92220394, the actual format of the validation errors emitted by our API is not particularly friendly.  I guess another approach to all this might be to first change it to be friendlier, at which point creating a client-side "friendly error" module would involve less code than it does in this PR.

toolness:
Sorry, just pushed two commits to resolve some CodeClimate smells, all is ready now!

api/document/js/friendly-error.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+  if (typeof err === 'object' && err !== null) {
+    if (Object.keys(err).length === 1 &&
+        typeof(err.detail) === 'string') {
+      return err.detail;
+    }
+    if (XMLErrorInfo.isXMLError(err)) {
+      return (new XMLErrorInfo(err)).toString();
+    }
+  }
```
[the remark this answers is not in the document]
cmc333333: Ah, got to this in a latter commit!

api/document/js/friendly-error.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  static isXMLError(err: any): boolean {
+    return (typeof err === 'object' && err !== null &&
+            '_sourceline' in err);
+  }
+
+  constructor(err: object) {
+    this.nestedErrors = [];
+    this.errors = [];
+    this.sourceline = parseInt(err['_sourceline'], 10);
```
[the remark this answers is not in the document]
toolness: Yeah, I feel like they should, and I'm not sure how/why/where DRF is converting them to strings, because we're setting them as ints in our code. Just filed #975 to address it later.

PR #972, opened 2018-01-30
Fix delete-paragraph bug

cmc333333:
Building on #969, this fixes a bug in which deleting a paragraph in the ProseMirror frontend wouldn't delete its entry in the ProseMirror doc. Then that doc was serialized, it'd include an empty paragraph, which led to very unintuitive problems.

### DIFF
--- a/api/document/js/schema.ts
+++ b/api/document/js/schema.ts
@@ -39,7 +39,7 @@ const schema = new Schema({
       toDOM: node => [`h${node.attrs.depth}`, { class: 'node-heading' }, 0],
     },
     para: {
-      content: 'inline? block*',
+      content: 'inline block*',
       group: 'block',
       toDOM: () => ['div', { class: 'node-paragraph' }, 0],
     },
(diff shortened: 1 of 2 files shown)


api/document/js/schema.ts
```diff
@@ -39,7 +39,7 @@ const schema = new Schema({
       toDOM: node => [`h${node.attrs.depth}`, { class: 'node-heading' }, 0],
     },
     para: {
-      content: 'inline? block*',
+      content: 'inline block*',
```
toolness: So it seems like this is the actual fix... interesting... so was the `inline` basically being deleted, leaving only the `block`, e.g. a paragraph with no `<content>` but still having children?

cmc333333: You got it, the `para`'s schema was allowing empty `content`, so deleting that text wasn't actually deleting the `para`. Combine that with an assumption of text with loading `para`s and we kill two birds with one stone by making the schema more rigid.

We *may* want to allow `(inline block*) | block+` in the future, I'm not sure. I suspect we'd need to add custom logic to the deletion commands to "fix" broken docs.

cmc333333:
Thanks @tadhg-ohiggins and @toolness. I'll rebase and merge.

PR #973, opened 2018-01-30
Add css in admin override and provide organized CSS structure

rememberlenny:
Our admin section is built off Django, but we want it to feel closer to the rest of our viewer app. This PR takes another step in that direction by overriding several of Django's built-in style rules with "overrides" that are closer to our mockups. I've tried to specify which Django file and use their selectors for each rule so we can maintain this going forward.

Along the way:
* Moved the insertion of these styles to the end of the `<head>` tag to ensure they are last (mostly)
* Import normalize.css to set some sane defaults

@toolness and @tadhg-ohiggins - any major red flags with this approach? It's a lot of seemingly random overrides, but we should be able to trace everything. @cmc333333'll also take another pass and try to comment as many of the rules as possible.

@cmc333333'll also work with @austinhernandez and @rtwell today to do a design pass.


toolness:
Sorry, is this purely a refactoring of existing admin CSS code, or is it altering the styling?  If it's altering the styling, what is it doing?  (Because I'm almost completely unfamiliar with the admin CSS, it's hard for me to tell.)

cmc333333:
Apologies, @toolness, let me update the PR description.

cmc333333:
cc @xtine

cmc333333:
Thanks @tadhg-ohiggins! I'll rebase to try to clean up this history a bit, then merge.

PR #978, opened 2018-01-31
Serialize section title in ProseMirror editor

### LINKED ISSUE #947: Serialize section title from editor
The ProseMirror editor isn't adding a title field to sections when it serializes them, meaning we lose that from the database.

This may be a good time to encode the associated schema as
```
sect: heading block+
``` 
to ensure we'll always have a heading to grab a title from. If we do that, though, we'll need to verify the pdf parser is emitting that consistently.

toolness:
This fixes #947.

Note that, as explained in https://github.com/18F/omb-eregs/issues/947#issuecomment-362007248, the PDF parser is guaranteed to only create sections with headings after them.

## To do

- [x] Ensure sections start w/ a heading in `schema.ts`.
- [x] During serialization, take the heading's content and set the `title` property of the section to it.
- [x] Add tests.


toolness:
Well, this is unexpected. Simply declaring that a `heading` appear as the first child of a `sec` is triggering an error of `Only non-generatable nodes (heading) after a match state (in content expression 'heading block+')`, which comes from [`content.js` in `prosemirror-model`](https://github.com/ProseMirror/prosemirror-model/blob/8a33390e9a066d421bd2df27ea73c97fb95be339/src/content.js#L360).  I have no idea what "non-generatable nodes" means and web searching doesn't help, though the function throwing the error is called `checkForDeadEnds`, which might provide some clues.

toolness:
Hmm, it looks like if I remove `attrs` from the definition of `heading`, the error is no longer raised. Perhaps by "non-generatable nodes" it means a node that ProseMirror can't auto-create itself if it needs to?

Oh, funky, so if I add a default value for the heading's `depth`, everything works.  I'm not sure why this type of node needs to be something that ProseMirror can generate on its own, but I guess I will set the default to `1` for now, just to appease ProseMirror (it probably won't be used in practice since we're always providing an explicit `depth`).

cmc333333:
:+1: just came to the same conclusion when working through list stuff. Lots of unexpected nooks and crannies!

toolness:
Ok, I think this is ready for review!  While the `converts nested nodes` test happens to exercise this functionality, I thought it best to make a new test that specifically targets it.  I can remove it if you want, though.

api/document/js/__tests__/serialize-doc.test.ts
```diff
@@ -44,6 +45,14 @@ describe('serializeDoc()', () => {
     });
   });
 
+  it('converts heading text to section title', () => {
+    const node = factory.sec([
+      factory.heading('Stufff', 1),
+    ]);
```
cmc333333: Would it make sense to modify the `sec` factory to take required args for headings?
```
factory.sec('Stuff', 1, [other-children])
```
That'll force us to deal with invalid input from the api, too.

toolness: I like it! Just filed #983 to deal with that in a separate PR.

PR #980, opened 2018-01-31
Allow non-public documents to be edited by admins.

cmc333333:
Previously, we weren't allowing documents associated with non-public policies
from being loaded into the editor (nor could they be retrieved in the API).
This should allow *logged in* users to see both.

It also adds a permissions check to the document load logic, which it looks
like we forgot. We don't really worry about different user groups anywhere
(other than logged in vs non), but it's best to lay that groundwork now.

### DIFF
--- a/api/document/views.py
+++ b/api/document/views.py
@@ -23,7 +23,10 @@ class TreeView(GenericAPIView):
     queryset = DocNode.objects.none()   # Used to determine permissions
 
     def get_object(self, prefetch_related=True):
-        policy = policy_or_404(self.kwargs['policy_id'])
+        policy = policy_or_404(
+            self.kwargs['policy_id'],
+            only_public=not self.request.user.is_authenticated,
+        )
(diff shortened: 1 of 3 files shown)


api/document/views.py
```diff
@@ -23,7 +23,10 @@ class TreeView(GenericAPIView):
     queryset = DocNode.objects.none()   # Used to determine permissions
 
     def get_object(self, prefetch_related=True):
-        policy = policy_or_404(self.kwargs['policy_id'])
+        policy = policy_or_404(
+            self.kwargs['policy_id'],
+            only_public=not self.request.user.is_authenticated,
```
tadhg-ohiggins: Overall I think this PR is very straightforward, but I wonder if it would be better to create a variable outside of this call so as to not have to follow the `not` logic in a single line?

cmc333333: Good idea, I'll add. I'll try to resolve the cyclomatic complexity warning, too.

PR #981, opened 2018-02-01
Add "Save and edit document" to policy admin.

cmc333333:
We'd like a stop-gap to get to the editor page from a policy, so we'll add a
new button at the bottom of the Policy change form. We call the Django
template tags and methods and copy-paste the minimum template to reduce
duplication.

## Looks like
save-and-edit

### DIFF
--- a/api/reqs/admin.py
+++ b/api/reqs/admin.py
@@ -1,6 +1,8 @@
 from django import forms
 from django.contrib import admin
 from django.core.exceptions import ValidationError
+from django.http import HttpResponseRedirect
+from django.urls import reverse
 
 from ereqs_admin.revision_admin import EReqsVersionAdmin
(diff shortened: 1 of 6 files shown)


api/reqs/admin.py
```diff
@@ -40,6 +42,15 @@ class PolicyAdmin(EReqsVersionAdmin):
         'workflow_phase',
     ]
 
+    def response_post_save_change(self, request, obj):
```
tadhg-ohiggins: Ah. This is the piece that I wasn't sure about when I was looking at this today.

tadhg-ohiggins: That and how to use templatetags.

cmc333333:
Thanks, @tadhg-ohiggins! I had actually assumed it'd be much easier, that there'd be an array of actions to edit or something similar. Luckily, StackOverflow had my back.

I'll rebase and merge.

PR #984, opened 2018-02-01
Attempt to make "docker-compose up" work, and simplify docker setup.

toolness:
This simplifies our Docker setup and makes development easier in a number of ways:

* `docker-compose up` now works as expected, logging the output of all containers to stdout.
* There are now just 5 containers, corresponding to parts of the system architecture that are (hopefully) easy to remember:
  * `persistent_db` is the Postgres database (this is unchanged from previously).
  * `minio` is the mock S3 backend (also unchanged from previously).
  * `api` is the Django-based API.
  * `api-ui` is the NodeJS-based pipeline that's responsible for creating static assets used by the API UI (perhaps a slightly better name for this might be `admin-ui`, since it's the UI that staff/admin use).
  * `ui` is the Next.js-based app that presents a user interface to the general public.
* The `proxy` container is gone, and containers access each other using standard Docker networking. However, one side effect of this is that I had to add support to the Next.js app to use a different API URL depending on whether the code was executing on the server or client. I did this by adding an optional environment variable called `INTERNAL_API_URL`, which is set when developing with Docker.
* Development tools relevant to each of `api`, `api-ui`, and `ui` can be accessed via `docker-compose run <container-name> <tool-cli-name> <arguments>`, e.g. `docker-compose run ui npm test`.
  * Because `tool-cli-name` is just the name of the tool as it's accessed via the command-line, it's easy for developers to predict, and we don't need to create separate one-off commands every time we want to make one of the tools accessible from the Docker host.
  * This also gives us easy access to commands that we hadn't exposed before, like `isort`.
  * If really weird users (like me) want to, they can simply specify `bash` as the `tool-cli-name` for quick access to all the tools relevant to a particular container.
  * Some development tools, particularly running `psql` and `ptw` with its list of packages, are exceptions to this general rule, but I think we can figure out better ways of accessing these in future PRs.
* Rather than having separate container names for dev/prod instances of containers, we just default to dev mode, and if users want to run prod, we direct them to create a copy of the `.env.sample` file and uncomment the two lines relevant to prod mode.
  * If we _really_ want dev/prod to run on different ports, as they did before, we can do that, we just have to make the user create a `docker-compose.override.yml` file instead of an `.env` file. But I was unclear on why the different modes should run on different ports, so I didn't pursue this option.
* The structure of the `docker-compose.yml` file is simplified, as there are no longer any references to YAML anchors and aliases.  Aside from being (hopefully) easier to understand, it should also make it easier to [extend the compose file via conventional means](https://docs.docker.com/compose/extends/).

## Disadvantages and Trade-offs

The one major disadvantage of this approach is that each container _always_ waits for its dependencies before it runs, because there's no easy way to tell what dependencies will be needed by a particular tool. This can potentially slow things down. However, in practice, it seems like the dependencies are almost always already running, so I'm hoping this shouldn't be a big deal.

Also, developers now need to type the container name in addition to the name of the dev tool they want to use.  However, I think that the trade-offs are worth it here, since we benefit in simpler docker-compose files and a larger selection of dev tools.  Furthermore, since there are identical tools available to different containers, we had to namespace the dev tools anyways (e.g. `docker-compose run api-npm` for the API UI and `docker-compose run npm` for the regular UI), which was just as bad as having to specify a container name.

## To do

- [x] Figure out how to make scripts accessible (e.g. `docker-compose run manage.py`)
- [x] Update `README.md` to mention the new way of using scripts.
- [x] Make CircleCI work.


ui/util/api/endpoints.js
```diff
@@ -14,7 +14,8 @@ const CACHE_CONFIG = {
 
 export class Endpoint {
   constructor(endpoint) {
-    this.client = axios.create({ baseURL: process.env.API_URL });
+    const baseURL = process.env.INTERNAL_API_URL || process.env.API_URL;
```
toolness: I could add some logic here that only uses `INTERNAL_API_URL` if `process.browser` is falsy; while it isn't strictly required to make the code work--`process.env.INTERNAL_API_URL` isn't passed to webpack's environment plugin and therefore will always be falsy when run on the browser-side--it could be helpful in communicating intent to readers.

README.md
```diff
@@ -130,41 +141,40 @@ import into the policy library.
...
-  * `mypy`
-  * `npm`
-  * `pip-compile`
-  * `psql`
-  * `ptw`
-  * `py.test`
-  * `webpack`
+**Note:** You can add `--rm` to the following commands to delete the images
+after running the commands; alternatively, you can run
+`docker system prune` at regular intervals to do the same thing.
```
toolness: IMO we should really remove `--rm` from all our examples and just tell users to either add it or regularly use `docker system prune`.  Having to type `docker-compose run <container-name>` before everything is already quite verbose, so the addition of the `--rm` makes our instructions seem unnecessarily more complex than they need to be, IMO.  But I figured this is out of scope of this PR so I didn't make the change.

toolness:
Note that this PR started out as an experiment but I turned it into a "real" PR once I found a decent way of doing things, but this means that the commit history is a bit convoluted.  If that's a problem, I can either start a new PR, or we can merge this one with squashing.

ui/util/api/endpoints.js
cmc333333: That's a smart idea; it is rather cryptic now.

api/.docker/modify_path_then
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+export PATH=$PATH:$PWD/node_modules/.bin
```
cmc333333: Not for now, but this is probably an indication that we should push these into npm scripts, hey?

cmc333333:
Re merging: I didn't look too too closely, but the commits seem incremental. If we had wrote a bunch of code and then deleted it all, I'd favor squashing, but I don't _think_ that's this situation?

README.md
cmc333333: Where we can automate (e.g. wrapping scripts), we should, but I agree. My other dev environment is a netbook, so some of that is reflexive to me.

api/.docker/modify_path_then
toolness: Hmm, I'd argue this is a "both/and" situation rather than an "either/or" one.  For example, a lot of node developers (myself included) put `./node_modules/.bin` on their `PATH` just to make it super easy to access CLI tools that are part of a particular project--even if they are exposed via a script entry in `package.json`, there are advantages/disadvantages to using each.

IMO we should give devs the ability to use whichever approach is most comfortable or advantageous for their particular use case, and in this case I think that means we should _both_ modify the `PATH` _and_ expose more tools via `script` entries.

toolness:
Woot thanks!  I will follow this PR up with one that adds a bit more documentation about the container structure, and makes the `INTERNAL_API_URL` env var code less cryptic.

PR #985, opened 2018-02-01
Add save buttons to the editor

cmc333333:
Per convo with @rtwell and @austinhernandez, this adds a "save" button and adds a path to get to the XML editor.

## Looks like
screen shot 2018-02-01 at 12 43 37 pm


cmc333333:
Thanks @tadhg-ohiggins ! I'll rebase+merge. I'll do some more research around those mocks, too. The `as any` issue will keep coming up.

PR #986, opened 2018-02-01
Prevent list markers from being edited.

cmc333333:
As a first step towards allowing lists/markers to be added, this removes the ability to edit the list item marker.


cmc333333:
Thanks @tadhg-ohiggins ! I'll rebase/merge now, but would love additional review if you're available @toolness .

PR #987, opened 2018-02-01
Bump Python version.

cmc333333:
Our deploys have been failing today as cloud.gov no longer supports 3.6.2.


cmc333333:
Thanks @tadhg-ohiggins. I'll merge when the lights are green.

Issue #993, opened 2018-02-03
Bring back PDF upload (or remove skipped PDF upload code)


toolness:
PDF upload functionality was disabled in #956, and in #984 access to the `minio` container during development was broken.  We should eventually either re-enable this functionality, or remove the vestigial code if we no longer need it.

PR #994, opened 2018-02-03
Comment out minio for now

toolness:
Just as we're currently skipping tests related to PDF upload functionality, I think we should comment out the related Docker functionality, so developers aren't confused by its presence (and, tangentially, to make the Docker setup leaner for the time being).

Another alternative may be to simply _remove_ the code instead of commenting it out; it can always be brought back by inspecting the original commit that removed it (which we could make easy to find, e.g. by linking to it from #993).


toolness:
Woot thanks!  Looks like there was one last test that actually depended on it, but I was pretty sure the test should be skipped if PDF upload is disabled, so I have disabled it and we can re-enable it when we fix #993.

PR #1000, opened 2018-02-06
Configure CircleCI to deploy the app daily

### LINKED ISSUE #988: Replace "redeployer" with CircleCI's "schedule"
We wrote the auto-redeploy tool ("redeployer") prior to CircleCI supporting [cron jobs](https://circleci.com/docs/2.0/configuration-reference/#schedule). Let's replace it, saving a bit of our cloud.gov quota and removing a maintenance burden.

cmc333333:
Previously, we were using a small "redeployer" script which would hit CircleCI's API every day. This replaces that with a new-ish config. As a (small) benefit, we'll be able to retake the memory quota used by the redeployer app.

[See it in action](https://circleci.com/gh/18F/workflows/omb-eregs/tree/988-scheduled) (note the entries with no face!)

This should resolve #988


tadhg-ohiggins:
Approved despite the eeriness of the faceless entries.

cmc333333:
Thanks @tadhg-ohiggins ! I'll rebase and merge.

PR #1002, opened 2018-02-07
Add a bulleted list button

cmc333333:
This creates a mechanism to add a new bulleted list (or sublist) after the cursor. It doesn't allow additional bullets to be added or fully allow bullets to be removed yet, but it's a start.

## Looks like
new-bullet


api/document/js/list-utils.ts
```diff
@@ -0,0 +1,29 @@
+import { Node, ResolvedPos } from 'prosemirror-model';
+
+import schema from './schema';
+
+const defaultBullet = '●';
+const bulletFollows = {
+  '●': '○',
+  '○': '■',
+};
+
```
tadhg-ohiggins: This class is kind of hilarious.

cmc333333:
Thanks @tadhg-ohiggins!

PR #1003, opened 2018-02-08
[WIP] Support footnote and citation display in ProseMirror editor

toolness:
This is a work in progress! It's horrible!

I think I will probably save the actual editing and/or re-numbering of footnotes for a separate PR.

## To do

- [ ] Tests
- [ ] Delete a footnote if it is empty, and if no citation references it.
- [x] Ensure that user edits can't "break" the document
- [x] Implement (lossless) serialization
- [x] Add **temporary** styling (full styling that re-uses styles from the viewer should wait for another PR)

### DIFF
--- a/api/document/js/create-editor-state.ts
+++ b/api/document/js/create-editor-state.ts
@@ -7,6 +7,7 @@ import fixupDoc from './fixup-doc';
 import keyboard from './keyboard';
 import menu from './menu';
 import parseDoc from './parse-doc';
+import footnoteValidator from './footnote-validator';
 
 export default function createEditorState(data, api: Api): EditorState {
   const doc = parseDoc(data);
@@ -23,6 +24,7 @@ export default function createEditorState(data, api: Api): EditorState {
       keyboard(api),
       history(),
       fixupDoc,
+      footnoteValidator,
     ],
   });
 }
--- a/api/document/js/footnote-validator.ts
+++ b/api/document/js/footnote-validator.ts
@@ -1,0 +1,47 @@
+import { Node } from 'prosemirror-model';
+import { Plugin } from 'prosemirror-state';
+
+import schema from './schema';
+
+export function validateFootnoteCitations(doc: Node): boolean {
+  const citations = new Set<string>();
+  const footnotes = new Set<string>();
+
+  doc.descendants((node, pos, parent) => {
+    node.marks.forEach((mark) => {
+      if (mark.type === schema.marks.footnoteCitation) {
+        if (!node.text) {
+          throw new Error('Nodes with footnote citations must have text!');
+        }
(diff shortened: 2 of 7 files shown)


toolness:
@cmc333333 I tried setting `contenteditable="false"` on the footnote citations, and while it does _appear_ to work, it has weird bugs: for example, while the cursor successfully skips past the non-editable content when I use the arrow keys, _clicking_ on space within non-editable content (e.g. the space between the two digits in an `11` footnote) causes the cursor to disappear and the editor to lose focus entirely, at least in Chrome.

Aside from that, there also seem to be weird issues with changing `contenteditable` on other platforms, e.g. https://github.com/ProseMirror/prosemirror/issues/565.  So it seems like avoiding that might be prudent, if at all possible.

toolness:
Having `footnoteCitation` be a mark is proving to be increasingly problematic.  Aside from the issues mentioned above, it's not possible to create a "view" for the mark, which would be useful to e.g. provide some kind of UI affordance that allows users to jump to the actual footnote content from the citation.

Looking at ProseMirror's own [footnote example](https://prosemirror.net/examples/footnote/), it seems they decided to make it a _node_ in the `inline` group, rather than a mark.  We might want to just do this, although it will require altering out serialization/deserialization strategy, since there would no longer be a direct mapping between DB nodes => ProseMirror nodes and DB content => ProseMirror marks.

toolness:
@cmc333333 so I can think of two ways forward from here...

1. Close this PR unmerged, and create a new PR that re-envisions footnote citations as nodes instead of marks.  This will likely take a bit of work, since `parse-doc.ts` and `serialize-doc.ts` are architected to assume a mapping where DB nodes <=> ProseMirror nodes and DB content <=> ProseMirror marks.

2. Continue with my current, sub-optimal design, which will at least allow users to edit the content of footnotes, and replace it with (1) later.  This would also give us more time to think about what we want the actual _design_ of the ideal UX to be; if I barreled forth on (1) right now, I'd be creating an ad-hoc UI that might not be particularly good.

3. Put this PR on hold for now, and wait until we can talk about the design of the ideal UX.  From there we can bust forth with an implementation that suits the UX we want.

cmc333333:
I agree that editing footnote citation text does seem pretty wonky. I suspect we'll want to match Word/Docs' interface (where footnote tracking is automatic). That said, I'm generally in favor of adding functionality and then refining it when we don't have an interface designed, so I'd lean a bit towards 2. Let's get feedback from @rtwell and @austinhernandez.

In the meantime, what do you think about finishing the display part of the footnote text? If the marker's not editable, I don't think we could corrupt any existing data.

toolness:
Ok, I'm replacing this PR with #1021.

Issue #942, opened 2018-01-24
ProseMirror adding schema/display for remaining inlines: Footnote citations


yowill:

yowill:
Finishing up with https://github.com/18F/omb-eregs/issues/915

Issue #915, opened 2018-01-22
Adding schema/display for remaining content types: Footnotes


yowill:
AC:
- [ ] Edit footnote content
- [ ] Delete footnote and footnote reference
- [ ] Create footnote and footnote reference

PR #1005, opened 2018-02-08
Resolve some issues with text deletion in the editor

cmc333333:
Previously, we had kept our document schema relatively rigid -- all paragraphs
must have text, all lists: listitems, etc. This doesn't play well with
ProseMirror's internals, which need to be able to delete blocks, text, etc. at
a very fundamental level. The strict schema prevented some of these actions
and would give unexpected results (e.g. replacing a paragraph text with
"undefined") in an effort to maintain a valid document.

This takes a different approach by keeping the schema flexible, but using a
plugin to make a second pass on the modified document and performing fixups
around deleting empty pieces of content.

## Before
old-delete

## After
new-delete

### DIFF
--- a/api/document/js/fixup-doc.ts
+++ b/api/document/js/fixup-doc.ts
@@ -1,0 +1,40 @@
+import { Node } from 'prosemirror-model';
+import { EditorState, Plugin, Transaction } from 'prosemirror-state';
+
+import schema from './schema';
+
+const shouldDeleteChecks = {
+  list: (node: Node) => node.content.childCount === 0,
+  listitem: (node: Node) => node.content.childCount === 0,
+  para: (node: Node) => (
+    node.content.childCount === 1 // Just the inline
+    && node.textContent === ''
+  ),
+};
+
+export function deleteEmpty(transaction: Transaction) {
+  let newTransaction;
+  transaction.doc.descendants((node, pos) => {
+    if (newTransaction) return false;
+
+    const shouldDeleteCheck = shouldDeleteChecks[node.type.name];
+    if (shouldDeleteCheck && shouldDeleteCheck(node)) {
+      newTransaction = transaction.delete(pos, pos + node.nodeSize);
+      return false;
+    }
+    return true;
+  });
+  if (newTransaction) {
+    const subTransaction = deleteEmpty(newTransaction);
+    return subTransaction || newTransaction;
+  }
+  return newTransaction;
+}
+
+export default new Plugin({
+  appendTransaction(transactions, oldState, newState) {
+    if (oldState.doc !== newState.doc) {
+      return deleteEmpty(newState.tr);
+    }
+  },
+});
--- a/api/document/js/keyboard.ts
+++ b/api/document/js/keyboard.ts
@@ -1,4 +1,4 @@
-import { baseKeymap } from 'prosemirror-commands';
+import { baseKeymap, chainCommands, deleteSelection, selectNodeBackward, selectNodeForward } from 'prosemirror-commands';
 import { undo, redo } from 'prosemirror-history';
 import { keymap } from 'prosemirror-keymap';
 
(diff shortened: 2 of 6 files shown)


api/document/js/keyboard.ts
```diff
@@ -1,16 +1,41 @@
...
+// Removing joinBackwards/forwards from actions taken when deleting as they
+// need to be tuned a bit better to our use case.
+const backspace = chainCommands(deleteSelection, selectNodeBackward);
+const del = chainCommands(deleteSelection, selectNodeForward);
+
 export default function menu(api: Api) {
   return keymap({
     ...baseKeymap,
+    // tslint:disable-next-line object-literal-key-quotes
+    'Backspace': backspace,
```
toolness: FWIW, I quite dislike this particular `object-literal-key-quotes` tslint directive, because of how it acts in complete disregard of the context of the code it's placed in.  Particularly when we're force-disabling it through a very hard-to-remember comment directive like this, I'd rather just get rid of the rule than have to constantly override it (and honestly, even if the rule doesn't end up getting enforced in situations where it could/should be, I don't think it's that big a deal).

api/document/js/fixup-doc.ts
```diff
@@ -0,0 +1,40 @@
+import { Node } from 'prosemirror-model';
+import { EditorState, Plugin, Transaction } from 'prosemirror-state';
+
+import schema from './schema';
+
+const shouldDeleteChecks = {
+  list: (node: Node) => node.content.childCount === 0,
+  listitem: (node: Node) => node.content.childCount === 0,
+  para: (node: Node) => (
```
toolness: I wish these sorts of things could somehow be embedded as methods (alongside e.g. `toDOM()`) on the nodes' schema entries.  It'd provide a bit more context for why the schemas _seem_ to allow for empty content, and it'd also make it harder to forget to add "should delete" logic related to a node... one concern with putting this object here is that folks who add new nodes to the schema might forget to think about whether they should also make an entry for the new node here, too.

api/document/js/schema.ts
```diff
@@ -2,15 +2,15 @@ import { Node, NodeSpec, Schema } from 'prosemirror-model';
...
+    content: 'listitem*',
     group: 'block',
     toDOM: () => ['ol', { class: 'node-list' }, 0],
   },
   listitem: {
     attrs: {
       marker: { default: '1.' },
     },
-    content: 'block+',
+    content: 'block*',
```
toolness: There's a section in the [ProseMirror guide](https://prosemirror.net/docs/guide/) that says:

> It is recommended to always require at least one child node in nodes that have block content (such as "doc" and "blockquote" in the example above), because browsers will completely collapse the node when it's empty, making it rather hard to edit.

Obviously our plugin will make sure that the node is deleted instead of being an empty node that the user can't see; however, my general concern is that I'm not sure if this notion of having a less stringent schema in exchange for having plugins that "fix up" the doc is something that is "doing things the ProseMirror way" or "fighting against ProseMirror".  I'm too unfamiliar with ProseMirror commands/transactions/steps in particular to be able to even answer that question, alas.

api/document/js/keyboard.ts
cmc333333: Sure, we can turn it off.

api/document/js/fixup-doc.ts
cmc333333: What do you think about creating a wrapper around the ProseMirror Node schema to collect all of these types of methods? Then we'd ensure we had factories, serializers, etc. in one place.

api/document/js/schema.ts
cmc333333: Agreed. I get the sense that ProseMirror pushes for a strict schema with tight control over all the transforms, but thus far, that's led to a very fragile document. In that gif, for example, select-all-ing magically replaces a node with "undefined" (presumably because the underlying data model was corrupted). I think we could push this strategy further, but we'd likely need to replace many of the core transforms (including simple things, like deletion) with our own.

Taking this fix-it-later approach is definitely not my first choice, but it seems to work reasonably well. I suspect we'll find better alternatives as we progress.

Two downstream projects might be worth investigating: [Dovetail](https://github.com/heydovetail/editor/blob/4437d450a8a5164e488f8c9e0f5b8cd37bd9a442/src/schema.ts), [pubpub](https://github.com/pubpub/pubpub-packages/blob/2445714e87f25580b9c4e2c24bda837e14828cd5/packages/pubpub-editor/src/prosemirror-setup/schema/schemaDefinition.js)

toolness:
Cool! Um, I hope you don't mind if I merge this now, as it is going to be very helpful in my implementation of #942 (footnote citations) and #915 (footnotes), which I am working on now.

PR #1006, opened 2018-02-09
Add new li when pressing enter at end of existing li

cmc333333:
If the cursor's at the end of the text of an existing li (and there are no child nodes, like footnotes), pressing Enter will create a new li immediately following the current, renumber the lis, and place the cursor in the new one.

Renumbering inspects the first li and uses that as a template for all following (meaning we can have arbitrary text in the marker). It'd be a good idea to call the renumbering operating when deleting lis and changing indents, but this doesn't implement those scenarios.

## Looks Like
add-bullet
add-ordered
renumbering

### DIFF
--- a/api/document/js/__tests__/commands.test.ts
+++ b/api/document/js/__tests__/commands.test.ts
@@ -2,19 +2,22 @@ jest.mock('../Api');
 jest.mock('../serialize-doc');
 window.location.assign = jest.fn();
 
+import { Node } from 'prosemirror-model';
 import { EditorState, TextSelection } from 'prosemirror-state';
 
 import { JsonApi } from '../Api';
 import {
+  addListItem,
   appendBulletListNear,
   appendNearBlock,
   appendParagraphNear,
   makeSave,
   makeSaveThenXml,
 } from '../commands';
+import { collectMarkers } from '../list-utils';
 import schema, { factory } from '../schema';
 import serializeDoc from '../serialize-doc';
-import pathToResolvedPos, { NthType } from '../path-to-resolved-pos';
+import pathToResolvedPos, { NthType, SelectionPath } from '../path-to-resolved-pos';
 
 function executeTransform(initialState: EditorState, transform): EditorState {
   const dispatch = jest.fn();
@@ -191,5 +194,55 @@ describe('makeSaveThenXml()', () => {
 
     const param = locationAssign.mock.calls[0][0];
     expect(param).toMatch(/akn$/);
+  });
+});
+
+describe('addListItem()', () => {
+  const doc = factory.policy([
+    factory.para('intro'),
+    factory.list([
+      factory.listitem('a:', [factory.para('aaa')]),
+      factory.listitem('b:', [
+        factory.para('bbb first'),
+        factory.para('bbb second'),
+      ]),
+    ]),
+  ]);
+
+  function makeState(path: SelectionPath) {
+    const selection = new TextSelection(pathToResolvedPos(doc, path));
+    return EditorState.create({ doc, selection });
+  }
+
+  it('requires the cursor be in the right position', () => {
+    const inIntro = makeState(['para', 'inline', 'int'.length]);
+    const endIntro = makeState(['para', 'inline', 'intro'.length]);
+    const middleOfLi = makeState(['list', 'listitem', 'para', 'inline', 'a'.length]);
+    const endOfFirstPara = makeState(
+      ['list', new NthType(1, 'listitem'), 'para', 'inline', 'bbb first'.length]);
+    [inIntro, endIntro, middleOfLi, endOfFirstPara].forEach(state =>
+      expect(addListItem(state)).toBe(false));
+  });
+
+  it('adds a new li', () => {
+    const init = makeState(['list', 'listitem', 'para', 'inline', 'aaa'.length]);
+    expect(doc.content.child(1).content.childCount).toBe(2);
+    const result = executeTransform(init, addListItem);
+    const list = result.doc.content.child(1);
+    expect(list.content.childCount).toBe(3);
+    expect(list.content.child(1).textContent).toBe(' ');
+  });
+
+  it('puts the cursor in the li', () => {
+    const init = makeState(['list', 'listitem', 'para', 'inline', 'aaa'.length]);
+    const result = executeTransform(init, addListItem);
+    expect(result.selection.anchor).toBe(pathToResolvedPos(
+      result.doc,
+      ['list', new NthType(1, 'listitem'), 'para', 'inline'],
+    ).pos);
+    expect(result.selection.head).toBe(pathToResolvedPos(
+      result.doc,
+      ['list', new NthType(1, 'listitem'), 'para', 'inline', ' '.length],
+    ).pos);
   });
 });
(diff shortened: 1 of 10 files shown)


api/tslint.yaml
```diff
@@ -7,3 +7,5 @@ rules:
   # Deprecated in favor of no-this-assignment
   no-var-self: false
   no-this-assignment: true
+  # Removing this as we don't want to have to mix str and literal keys
+  object-literal-key-quotes: false
```
toolness: Ah nice, thanks for adding this! 🎉

api/package.json
```diff
@@ -43,6 +43,8 @@
     "prosemirror-menu": "^1.0.0",
     "prosemirror-state": "^1.0.0",
     "prosemirror-view": "^1.0.5",
+    "repeat-string": "^1.6.1",
+    "romanize": "^1.1.1",
```
toolness: Sometimes I wish JSON files supported comments, so we could add a little bit of context briefly explaining what a module does and why we need it.

Also, `repeat-string` makes me wish JS supported `"some string" * 5` like Python does.

api/document/js/list-utils.ts
```diff
@@ -1,29 +1,83 @@
...
   }
   return defaultBullet;
 }
+
+function makeIntToLetter(initial: string) {
+  const offset = initial.charCodeAt(0);
+
+  return (idx: number) => {
+    const count = Math.floor(idx / 26) + 1;
+    const letter = String.fromCharCode(offset + idx % 26);
```
toolness: I think the meaning of `26` is fairly obvious from context here, though ordinarily I'd suggest making it a named constant like `NUM_LETTERS_IN_ALPHABET`.  Not sure if this would actually be helpful in this case or not.

api/document/js/commands.ts
```diff
@@ -69,3 +70,46 @@ export function makeSaveThenXml(api: Api) {
...
   };
 }
+
+const inLi = (pos: ResolvedPos) => (
+  pos.depth >= 3
+  && pos.node(pos.depth).type === schema.nodes.inline
+  && pos.node(pos.depth - 1).type === schema.nodes.para
+  && pos.node(pos.depth - 2).type === schema.nodes.listitem
+  && pos.node(pos.depth - 3).type === schema.nodes.list
+);
```
toolness: This makes me wish we had a sort of "CSS selectors, but for ProseMirror document schema".

api/document/js/commands.ts
```diff
@@ -69,3 +70,46 @@ export function makeSaveThenXml(api: Api) {
...
+const atEndOfLi = (pos: ResolvedPos) => (
+  pos.depth >= 2
+  && pos.pos === pos.end(pos.depth)
+  && pos.pos === pos.end(pos.depth - 1) - 1
+  && pos.pos === pos.end(pos.depth - 2) - 2
+);
+
+export function addListItem(state, dispatch?) {
```
toolness: It might be useful, both for documentation and type-checking goodness, to annotate `state` as an [`EditorState`](https://prosemirror.net/docs/ref/#state.EditorState) here (and in other function declarations that take `state`).

It _might_ also be useful to define `dispatch` as `((tr: Transaction) => void) | null` (or make a type alias, since lots of functions take it) but up to you really.

api/document/js/commands.ts
cmc333333: Agreed. The `pathToResolvedPos()` function implements some of that, but it's pretty rudimentary.

api/document/js/commands.ts
cmc333333: Good call, can do.

api/document/js/__tests__/commands.test.ts
```diff
@@ -187,9 +191,59 @@ describe('makeSaveThenXml()', () => {
...
+  }
+
+  it('requires the cursor be in the right position', () => {
+    const inIntro = makeState(['para', 'inline', 'int'.length]);
+    const endIntro = makeState(['para', 'inline', 'intro'.length]);
+    const middleOfLi = makeState(['list', 'listitem', 'para', 'inline', 'a'.length]);
+    const endOfFirstPara = makeState(
+      ['list', new NthType(1, 'listitem'), 'para', 'inline', 'bbb first'.length]);
+    [inIntro, endIntro, middleOfLi, endOfFirstPara].forEach(state =>
+      expect(addListItem(state)).toBe(false));
```
toolness: This reminded me of py.test's concept of "parameterized tests" and after a bit of googling it turned up [`jest-each`](https://www.npmjs.com/package/jest-each).  If we think we'll be writing tests like this often, it might be worth adding later, outside of this PR.

api/document/js/__tests__/commands.test.ts
```diff
@@ -187,9 +191,59 @@ describe('makeSaveThenXml()', () => {
...
+    const init = makeState(['list', 'listitem', 'para', 'inline', 'aaa'.length]);
+    const result = executeTransform(init, addListItem);
+    expect(result.selection.anchor).toBe(pathToResolvedPos(
+      result.doc,
+      ['list', new NthType(1, 'listitem'), 'para', 'inline'],
+    ).pos);
+    expect(result.selection.head).toBe(pathToResolvedPos(
+      result.doc,
+      ['list', new NthType(1, 'listitem'), 'para', 'inline', ' '.length],
+    ).pos);
```
toolness: I wonder if it'd be more readable to write this as something like:

```js
const anchorPath = ['list', new NthType(1, 'listitem'), 'para', 'inline'];
expect(result.selection.anchor)
  .toBe(pathToResolvedPos(result.doc, anchorPath).pos);
expect(result.selection.head)
  .toBe(pathToResolvedPos(result.doc, [...anchorPath, ' '.length]).pos);
```

I dunno.

api/document/js/list-utils.ts
cmc333333: I'll play with it a bit to see if there's something helpful and concise. I'll also add a test around the boundary.

api/document/js/commands.ts
```diff
@@ -69,3 +70,46 @@ export function makeSaveThenXml(api: Api) {
...
+  const pos = state.selection.$head;
+  if (!inLi(pos) || !atEndOfLi(pos)) {
+    return false;
+  }
+  if (!dispatch) {
+    return true;
+  }
+
+  const endOfLi: number = pos.end(pos.depth - 2);
```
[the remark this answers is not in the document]
cmc333333: I'll take another stab with more comments and perhaps more consts. I agree that ProseMirror's `ResolvedPos` isn't super descriptive with its method names. We may want to extend it.

api/document/js/__tests__/commands.test.ts
```diff
@@ -2,19 +2,22 @@ jest.mock('../Api');
...
   appendNearBlock,
   appendParagraphNear,
   makeSave,
   makeSaveThenXml,
 } from '../commands';
+import { collectMarkers } from '../list-utils';
 import schema, { factory } from '../schema';
 import serializeDoc from '../serialize-doc';
-import pathToResolvedPos, { NthType } from '../path-to-resolved-pos';
+import pathToResolvedPos, { NthType, SelectionPath } from '../path-to-resolved-pos';
```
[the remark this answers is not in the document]
cmc333333: :+1: I'll add a follow up PR for that

api/document/js/__tests__/commands.test.ts
cmc333333: Good idea; I'll open a ticket.

api/document/js/__tests__/commands.test.ts
cmc333333: I agree, will implement.

cmc333333:
Thanks @toolness! Updated to address; I'll now rebase and merge.

PR #1009, opened 2018-02-09
Set workflow phase to 'published' in import_xml_doc command

toolness:
The current README instructions that instruct a user to use `manage.py import_xml_doc` to import an existing XML doc are a bit developer-unfriendly after #842 because the related policy's `workflow_phase` is still set to `no_doc`, which means that even though the document is stored in the database, it won't be linked-to in the viewer anymore.

This just fixes it so that the related policy's `workflow_phase` is set to `published`, which restores the previous developer experience (and also makes sense semantically).

### DIFF
--- a/api/document/management/commands/import_xml_doc.py
+++ b/api/document/management/commands/import_xml_doc.py
@@ -2,6 +2,7 @@ import argparse
 import logging
 
 from django.core.management.base import BaseCommand
+from django.db import transaction
 
 from document.parsers import AkomaNtosoParser
 from document.serializers.doc_cursor import DocCursorSerializer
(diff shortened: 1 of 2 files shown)


api/document/management/commands/import_xml_doc.py
```diff
@@ -18,6 +18,8 @@ def fetch_policy(identifier: str):
 
 
 def import_xml_doc(policy, xmlstream):
+    policy.workflow_phase = WorkflowPhases.published.name
+    policy.save()
```
cmc333333: Do you think should should happen after the `serializer.save()` step? Alternatively, we could wrap the whole thing in a transaction.

toolness: Good idea, I just wrapped it all in a transaction!

PR #1011, opened 2018-02-12
Improve API typings

toolness:
When I was working on #1003, I realized that I was frequently getting confused about what kind of object was a "Django app document node" vs. what kind of object was a ProseMirror node.  At first I started creating my own typing for the former, until I noticed that it already existed in `serialize-doc.ts`!  So I just moved the related typings from there to `Api.ts`, since that seemed like a more appropriate location for types called `ApiNode` and `ApiContent`.

I also changed `Api` to be parameterized based on whether the caller wants to use it for JSON or AKN+XML, since that alters the return type of the object.

### DIFF
--- a/api/document/js/Api.ts
+++ b/api/document/js/Api.ts
@@ -35,43 +35,32 @@ export interface ApiContent {
   text: string;
 }
 
-interface ApiTypeMap {
-  'json': ApiNode;
-  'akn+xml': string;
-}
-
-interface ApiOptions<T extends keyof ApiTypeMap> {
-  contentType: T;
-  csrfToken: string;
-  url: string;
-}
-
-export default class Api<T extends keyof ApiTypeMap> {
+export class Api<T> {
   url: string;
   contentType: string;
   csrfToken: string;
 
-  constructor(options: ApiOptions<T>) {
-    this.url = options.url;
-    this.contentType = `application/${options.contentType}`;
-    this.csrfToken = options.csrfToken;
+  constructor({ contentType, csrfToken, url }) {
+    this.url = url;
+    this.contentType = contentType;
+    this.csrfToken = csrfToken;
   }
 
-  async fetch(): Promise<ApiTypeMap[T]> {
+  async fetch(): Promise<T> {
     try {
       setStatus('Loading document...');
       const response = await axios.get(this.url, { headers: {
         Accept: this.contentType,
       } });
       setStatus('Document loaded.');
-      return response.data as ApiTypeMap[T];
+      return response.data as T;
     } catch (e) {
       setStatusError(e);
       throw e;
     }
   }
 
-  async write(data: ApiTypeMap[T]): Promise<void> {
+  async write(data: T): Promise<void> {
     try {
       setStatus('Saving...');
       await axios.put(this.url, data, { headers: {
@@ -83,5 +72,17 @@ export default class Api<T extends keyof ApiTypeMap> {
       setStatusError(e);
       throw e;
     }
+  }
+}
+
+export class JsonApi extends Api<ApiNode> {
+  constructor({ csrfToken, url }) {
+    super({ csrfToken, url, contentType: 'application/json' });
+  }  
+}
+
+export class AknXmlApi extends Api<string> {
+  constructor({ csrfToken, url }) {
+    super({ csrfToken, url, contentType: 'application/akn+xml' });
   }
 }
--- a/api/document/js/Api.ts
+++ b/api/document/js/Api.ts
@@ -18,31 +18,60 @@ export function setStatusError(e: Error) {
   setStatus(errMsg, 'editor-status-error');
 }
 
-export default class Api {
+export interface ApiNode {
+  children: ApiNode[];
+  content: ApiContent[];
+  marker?: string;
+  node_type: string;
+  type_emblem?: string;
+  title?: string;
+  identifier?: string;
+  text?: string;
+}
+
+export interface ApiContent {
+  content_type: string;
+  inlines: ApiContent[];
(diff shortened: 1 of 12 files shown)


api/document/js/Api.ts
```diff
@@ -18,31 +18,60 @@ export function setStatusError(e: Error) {
...
+export interface ApiNode {
+  children: ApiNode[];
+  content: ApiContent[];
+  marker?: string;
+  node_type: string;
+  type_emblem?: string;
+  title?: string;
+  identifier?: string;
+  text?: string;
+}
```
toolness: One minor concern I have is that there are differences in the shape of the objects the API returns vs. what the PUT endpoint expects.  For example, every `ApiNode` returned by the `GET` endpoint always has an `identifier`, but because identifiers are auto-assigned by the `PUT` endpoint, it's not required when passing the document to that endpoint.  However, this results in requiring a bunch of unnecessary assertions in code that handles the `GET` endpoint, in order to appease TypeScript.

We might be able to separate this into two different interfaces, one of which is a "stronger" version of the other, perhaps via interface inheritance or some other means.

api/document/js/Api.ts
```diff
@@ -18,31 +18,60 @@ export function setStatusError(e: Error) {
...
+interface ApiTypeMap {
+  'json': ApiNode;
+  'akn+xml': string;
+}
+
+interface ApiOptions<T extends keyof ApiTypeMap> {
+  contentType: T;
+  csrfToken: string;
+  url: string;
+}
```
toolness: IIRC, typed object destructuring isn't supported very well in TypeScript, so I've just split `ApiOptions` out into its own separate interface for now.  Not ideal but I'm not sure how else to do this.

api/document/js/serialize-doc.ts
```diff
@@ -41,8 +27,11 @@ export const apiFactory = {
   }),
 };
 
+type NodeConverterMap = {
+  [nodeName: string]: (node: Node) => ApiNode,
```
toolness: In the future, I'd like to improve upon this to actually require that `nodeName` be a valid node type in our schema.  That said, I might be venturing into "type astronaut" territory with that one...

api/document/js/parse-doc.ts
```diff
@@ -24,8 +25,16 @@ export function convertContent(content, marks: Mark[]): Node[] {
   return flatten(nestedChildNodes);
 }
 
-const NODE_TYPE_CONVERTERS = {
+
+type NodeConverterMap = {
+  [nodeName: string]: (node: ApiNode) => Node,
+};
```
toolness: In the future, I'd like to improve upon this to actually require that `nodeName` be a valid API node type.

api/document/js/Api.ts
cmc333333: Excellent point; I think having two interfaces might help us telegraph the intent of the objects we're creating, too. This one'll be used when you're loading data, this one when you're saving.

api/document/js/Api.ts
```diff
@@ -18,31 +18,60 @@ export function setStatusError(e: Error) {
...
+
+export default class Api<T extends keyof ApiTypeMap> {
```
cmc333333: What do you think about making this type signature less abstract? Right now it's `Api<'json'>` or `Api<'akn+xml'>`; would it be easier to understand as `Api<ApiNode>` and `Api<string>`? It might even make sense to lock the type var in by having `NodeApi` and `StringApi` as subclasses of an abstract `Api<T>`.

api/document/js/__tests__/Api.test.ts
```diff
@@ -51,7 +53,7 @@ describe('fetch()', () => {
     error.response = { data: { oh: 'noes', and: 6 } };
     axios.get = jest.fn(() => { throw error; });
 
-    await api.fetch();
+    try { await api.fetch(); } catch (_) {}
```
cmc333333: Since we're raising this exception now, should we explicitly check it? Something like
```
expect(async () => await api.fetch()).toThrow(error);
```

api/document/js/parse-doc.ts
```diff
@@ -24,8 +25,16 @@ export function convertContent(content, marks: Mark[]): Node[] {
...
+
+const NODE_TYPE_CONVERTERS: NodeConverterMap = {
   heading(node) {
+    if (!node.identifier)
+      throw new Error('Assertion failure, node needs identifier');
```
cmc333333: Nice catch; yay static types.

api/document/js/serialize-doc.ts
cmc333333: I think that meshes with your earlier concerns around logic living in lots of different places. I can imagine we'd have an interface like
```typescript
interface SchemaNode {
  ...
  fromApi: (node: ApiNode) => ProseMirror.Node?,
  toApi: () => ApiNode,
  deleteWhenEmpty: boolean,
  ...
}
```
at some point.

api/document/js/Api.ts
toolness: Yeah, that is what I originally wanted, but I wasn't sure how best to map the type into the `content-type` header that ultimately got passed to the server--I didn't want, for example, someone to be able to make an `Api<ApiNode>` that passed a `content-type` of `akn+xml`--so I looked up the type definition for `document.createElement()` for inspiration.  But yeah, it's not ideal...  I think the solution that combines the best of readability + flexibility might be your latter solution there.

api/document/js/__tests__/Api.test.ts
toolness: Ah good idea, I will bust it up.

PR #1013, opened 2018-02-12
Fix stylesheet for lists.

cmc333333:
ProseMirror's stylesheets alter how whitespace is collapsed by browsers.
Unfortunately, our .clearfix class (based on basscss) assumed whitespace
_would_ be collapsed.

This doesn't affect the viewer, but for the editor:

## Before
screen shot 2018-02-12 at 6 37 32 pm

## After
screen shot 2018-02-12 at 6 37 56 pm


cmc333333:
Thanks @xtine and @toolness

PR #1016, opened 2018-02-13
Button to add ordered lists; renumber when deleting

cmc333333:
This extends #1006.

It takes a different tact to handling list renumbering by tracking the marker template with the list node (rather than re-deriving from the first li). This ends up being necessary due to users potentially deleting that first li. Aside from that refactor, this adds a button and mimics much of the unordered-list-adding functions.


cmc333333:
Thanks, @tadhg-ohiggins. I'll get #1006 merged then rebase and merge this one.

PR #1017, opened 2018-02-13
Start of the Style Guide

xtine:
I've set up a Fractal instance with some initial documentation and component pages to start the style guide off with.
- Dockerized to localhost:3000 and i've also included a npm shortcut to save dependencies into (had to do that for installing `copy-webpack-plugin` for example). 
- Takes in the existing stylesheets (with some custom style guide only classes) and copies images for its own separate instance.

screen shot 2018-02-13 at 3 59 58 pm
screen shot 2018-02-13 at 4 00 07 pm
screen shot 2018-02-13 at 4 00 15 pm


cmc333333:
I don't understand all of the templates, but this seems like a good start to me. A few things we should probably resolve before merging:
* We should pull in master -- it'll include the updated CircleCI builds that are blocking this now
* We can delete several of the copied scripts from `.docker` -- we only need `deps_ok_then` `modify_path_then` and `build_run_styleguide`
* We can ignore the duplication warnings from CodeClimate, but I think it should be pretty straightforward to address most of the other issues; they're arising from our [scss-lint config](https://github.com/18F/omb-eregs/blob/a06c4029d25184b3c74c7218d3910d4be84e4112/.scss-lint.yml).

xtine:
@cmc333333: code climate isn't happy that the webpack config for styleguide is too similar to the ui. Can we set it to ignore this particular issue or is there another way to resolve this?

Issue #1010, opened 2018-02-12
As a policy analyst or contractor, I need to be able to identify preamble and metadata content in my document 


austinhernandez:
## assumptions 
- [ ] The user has uploaded a new document or is editing an existing one and must identify areas in the text where there is preamble and meta data content.

## actions 
- [ ] Create wireframes high or low fidelity and/or a task flow for how a user will select and identify preamble and metadata text in the WYSIWYG editor
- [ ] Document necessary warning messages and or error messages
- [ ] Make sure all paragraph and text styles are accounted for in the mockups
- [ ] Metadata and preamble text is finalized and accounted for in the mockups
- [ ] Validate interaction(s) with teams

## Suggestions, things to consider, notes, etc.
- [Most recent Sketch document is here on Google Drive](https://drive.google.com/open?id=1yf9ZVDOtE9RMIQf2ybC5JRW9t9S5d7a8)
- Component library is shared.
- I was using the [scrivener](https://www.literatureandlatte.com/scrivener/overview), Google docs, and Microsoft word as foundations for interaction patterns.

## Current mockups

![image](https://user-images.githubusercontent.com/23194720/36111135-1b433f8e-0ff3-11e8-8e31-d0688b2aa594.png)

![wysiwyg interactions](https://user-images.githubusercontent.com/12564977/36123401-18bd8d36-1012-11e8-979b-555b2891f654.png)

austinhernandez:
tbd in later sprint, design styles for preamble and meta data

rtwell:
![preamble 1](https://user-images.githubusercontent.com/12564977/36283316-c2ccd35a-1269-11e8-9a54-6b2f4fae50b3.png)
![preamble 2](https://user-images.githubusercontent.com/12564977/36283317-c2e6c1c0-1269-11e8-9342-36a3b0d5c076.png)
![preamble 3](https://user-images.githubusercontent.com/12564977/36283318-c304553c-1269-11e8-98d9-0b2a7ca7b5fe.png)
![preamble 4](https://user-images.githubusercontent.com/12564977/36283319-c3316d60-1269-11e8-916e-cb9dd34252e1.png)
![preamble 5](https://user-images.githubusercontent.com/12564977/36283320-c352b1e6-1269-11e8-8dea-4582a664dc84.png)

rtwell:
## the issuance date will need to be validated in two ways:

### trigger 1: document save
[...]

PR #1018, opened 2018-02-14
[WIP EXPERIMENT] Alternative approach to WYSIWYG footnote editing

toolness:
This takes an alternative approach to footnote editing by assuming the API is providing the editor with a different document structure than it actually (currently) provides it with.  It also assumes the API can take in this different document structure in PUT requests.

For now we'll call this different document structure the "inlined footnotes structure", because that's what it does: it just inlines footnotes at their citation.

In reality, a "munging" layer exists between the API and the rest of the code that modifies the document structure as needed.

The idea is that we'd eventually modify the actual API (and the viewer) to accept this alternative document structure; as discussed in Slack and proposed by @thebestsophist, this is likely what we'll need to do anyways in order to conform to AKN, and it seems to make sense regardless since there is a one-to-one correspondence between footnotes and their citations. I've implemented the munging layer for now instead of doing things "properly" partly because it will likely be a _lot_ of work to modify the rest of our architecture.

However, I'm also taking the munging approach because regardless of the rest of our architecture, the inlined footnotes structure is _much_ easier for us to deal with at the ProseMirror level, particularly given Austin's mockups (see #1010), which represent footnotes inline.

### DIFF
--- a/api/document/js/Api.ts
+++ b/api/document/js/Api.ts
@@ -2,6 +2,7 @@ import axios from 'axios';
 
 import { getEl } from './util';
 import { makeErrorFriendly } from './friendly-error';
+import * as footnoteMunging from './footnote-munging';
 
 function setStatus(msg: string, className: 'editor-status-error'|'' = '') {
   const status = getEl('#status');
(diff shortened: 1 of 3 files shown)


toolness:
Oh fiddlesticks, this isn't working, because of the way that ProseMirror's marks work.  I've got a different idea, though... closing this for now.

PR #1019, opened 2018-02-14
Proposal for an alternate schema organization

cmc333333:
Currently, we have our ProseMirror schema defined in one file and we place
logic for transforming elements of that schema in separate files (generally
with a object keyed by schema type name). As @toolness' pointed out, this
makes tracking all of the things that need to change when we add new
functionality rather difficult.

This proposes a different tract, defining a single, extend Node schema (called
"DocNodeSpec") that all of our node types would have to implement. The type
system would inform us when we needed to update a DocNodeSpec to add new
functionality, ensuring we don't forget any cases. It should also spell out
all of the functionality we expect from our nodes.

One annoyance with this approach is recursive imports -- the schema will
depend on the DocNodeSpec but the DocNodeSpec will need the schema to create
new nodes. We run into similar issues when handling recursive
serialization/deserialization. We already use recursive imports elsewhere, so
I'm not super worried, but it's worth pointing out.

As another implementation detail, this uses composition (with overrides)
rather than OO-inheritance. We'll only ever define a schema once, so this
seems like a reasonable solution.

Anyway, throwing this up for discussion. Is this easier to understand and less
error prone?


toolness:
Ah, I like this!  It also seems like it _might_ lend itself better to type-checking/intellisense.  Right now when I type `schema.nodes.` into my editor, for instance, it doesn't offer any suggestions, i think because it just knows that `schema` is a generic PM schema of some sort, but not that it's a _specific_ schema.  Your refactoring seems to indicate that kind of thing could perform better, mayhaps.

Anyways, I dig it. 👍

PR #1021, opened 2018-02-14
Support inline footnotes in WYSIWYG editor

toolness:
This implements footnotes in the WYSIWYG editor to be inline, as per Austin's mockup in #1010:

> screenshot-2018-2-14 document editor

This supersedes #1003 and #1018.

## Implementation

We accomplish this by taking advantage of the `footnote_node` that comes embedded in `footnote_citation` content from the API, and simply ignoring the actual `footnote` nodes that appear in the document.

When writing back to the API, we re-insert the `footnote` nodes by adding them as the first child of the parent node that contains the inline footnotes.  _However_, it should be noted that it might be appropriate for this to be done on the server-side.  It should also be noted that the document structure sent in a PUT might not _exactly_ resemble that of the original GET, since `footnote` nodes may technically be in different places (but this shouldn't ultimately affect the WYSIWYG editing or viewing experience in any way, though it will affect raw AKN XML editing).

## To do

- [x] Add tests.
- [x] Consider renaming the `inliney` group to something else. (`inline` is taken by the node type, and renaming that will require lots of changes to other code, which we can certainly do--I just didn't want to expend the effort before getting other folks' opinions first.)
- [x] Style the CSS to look more like Austin's mockups (specifically, use the pink color and sans-serif font).
- [x] Ensure that the `marker` of the generated `footnote` nodes is properly populated.


toolness:
@cmc333333 does my approach here seem OK to you?  And if so, do you think we should name `inliney` something else, or rename it to `inline` and rename the existing `inline` to something else? Or deal with that in a separate PR?

Anyways, if this seems like a decent approach I will start busting up some tests.

api/document/js/serialize-doc.ts
```diff
@@ -99,10 +115,26 @@ export function nestMarks(text: string, marks: Mark[]): ApiContent {
   return converted;
 }
 
+function convertInlineFootnote(node: Node): ApiContent {
+  const emblem = node.attrs.emblem;
+  return apiFactory.content('footnote_citation', {
+    inlines: [apiFactory.text(emblem)],
+    footnote_node: apiFactory.node('footnote', {
```
cmc333333: This field isn't writable, so we can probably skip it here for clarity.

cmc333333:
This is a really creative solution, @toolness, nice work. It sounds like we're on the same page that we should migrate marks over to nodes eventually, in which case we'd probably want a group of "inlines" and a group of "blocks". I think that means we should rename `inliney` to `inline` and `inline` to something else (maybe `content` to match AKN, or `paragraphText` to match our CSS classes). I thiiink that'd be a pretty quick search-and-replace, if we can get it done here.

toolness:
Woooooo thanks!  Ok I will rename stuff.  I am thinking `paragraphText` might be better than `content`, simply because [`content` has its own meaning within ProseMirror](https://prosemirror.net/docs/ref/#model.Node.content) and I don't want to cause any confusion.

api/document/js/serialize-doc.ts
toolness: Oh, so I'm actually "stashing" it there so I can retrieve it later in `extractFootnotes`.  I actually did this for a few reasons:

1. Maybe this field _should_ be writable?  It's clearly easier for this client code to write to it, so maybe it's just a better API.  In other words, maybe `extractFootnotes` should really be on the server-side. If we end up doing this, fixing this code would just involve getting rid of `extractFootnotes`, rather than re-architecting anything.

2. Because the functions in `serialize-doc.ts` return `ApiContent`, there's otherwise no easy way to return an `ApiNode` along with the content, so it's convenient to "stash" the footnote there. An alternative would be to modify the module to use functions that returned some sort of structure that included both types of content, e.g. the [`Uninlineds` structure from PR 1018](https://github.com/18F/omb-eregs/pull/1018/files#diff-7dbfa13a162e6c340db15585111acc4bR59).

cmc333333: Ah, cool, I did not follow that this was being used again on first read-through. I suspect we'll change the format pretty significantly to match AKN: removing distinct footnote nodes and instead including all their text in the paragraph, along with a footnote _annotation_. In that case, the API'd be something like
```js
{
  "content_type": "footnote",
  "inlines": [{"content_type": "__text__", "text": "some content here"}],
  "number": "12",
}
```

toolness:
Ok, I think this is finally good to go!

api/document/js/__tests__/serialize-doc.test.ts
```diff
@@ -53,6 +53,20 @@ describe('serializeDoc()', () => {
...
+      schema.nodes.paraText.create({}, [
+        schema.text('hello'),
+        schema.nodes.inlineFootnote.create({ emblem: '2' }, [
+          schema.text('i am a footnote'),
+        ]),
+      ]),
+    ]);
+
+    const result = serializeDoc(node);
+    expect(result).toMatchSnapshot();
```
toolness: There were so many different things to test here that I thought it'd be both easier to write and read to simply make it a snapshot test.

api/document/js/path-to-resolved-pos.ts
```diff
@@ -9,6 +9,9 @@ export class NthType {
   constructor(nth: number, nodeType: string | NodeType) {
     this.nth = nth;
     this.nodeType = nodeType instanceof NodeType ? nodeType : schema.nodes[nodeType];
+    if (!this.nodeType) {
+      throw new Error(`"${nodeType}" is not a valid node in the schema`);
```
toolness: This isn't a necessary change for this PR, but I figured it'd be more useful feedback than something more cryptic down the line, when something else tried accessing an attribute of `nodeType`.

api/ereqs_admin/static/admin/module/_editor.scss
```diff
@@ -6,6 +6,17 @@
...
 
+  .inline-footnote {
+    color: $color-pink;
+    @include font-san-serif;
+  }
+
+  .inline-footnote::before {
+    content: 'Footnote ' attr(data-emblem) ': ';
+    padding-left: 0.5em;
+    @include font-san-serif-bold;
```
toolness: Uh, is there a particular reason we're calling the font `san-serif` instead of `sans-serif`?

api/document/js/__tests__/schema.test.ts
```diff
@@ -21,6 +21,14 @@ describe('unimplementedNode', () => {
   });
 });
 
+describe('inlineFootnote', () => {
+  it('is cool', () => {
```
cmc333333: How about 'serializes to include a data-emblem attr'?

api/document/js/__tests__/serialize-doc.test.ts
```diff
@@ -53,6 +53,20 @@ describe('serializeDoc()', () => {
...
 
+  it('converts footnotes', () => {
+    const node = schema.nodes.para.create({}, [
+      schema.nodes.paraText.create({}, [
+        schema.text('hello'),
+        schema.nodes.inlineFootnote.create({ emblem: '2' }, [
+          schema.text('i am a footnote'),
+        ]),
+      ]),
+    ]);
```
cmc333333: If the need to include inlineFootnotes comes up frequently in our tests, we may want to modify the `para` factory to accommodate.

api/document/js/__tests__/serialize-doc.test.ts
cmc333333: I generally dislike snapshots as they're difficult to read: not only are they in a separate file, they include _everything_, making it difficult to highlight the important pieces. We may need to add unit tests around this in the future.

api/document/js/parse-doc.ts
```diff
@@ -11,11 +11,29 @@ export default function parseDoc(node): Node {
...
+    throw new Error('Assertion failure, footnote citation needs ' +
+                    'footnote node');
+  }
+  const emblem = footnote.type_emblem;
+  if (!emblem) {
+    throw new Error('Assertion failure, footnote needs emblem');
+  }
+  const nested: Node[][] = (footnote.content || [])
+    .map(c => convertContent(c, []));
+  return schema.nodes.inlineFootnote.create({ emblem }, flatten(nested));
```
cmc333333: Should we add a factory for this to ensure the `emblem` is always present?

api/document/js/parse-doc.ts
```diff
@@ -25,6 +43,13 @@ export function convertContent(content, marks: Mark[]): Node[] {
   return flatten(nestedChildNodes);
 }
 
+function mapChildren(node: ApiNode): Node[] {
+  return (node.children || []).filter((child) => {
+    // We're going to strip out footnotes, as we're using the
+    // footnote node embedded in the citation instead.
+    return (child.node_type !== 'footnote');
```
cmc333333: This raises a good point -- we're potentially losing data if the footnote isn't referenced anywhere, which we know is a frequent occurrence from out pdf parser's output. I think we'll need a way to catch those footnotes -- what do you think about keeping
```js
policy: node => factory.policy((node.children || []).map(parseDoc))
```
(and adding commentary explaining what's going on)? I'm pretty sure we put uncited footnotes as children of the root `policy`, so I think that'd remove the risk of lost data.

api/document/js/schema.ts
```diff
@@ -34,10 +34,22 @@ const schema = new Schema({
...
     },
-    inline: {
-      content: 'text*',
+    paraText: {
+      content: 'inline*',
       toDOM: () => ['p', { class: 'node-paragraph-text' }, 0],
     },
+    inlineFootnote: {
+      attrs: {
+        emblem: {},
```
cmc333333: Should we give this a default? May reduce the number of confusing errors down the line...

api/ereqs_admin/static/admin/module/_editor.scss
cmc333333: Good question; we can probably swap them all to sans. Let's check in with design real quick to confirm we want to change the font face in the middle of a paragraph, though.

api/document/js/__tests__/schema.test.ts
toolness: LOL, my bad; I always start tests by naming them "is cool" and then circle back to give them a useful name later, but I forgot to do that this time!

api/document/js/parse-doc.ts
toolness: Oh blarg, right.  Um, so when @rtwell and I were talking, we actually discussed changing things on the PDF importer side, rather than on the editor side--e.g. ensuring the PDF importer (and the schema validator that we have yet to write on the backend) only made footnotes that had citations, and any footnotes that lacked citations would simply be converted into paragraphs (potentially with some commentary from the importer mentioning that they thought it was a footnote, but couldn't find the citation).

Umm let me think about this a bit and get back 2u.

api/document/js/__tests__/serialize-doc.test.ts
toolness: Hmm, so I'd argue that in certain cases like this one, they're actually _more_ readable than doing a bunch of assertions: in this particular case, the vast majority of the snapshot has to do with the way PM-style footnotes are converted into API-style footnotes (the only "chaff" is the surrounding paragraph).  Also, being able to provide an extremely simple, self-contained example that effectively says "look, if you start with this, it should convert to this" can be more illuminating than making a bunch of assertions about the result that require the reader to work backwards from them to understand what the functionality is actually doing.

That said, I do agree that in a _lot_ of cases, they can have the disadvantages you mentioned.

I also agree that having the snapshots be in a separate file can make them harder to read.  I originally wanted to just copy-paste the JSON representation into an `expect().toEqual()`, but the complaining of the linter led me to avoid that approach.

api/ereqs_admin/static/admin/module/_editor.scss
toolness: Oh, I styled it this way b/c it's what Austin's mockups looked like, but yeah, we should double-check.

api/document/js/parse-doc.ts
toolness: So, I have some concerns about keeping uncited footnotes around the PM schema as "not-footnotes"--specifically, I want to eventually add logic for adding and removing footnotes with auto-numbering, but if that logic needs to account for uncited footnotes, it'll complicate the logic a lot.

(Also, I'm unclear on where the line of code you quoted is supposed to go, or how it'd be used...)

What if we, at the API level (although I guess we could do it as a pre-processing step in the editor too), actually output the document ensuring that all footnotes had citations, _inserting_ citations in their most appropriate locations if they didn't exist?  e.g., if a document had footnotes+citations for 1 and 3, but only a footnote w/o a citation for 2, it could simply find the citation for 1 and add a citation for 2 immediately after it.

cmc333333: I think that's an interesting strategy, though I'd put that at the pdf-to-database layer. Ideally, we're removing the `footnote` (api) node type altogether and replacing it with those inline versions; getting there will take some time...

toolness: Cool, that sounds like a good idea; I can file a PR for it later. For now I understand your one-line solution, now that you explained it to me over Slack, so I will bust that up!

toolness:
@cmc333333 ok, I think this is good to go!

toolness:
woooooooooooooooooooooooooooooooooooooooooooooooooooot

PR #1023, opened 2018-02-14
Refactor and document pathToResolvedPos better.

cmc333333:
Stemming from @toolness's suggestion in #1006, this adds more docs to pathToResolvedPos.
It also switches the calling syntax to variadic arguments rather than an array and shuffles some non-surfaced logic around.

What do you think?

### DIFF
--- a/api/document/js/__tests__/commands.test.ts
+++ b/api/document/js/__tests__/commands.test.ts
@@ -17,7 +17,7 @@ import {
 import { collectMarkers } from '../list-utils';
 import schema, { factory } from '../schema';
 import serializeDoc from '../serialize-doc';
-import pathToResolvedPos, { NthType, SelectionPath } from '../path-to-resolved-pos';
+import pathToResolvedPos, { NthType, Selector } from '../path-to-resolved-pos';
 
 function executeTransform(initialState: EditorState, transform): EditorState {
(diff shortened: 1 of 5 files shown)


api/document/js/__tests__/commands.test.ts
```diff
@@ -41,7 +41,9 @@ describe('appendNearBlock()', () => {
     const selection = new TextSelection(pathToResolvedPos(
       doc,
       // Inside the 'bbb' paragraph
-      [new NthType(1, 'para'), 'inline', 'b'.length],
+      new NthType(1, 'para'),
+      'inline',
+      'b'.length,
```
toolness: I was throwing around the idea of whether it'd be more readable to have an `NthChar` class analogous to your `NthType` class.  I think it might be a bit more readable, but it'd also be more verbose... 🤷‍♂️

toolness:
Ummm this PR is going to wreak havoc with my renaming of `inline` in #1021. Blarg.

toolness:
Is it ok if we hold off on merging this until #1021 is merged, and then I deal with the conflicts in this branch?  I am just super tired of working on footnotes and would like to get that merged ASAP if possible.

cmc333333:
Absolutely; this can wait till post demo.

PR #1025, opened 2018-02-14
Add basic tests for ProseMirror menu buttons.

tadhg-ohiggins:
If all the buttons
Are clicked in the deep forest
Will anyone know?

Adds a basic check to verify that the ProseMirror editor menu bar has the buttons we think it should have.

### DIFF
--- a/api/document/js/__tests__/menu.test.ts
+++ b/api/document/js/__tests__/menu.test.ts
@@ -1,0 +1,36 @@
+import { flatten } from 'lodash/array';
+import { EditorView } from 'prosemirror-view';
+import schema, { factory } from '../schema';
+import { EditorState } from 'prosemirror-state';
+import menu from '../menu';
+import { Api, JsonApi } from '../Api';
+const jsdom = require('jsdom');
+const { JSDOM } = jsdom;
+
+describe('menu functions', () => {
+  it('has the expected buttons', () => {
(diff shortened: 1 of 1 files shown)


api/document/js/__tests__/menu.test.ts
```diff
@@ -0,0 +1,34 @@
+import { flatten } from 'lodash/array';
+import { EditorView } from 'prosemirror-view';
+import schema, { factory } from '../schema';
+import { EditorState } from 'prosemirror-state';
+import menu from '../menu';
+import { Api, JsonApi } from '../Api';
+
+describe('menu functions', () => {
+  it('has the expected buttons', () => {
```
toolness: Should this be "_have_ the expected buttons"?  I'm just trying to read out the first args of `describe()` and `it()` in a sentence, and it comes out to "menu functions has the expected buttons" which sounds a bit funky.

cmc333333: Maybe s/menu functions/menu function? `it('have ` also isn't quite right...

tadhg-ohiggins: This is an amusing set of comments to find… I was reading the “sentences” purely as “it…” and not as including the argument to `describe`. I think the easiest change to make here is to `s/menu functions/menu`.

tadhg-ohiggins:
Thanks @cmc333333 for helping me find my way through the ProseMirror woods to the relevant menu details.

PR #1027, opened 2018-02-15
[WIP EXPERIMENT] Improve cursor behavior around footnotes

toolness:
This is an experimental PR atop #1021 that tries to make footnotes a bit easier to edit.

At the moment, the cursor can't actually be positioned at the very beginning of a footnote: it can be positioned either immediately before the footnote or after its first character, but not in between.  This PR attempts to resolve that in a kind of weird way: by placing a [zero-width space](https://en.wikipedia.org/wiki/Zero-width_space) at the beginning of each footnote when parsing the document, and stripping it out before serializing the document.

We also need to fix up the document when the user decides to press backspace at the beginning of a footnote, thereby removing the zero-width space.  Right now I'm simply deleting the entire footnote, but I'm only really doing that because it was trivial to implement: ideally I suspect the content of the footnote should be "lifted" into its surrounding paragraph, but I'm not sure (we should talk to design about this).

## Other things I tried

Before settling on the zero-width space solution, I tried creating a new node type called a `footnoteMarker` that would be placed at the beginning of a footnote, but that resulted in weird behavior from browsers (I also tried setting its `atom` to `true` and that didn't improve things).  We could revisit that solution again if there's something I missed, though.

## Future challenges

I'm realizing that there are downsides to making footnotes an inline node instead of a mark.  Most importantly, I'm not sure how to deal with the situation when the cursor is at the end of a footnote: right now, typing anything at the end of a footnote makes the new text be part of a footnote.  If the footnote were a mark type, we could simply add a button that toggled the active marks to not include the footnote mark, and users could treat footnotes just like they treat bold/italic/etc in other WYSIWYG editors.  However, because footnotes _aren't_ marks, I'm not really sure what to do.

### DIFF
--- a/api/document/js/__tests__/parse-doc.test.ts
+++ b/api/document/js/__tests__/parse-doc.test.ts
@@ -150,7 +150,7 @@ describe('convertContent()', () => {
     expect(result[0].type.name).toBe('inlineFootnote');
     expect(result[0].attrs.emblem).toBe('5');
     expect(result[0].content.toJSON()).toEqual([{
-      text: 'Some text ',
+      text: '\u200bSome text ',
       type: 'text',
     }]);
   });
--- a/api/document/js/__tests__/serialize-doc.test.ts
+++ b/api/document/js/__tests__/serialize-doc.test.ts
@@ -58,7 +58,7 @@ describe('serializeDoc()', () => {
       schema.nodes.paraText.create({}, [
         schema.text('hello'),
         schema.nodes.inlineFootnote.create({ emblem: '2' }, [
-          schema.text('i am a footnote'),
+          schema.text('\u200bi am a footnote'),
         ]),
       ]),
     ]);
--- a/api/document/js/fixup-doc.ts
+++ b/api/document/js/fixup-doc.ts
@@ -2,7 +2,7 @@ import { Node } from 'prosemirror-model';
 import { EditorState, Plugin, Transaction } from 'prosemirror-state';
 
 import { renumberList } from './list-utils';
-import schema from './schema';
+import schema, { BEGIN_FOOTNOTE } from './schema';
 
 const shouldDeleteChecks = {
   list: (node: Node) => node.content.childCount === 0,
@@ -10,6 +10,9 @@ const shouldDeleteChecks = {
   para: (node: Node) => (
     node.content.childCount === 1 // Just the paraText
     && node.textContent === ''
+  ),
+  inlineFootnote: (node: Node) => (
+    node.textContent[0] !== BEGIN_FOOTNOTE
   ),
 };
 
--- a/api/document/js/parse-doc.ts
+++ b/api/document/js/parse-doc.ts
@@ -3,7 +3,7 @@ import { flatten } from 'lodash/array';
 import { Mark, Node } from 'prosemirror-model';
 
 import { ApiNode, ApiContent } from './Api';
-import schema, { factory } from './schema';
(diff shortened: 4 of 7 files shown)


cmc333333:
I think the problem ultimately stems from browser support around `::before` and `contenteditable`. What if we did this instead?
```diff
diff --git a/api/document/js/schema.ts b/api/document/js/schema.ts
index 2c5cce17..02b68d6d 100644
--- a/api/document/js/schema.ts
+++ b/api/document/js/schema.ts
@@ -45,10 +45,16 @@ const schema = new Schema({
       content: 'text*',
       group: 'inline',
       inline: true,
-      toDOM: node => ['span', {
-        class: 'inline-footnote',
-        'data-emblem': node.attrs.emblem,
-      }, 0],
+      toDOM: node => [
+        'span',
+        { class: 'inline-footnote' },
+        [
+          'span',
+          { class: 'footnote-marker', contenteditable: 'false' },
+          `Footnote ${node.attrs.emblem}:`,
+        ],
+        ['span', 0],
+      ],
     },
     heading: {
       content: 'text+',
diff --git a/api/ereqs_admin/static/admin/module/_editor.scss b/api/ereqs_admin/static/admin/module/_editor.scss
index ed0f972e..3b891dfe 100644
--- a/api/ereqs_admin/static/admin/module/_editor.scss
+++ b/api/ereqs_admin/static/admin/module/_editor.scss
@@ -11,9 +11,10 @@
     @include font-san-serif;
   }

-  .inline-footnote::before {
-    content: 'Footnote ' attr(data-emblem) ': ';
-    padding-left: 0.5em;
+  .footnote-marker {
+    margin-left: 0.5em;
+    margin-right: 0.5em;
+    white-space: nowrap;
     @include font-san-serif-bold;
   }
```

Works great in Firefox:
firefox

Still has the annoying padding issue in Chrome/Safari, but does resolve the cursor-in-footnote issue:
safari
chrome

toolness:
Noice! I am scared of setting `contenteditable` to `false` but we can give it a shot.

For now I've solved the "how do we add normal text after a footnote" problem by adding a "F" (for "footnote") button in the menu bar that effectively treats footnotes like a mark--that is, it appears "active" when the cursor is in a footnote, and "inactive" when the cursor is outside a footnote.  If the button is clicked while inside a footnote, it moves to the end of the footnote and adds a single space as non-footnoted text and moves the cursor there.  Otherwise, clicking the button will create a new footnote.

toolness:
@cmc333333 hmm, so I tried the [alternative solution](https://github.com/18F/omb-eregs/pull/1027#issuecomment-366266963) you suggested and it seems like browser behavior diverges when <kbd>Backspace</kbd> is pressed at the beginning of a footnote: on Chrome, just the span containing the footnote marker is removed, but _without actually triggering any kind of editor transaction_, meaning that it's not even possible to "catch" the backspace and prevent it from deleting that span, or undo the deletion of the span. On Edge, pressing <kbd>Backspace</kbd> does the exact same thing as <kbd>←</kbd>, i.e. the cursor moves before the footnote marker span and nothing is deleted.

So given that both solutions are kind of weird hacks, but mine seems to behave consistently across browsers and not raise any additional new strange behaviors, I think we should probably just go with my solution for now, as odd as it is.  Blarg, `contenteditable` is so weird.

toolness:
Ok, I think I am going to pronounce this experiment **DONE**!

I will bust out separate individual PRs that implement the functionality in here for production.

Issue #588, opened 2017-10-27
Don't use '--quiet' when running npm


toolness:
When following the instructions in the README to run `docker-compose up dev`, my system printed `Installing Node dependencies` and then did nothing for a good 15 minutes or so. I tried visiting `http://localhost:8002/` to no avail, so I thought that perhaps the docker container had hung somehow, and I pressed CTRL-C to abort it.

Unfortunately, I think that it _was_ running NPM, but in `ui/.docker/deps_ok_then`, it's run with `--quiet`, so nothing was displayed, and I had no easy way to tell what was going on. When I re-ran `docker-compose up dev`, my container was in a partially-installed state; `deps-ok` seemed to exit with a zero exit code, so `npm install` wasn't run, and then when `watch-run.sh` ran, it mentioned that the `webpack` binary didn't exist.  So I had to manually go into the container and run `npm install` to finish installing everything.

Anyways, this is all to say that I'm confused as to why `--quiet` is passed to `npm install`, especially when it's gated by `deps-ok` (and thus, I assume, only actually run when something needs to be installed/upgraded).  If there's a good reason for it, I think that at the very least, `Installing Node dependencies` should be changed to something like `Installing Node dependencies (this could take some time, your system is not hanging if you don't see anything printed for a while)`.

cmc333333:
Per our conversation on Friday, we added the `--quiet` to make CI logs easier to read, so removing it outright isn't ideal. We talked about seeing if yarn has better output in CI (and other non-rewritable terminals) and switching if so.

PR #1037, opened 2018-02-20
Update everything to Node 8

### LINKED ISSUE #1030: Update to Node 8
We're using Node 6 in our docker images (and in cloud.gov), yet we're also using package-lock.json files, which aren't part of the bundled npm until Node 8. Let's upgrade everything to the latest LTS and remove the shrinkwrap files.

cmc333333:
The Node 8 image uses a new version of `npm`, which removes the `npm-shrinkwrap.json` file in favor of `package-lock.json` (which now also includes dev dependencies).

Unfortunately, our binary libraries will likely fail out due to being built for Node 6. To resolve, it's easiest to:
```sh
rm -rf */node_modules
# these commands can be ran in parallel
bin/api-npm install
bin/styleguide-npm install
bin/ui-npm install
```
Note that, particularly the ui step, may take a while (~20m) to run.

This resolves #1030


api/.docker/deps_ok_then
```diff
@@ -2,7 +2,7 @@
 
 if [ ! -x node_modules/.bin/deps-ok ]; then
   echo "Installing deps-ok"
-  npm install --quiet deps-ok
+  npm install --save-dev --quiet deps-ok
```
toolness: Sorry, why are we doing `--save-dev` now?

toolness:
Oh, also, I just noticed that we're now using CircleCI's fancy caching mechanism as of #999, which I _think_ should mean that CircleCI logs shouldn't be very noisy if we remove the `--quiet` from all our `npm install` invocations (except for when we clear the CircleCI cache).

As I mentioned in #588, I think that removing the `--quiet` would really help responsiveness for developers who are onboarding (or who have just blown away their `node_modules`). Do you think we should do that here or a separate PR?

api/.docker/deps_ok_then
cmc333333: The `deps-ok` lib (which confirms top-level modules are installed) is only used during development. Previously, we were running `npm install` _without_ `--save`, which would add the lib to our environment but not affect our package/shrinkwrap files. The new version of npm automatically updates these files, however, so we need to place it differently.

cmc333333:
Good call, happy to remove the `--quiet`s now

PR #1040, opened 2018-02-20
Log unfriendly Api errors in editors.

toolness:
Currently, our `Api` class catches errors and, for ones that aren't actually related to REST API calls, just sets the current status to `An error occurred.` and does nothing else.  If the calling code doesn't re-throw the error, this is actually really unhelpful for developers, because instead of seeing useful information about an error, they just see the uninformative status message and nothing logged to the console.

There's multiple solutions to this problem. This one fixes things by just logging the error to the console if it detects that the error is an unhelpful one.

### DIFF
--- a/api/document/js/Api.ts
+++ b/api/document/js/Api.ts
@@ -14,6 +14,8 @@ export function setStatusError(e: Error) {
   const data = e['response'] && e['response']['data'];
   if (data) {
     errMsg += '\n' + makeErrorFriendly(data);
+  } else {
+    console.error(e);
   }
   setStatus(errMsg, 'editor-status-error');
 }


api/document/js/Api.ts
```diff
@@ -14,6 +14,8 @@ export function setStatusError(e: Error) {
   const data = e['response'] && e['response']['data'];
   if (data) {
     errMsg += '\n' + makeErrorFriendly(data);
+  } else {
+    console.error(e);
```
toolness: Rather than logging it here, we could instead log the error in `create-editor-state.ts`'s `catch` clause, because that's currently the only place we're calling `setStatusError` without re-throwing the error.  I'm not sure which strategy is best, though.

toolness: Also, I'm not sure if we want to be extra careful and test to see if `window.console.error` even exists before attempting to call it.  Do all browsers have that function these days?

cmc333333: MDN claims support back to IE8, so I think we're good. Thanks for thinking of it, though!

PR #1041, opened 2018-02-21
Weird Windows fixes.

toolness:
Strangely, the binary symlinks installed in the node environments of `ui` and `api-ui` work okay, but I'm frequently running into https://github.com/docker/for-win/issues/137 for the node environment of `styleguide`, whereby the container's processes think that the symlinks in `node_modules/.bin` are files rather than symlinks.

This works around the bizarre problem by passing a direct path to the relevant scripts.

It also exposes `USE_POLLING` to the `styleguide` container, so that polling works for webpack.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -91,6 +91,8 @@ services:
     working_dir: /usr/src/app/styleguide
     entrypoint: .docker/deps_ok_then .docker/modify_path_then
     command: .docker/build_run_styleguide
+    environment:
+      USE_POLLING: ${USE_POLLING:-false}
     ports:
       - 3000:3000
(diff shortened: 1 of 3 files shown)


toolness:
Yeah, that is probably not a bad idea... I guess I will bust up a separate PR for that kind of thing.

I should also mention that I originally thought there was an alternative solution to this: `npm install` has an option called `--no-bin-links` which "will prevent npm from creating symlinks for any binaries the package might contain".  I thought this meant it might make _copies_ of the relevant files instead of symlinks, but it doesn't do that--it just avoids doing anything with that `.bin` directory altogether.  Ah well.

PR #1044, opened 2018-02-21
Fixup node 8 deployment by bumping libs

cmc333333:
In production, we were missing a necessary dependency. This was likely due to
trying to generate the ui's package-lock from the shrinkwrap file. This
updates the package-lock files by regenerating them after clearing the
`node_modules` dir.

See [successful deploy](https://circleci.com/gh/18F/omb-eregs/3816).


cmc333333:
Thanks @toolness !

